Issue #30, opened 2022-05-21
Update mermaid graph to show protocol used between services


cartersocha:
grpc & http communication should be labeled between the different services as an enhancement to the initial graph like http is used at the top.

https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/27

cartersocha:
This is generally called out if we're NOT using gRPC. Closing for now. @julianocosta89 fyi

PR #47, opened 2022-05-25
Add desired service map

austinlparker:
Contributes to #30 

Adds desired end state service graph + color coding for services + legend


austinlparker:
bad at git, please ignore

PR #74, opened 2022-05-30
Feature/docker compose

### LINKED ISSUE #24: Migrate build to use `docker compose`
The `docker compose build` will build all services on all platforms

- [ ] Create a `docker-compose.yml` at the root of the repository
- [ ] Adapt all services `Dockerfile` files
  - [ ] Remove duplication of `demo.proto` file
  - [ ] Compile protobuf source files during build
- [ ] Testing

References
- https://docs.docker.com/compose/

julianocosta89:
Partially fixes #24 .

## Changes

- Added `.env` file with all environment variables which will be used by docker compose.
- Added to README the steps on how to deploy using docker compose.
- Added `compose.yml` file with all the configurations to start the webstore-demo-app.
- Removed unused files
- Renamed the `src/otelcollector/conf.yml` to `src/otelcollector/otelcol-config.yml` .
- Updated otel-collector Jaeger exporter configuration with the new sintaxe: `jaeger.insecure.true`


julianocosta89:
I believe we should split #24 .

The docker compose is working and we should probably think in a way of testing it, to ensure future updates won't break it.

Besides that, the following tasks should be split into separated tickets, as they are not related to the docker compose task:
- Adapt all services Dockerfile files
- Remove duplication of demo.proto file
- Compile protobuf source files during build

cartersocha:
Just got this to run locally on my mac & verified the jaeger trace / console output. 

Took me about 11-12 mins to from clone to the end output. Pretty cool to be able to spin it up so easily. 

We'll need to start looking at the delta between current output & otel requirements.

LGTM.

cartersocha (approved):

cartersocha:
We can flesh out the docs in another pr

reyang (approved):
LGTM.

PR #75, opened 2022-05-31
Small readme format updates

cartersocha:
Removed some small sections & added structure for the local quickstart.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,16 +2,16 @@
 
 ## Under Construction
 
-This repo is a work in progress
+This repo is a work in progress. If you'd like to help, check out our [contributing guidance](#contributing).
 
-## Quickstart
+## Local Quickstart
 
 ### Pre-requisites
 
 - Docker
 - [Docker Compose](https://docs.docker.com/compose/install/#install-compose) v2.0.0+
 
-### Steps
+### Step 1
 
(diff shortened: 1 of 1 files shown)


cartersocha:
Please review & merge @julianocosta89

README.md
```diff
@@ -2,37 +2,44 @@
...
-## Quickstart
+## Local Quickstart
 
 ### Pre-requisites
 
 - Docker
 - [Docker Compose](https://docs.docker.com/compose/install/#install-compose) v2.0.0+
 
-### Steps
+### Step 1
```
reyang, line 15: minor: for folks who want to save the links/anchors, this might introduce problems. Doesn't have to block this PR though.

cartersocha: How would this potentially break links or introduce issues? I can adjust to a better layout

reyang: e.g. if something stored the current link, it'll be something like <https://github.com/cartersocha/opentelemetry-demo-webstore/tree/carterdocumentation2#step-3>, later if we added two more steps, it'll become `step-5`.

cartersocha: Ahh makes sense. I’ll adjust to actual step titles rather an ordered list

julianocosta89 (approved):

PR #79, opened 2022-06-01
Add default server port constant

laziobird:
## Changes

A port constant is defined in one place and is used anywhere .

### DIFF
--- a/src/adservice/src/main/java/hipstershop/AdConstant.java
+++ b/src/adservice/src/main/java/hipstershop/AdConstant.java
@@ -1,0 +1,21 @@
+ package hipstershop;
+ /*
+  * Copyright 2018, Google LLC.
+  *
+  * Licensed under the Apache License, Version 2.0 (the "License");
+  * you may not use this file except in compliance with the License.
+  * You may obtain a copy of the License at
+  *
+  *     http://www.apache.org/licenses/LICENSE-2.0
+  *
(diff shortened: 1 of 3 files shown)


mic-max:
I feel like each service requiring an environment variable before running would be okay for this demo. So if it's not present I think crashing early instead of even trying to use a default port number would be better since the port number `9555` in this service's case is also defined in the `compose.yml` and the frontend which `depends_on` this service being present at the address:port defined in the `.env` file is not found but the `adservice` is running, just on a different port, could cause confusion.

My idea:

Define all addresses in the `.env` file. If a service can't load it's port or another service it depends on from the environment variables then the program should exit with and error.

laziobird:
Thanks for so good suggestions and notices !   @reyang  @mic-max   .

cartersocha:
LGTM. @mic-max, @reyang, any further suggestions or comments?

mic-max:
> LGTM. @mic-max, @reyang, any further suggestions or comments?

I prefer the environment variable only approach instead of this

reyang:
> > LGTM. @mic-max, @reyang, any further suggestions or comments?
> 
> I prefer the environment variable only approach instead of this

+1

laziobird:
> > > LGTM. @mic-max, @reyang, any further suggestions or comments?
> > 
> > 
> > I prefer the environment variable only approach instead of this
> 
> +1

OK，i can add this environment variable tomorrow ☺

Issue #49, opened 2022-05-30
Review and enhance tracing support for frontend (Go) 


cartersocha:
The following is a list of requirements that we need to evaluate before declaring v1 for trace telemetry. These requirements are across the entire application; Not all services will meet all requirements. Determine the relevant features for this service. 

- [x] Automatic Instrumentation is being used where appropriate. - not appropriate
- [x] Library instrumentation is being used if automatic instrumentation is unavailable.
- [ ] Services extend automatic instrumentation.
  - [x] New attributes/events attached to existing spans.
  - [ ] New spans are being created from existing spans.
- [ ] Automatic and Manual Context Propagation are being demonstrated.
  - [x] Interprocess context propagation is demonstrated.
- [x] Telemetry is not being sampled upfront.
- [x] Telemetry is not being filtered upfront.
- [ ] Baggage is being set and read appropriately (i.e., baggage must be explicitly set as attributes on spans)

Referencing: https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/42

Service: https://github.com/open-telemetry/opentelemetry-demo-webstore/tree/main/src/frontend

fatsheep9146:
1. Automatic Instrumentation is being used where appropriate.

I think the project `frontend` meets this requirements. Since the two main libraries that project used are both instrumented appropriately. 

- `google.golang.org/grpc` 
- `github.com/gorilla/mux`

the instrumentation is initialized in following lines:

- https://github.com/open-telemetry/opentelemetry-demo-webstore/blob/a49f3b8f30e344c88978d09ea851230540981b3f/src/frontend/main.go#L129
- https://github.com/open-telemetry/opentelemetry-demo-webstore/blob/a49f3b8f30e344c88978d09ea851230540981b3f/src/frontend/main.go#L166-L170

cartersocha:
I think that covers the instrumentation library requirement. Auto instrumentation is one of those words that means something different to everyone. 

In this case auto instrumentation is the agent / collector running in a sidecar right @austinlparker ?

cartersocha:
@puckpuck could you check off the requirements you completed in your pr?

From a quick glance it seems like telemetry is not being sampled / filtered, custom attributes are being added to all spans, & library instrumentation is pre-existing. Did I miss anything & would we consider this interprocess context prop with the middleware attribute example too?

puckpuck:
We can check off the following from the list.

* [ ]  Services extend automatic instrumentation.
[...]

Issue #42, opened 2022-05-23
Ensure sufficient breadth in trace features across services


austinlparker:
The following is a list of requirements that we need to evaluate before declaring v1 for trace telemetry. These requirements are across the entire application; Not all services will meet all requirements.

- [ ] Automatic Instrumentation is being used where appropriate.
- [ ] Automatic instrumentation is being injected through the OpenTelemetry Operator.
- [ ] Library instrumentation is being used if automatic instrumentation is unavailable.
- [ ] Services extend automatic instrumentation.
  - [ ] New attributes/events attached to existing spans.
  - [ ] New spans are being created from existing spans.
- [ ] Automatic and Manual Context Propagation are being demonstrated.
  - [ ] Interprocess context propagation is demonstrated.
  - [ ] Intraprocess context propagation is demonstrated.
- [ ] Trace links are being demonstrated in appropriate places (security boundary, fan-in/fan-out)
- [ ] Telemetry is not being sampled upfront.
- [ ] Telemetry is not being filtered upfront.
- [ ] Baggage is being set and read appropriately (i.e., baggage must be explicitly set as attributes on spans)

cartersocha:
depending on timeline for new services this might be a rolling issue as new components are added. Should we plan on splitting these into stages?

cartersocha:
We are done on tracing!

PR #82, opened 2022-06-02
add span attributes to frontend

puckpuck:
Updates #49 to add new attributes to existing spans.

## Changes

Adds attributes to frontend service spans
All handlers are wrapped with common attributes (session, request, currency, user). 
Most handlers have application-specific attributes added
Defines an `instr` package that contains some semantic conventions

* [x] Design discussion issue #42  and #49 

This adds some conventions (ie: `app.` prefix) to attribute names.  We should settle on clean conventions before pushing similar instrumentation to other services.


cartersocha:
When I run your branch the spans look the same as the current version. Could you post a screenshot of your output?

cartersocha:
Thanks for the pic. I was checking for attributes but had been glancing through the console output which doesn't have them for whatever reason. 

LGTM

cartersocha (approved):

cartersocha:
Are you planning on documenting this in the service readme?

puckpuck:
> Are you planning on documenting this in the service readme?

That's the new PR ;)

puckpuck:
> Are you planning on documenting this in the service readme?

That's the next PR 😉

PR #92, opened 2022-06-02
Recommendation Service Update - Python

mic-max:
Similar ideas of replacing generated code with generating such code in the Dockerfile and making environment variables a requirement rather than falling back onto a default value are intended to be done in other projects.

## Changes

- Replace pb2 generated files with them being compiled in the `Dockerfile`
- Make port and depended on service environment variables required, otherwise program will throw an exception
- Update client program to work
- Delete `genproto.sh`

PS: This is the only python service that we intend to keep so that's why we shouldn't try to make this change to the `emailservice`


src/recommendationservice/client.py
```diff
@@ -24,18 +23,13 @@
...
-    # get port
-    if len(sys.argv) > 1:
-        port = sys.argv[1]
-    else:
-        port = "8080"
-
     # set up server stub
-    channel = grpc.insecure_channel('localhost:'+port)
-    
+    channel = grpc.insecure_channel('localhost:8080')
```
julianocosta89, line 27: Maybe we should consider using the env var `RECOMMENDATION_SERVICE_PORT` here as well.

julianocosta89 (changes requested):

compose.yml
```diff
@@ -141,11 +141,14 @@ services:
   # RecommendationService
   recommendationservice:
     build:
-      context: ./src/recommendationservice
+      context: ./
+      dockerfile: ./src/recommendationservice/Dockerfile
     ports:
-      - "${RECOMMENDATION_SERVICE_PORT}"
+      - "${RECOMMENDATION_SERVICE_PORT}:${RECOMMENDATION_SERVICE_PORT}"
```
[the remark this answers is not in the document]
mic-max: I added this so the `client.py` would work, I could remove and add an instruction to the `README.md` file with an extra step before running it to add the `:${RECOMMENDATION_SERVICE_PORT}` back to this. That probably makes more sense to do.

src/recommendationservice/client.py
mic-max: That would be better. I could load the port or address from the `.env` file using a library such as https://pypi.org/project/python-dotenv/

julianocosta89 (approved):
LGTM!

PR #94, opened 2022-06-02
Remove all `genproto.sh` files

mic-max:
This change is towards the goal of moving the generation of the generated protobuf files to the `Dockerfile` and not checking them into the repository.

These files are not used by anything.

## Changes

Remove all `genproto.sh` files


julianocosta89:
LGTM and I have only one note (not a blocker).
Removing the `genproto.sh` files will not impact the services, but it may impact on ppl trying to build/compile the project locally.
Hence, it would be good to have a README, explaining how to compile the `proto` files per service.
Something like @mic-max have done in this PR already: https://github.com/open-telemetry/opentelemetry-demo-webstore/blob/300001bb6f895bffca5910eccaa7569285308f8e/src/recommendationservice/README.md

julianocosta89 (approved):

mic-max:
By build/compile the project locally I guess you also mean without using docker, right? for example, they run `go build` and it complains about the missing files

In follow-up PRs we could include this instruction in the service READMEs like you linked the example where I did so 👍

PR #109, opened 2022-06-04
Rewrite emailservice in ruby

### LINKED ISSUE #37: Reimplement 'emailservice' in Ruby
**Objective**

Re-write the emailservice to Ruby.

ahayworth:
## Changes

This PR rewrites the `emailservice` in Ruby. We make the following notable changes:
- This is an **http** service, since Ruby and GRPC are not the best of friends. We modify the calling `checkoutservice` accordingly.
- I pinned us to otelcol 0.52.0 - whatever version I had locally on my M1 Mac was segfaulting, and pulling the latest version fixed it. This necessitated a small change to the config, since some keys shuffled around in the `jaeger` exporter section.
- I changed some of the relevant environmental variables to reflect the HTTP nature of this service, and to properly configure Ruby's OTLP/HTTP exporter; but I tried to change those in a non-invasive way for any other service.

We were able to do this pretty easily, especially since the GRPC request proto for emailservice was very simple and readily serialized to JSON.

There are two potential things wrong with it:
- I'm pretty sure that the formatting of the email message looks slightly different then the original python version intended ... but honestly I don't think anyone ever looked at it before. :shrug:
- We render the email body into the log, which is a tad bit noisy. On the other hand, the logs are very noisy as-is.

Fixes #37


ahayworth:
Obligatory screenshot!

Ruby spans in checkout trace

julianocosta89:
Hello @ahayworth, why is gRPC not best friends with Ruby?
There are no limitations listed here: https://grpc.io/docs/languages/ruby/

The PR looks good to me,  but before approving this PR I'd like to hear the opinion from @cartersocha and @austinlparker.

I personally do not care if it is gRPC or HTTP, but as far as I remember we would keep gRPC wherever possible in the sample.

ahayworth:
> Hello @ahayworth, why is gRPC not best friends with Ruby? There are no limitations listed here: https://grpc.io/docs/languages/ruby/

Lack of limitations does not imply that it works well, sadly. I wish they were better friends.

Technical things:
- The GRPC Ruby code is difficult enough to understand that I personally [gave up](https://github.com/open-telemetry/opentelemetry-ruby/pull/854) on implementing GRPC auto-instrumentation.
- A lot of why it's hard to understand is because it's a C-extension doing a lot of the hard work. [That can be hard to debug](https://news.ycombinator.com/item?id=20026853).
- There are quite a lot of [issues](https://github.com/grpc/grpc/issues?q=is%3Aissue+label%3Alang%2Fruby+) for Ruby, and a lot of them are "the gem is missing" or "it segfaults" or "it's unable to compile on these platforms" etc. An upsettingly high number.
- Some serious issues have been open for years: [support for resetting connections after fork](https://github.com/grpc/grpc/issues/8798), [GRPC reflection](https://github.com/grpc/grpc/issues/2567), [adding an option to disable fail-fast](https://github.com/grpc/grpc/issues/6768), [worker threads hanging if a bad response is sent](https://github.com/grpc/grpc/issues/25373), [ruby server doesn't follow spec](https://github.com/grpc/grpc/issues/25852), etc.

Socio-technical things:
- There has been _very_ little activity in the Ruby GRPC library in recent years, once you exclude autogenerated C-extension changes. In fact, in `src/ruby/lib/grpc/generic` (where the high-level bits are, before you drop down to the C extension) there has only been 1 commit in the past _year_.
- It kinda seems like the folks over at the GRPC project aren't interested in providing first-class support for Ruby. [Others have noted the same](https://news.ycombinator.com/item?id=31591443), but even just [benchmarking shows that it's pretty awful](https://github.com/LesnyRumcajs/grpc_bench/wiki/2022-04-23-bench-results) (granted, python doesn't look too great there either).
- Those same C-extensions mean that GRPC flat-out [doesn't work with `jruby`](https://github.com/grpc/grpc/issues/6705), which is still an important platform for Ruby. This alone holds back broader adoption in the broader Ruby community.
- The "Rails effect" means that just most things in the Ruby ecosystem tend to revolve around Rails - and in turn, that means they tend to be focused on HTTP.

OpenTelemetry-specific things:
- We were recently asked in the Ruby SIG if we could implement GRPC as a required OTLP transport mechanism. The [comment here](https://github.com/open-telemetry/opentelemetry-specification/issues/1885#issuecomment-928200776) speaks to the issues with Ruby and GRPC in general, and was part of why OTLP/HTTP is actually the required OTLP transport now for SDKs.

---

No small amount of these issues are related / intertwined with `protobuf`, too - it's also an issue on Ruby. I've been writing this comment for quite awhile now so I won't dig up as much for discussion, but here's a few things:
- [currently open issue for memory leaks on Ruby](https://github.com/protocolbuffers/protobuf/issues/10106)
- [doesn't seem to work well on jruby, sometimes not at all](https://github.com/protocolbuffers/protobuf/issues/1594)
- [doesn't support custom options, after 6 years](https://github.com/protocolbuffers/protobuf/issues/1198)
- [massive memory inflation between 3.14 -> 3.15 causing production outages](https://github.com/protocolbuffers/protobuf/issues/8337) (this one was especially upsetting because there was a known issue on older versions of ruby that weren't yet EOL - and it seems like nobody even tested them)
- [no single version of ruby can work with protobuf 3.20](https://github.com/protocolbuffers/protobuf/issues/9733)
- [utterly incorrect method aliasing](https://github.com/protocolbuffers/protobuf/issues/7806)
- [unfriendly APIs with user pain, no resolution](https://github.com/protocolbuffers/protobuf/issues/4969)

---

All of this aside: it's definitely possible to use GRPC with Ruby. But it's generally a poor experience, and I've _personally_ been annoyed every time I try to do it.

There are a lot of outstanding technical issues with both GRPC and protobuf, and the ruby wrappers don't see a lot of love from the maintainers. There are frequently issues with new versions, and my sense is that "nobody on these projects is testing these releases with Ruby" - just the sheer number of "we can't install the new version on ruby" issues alone demonstrates as much.

Setting that aside, using the grpc gem is just confusing and not ruby-like. Trying to dive in the code is a huge mess, from my experience. It's not impossible, but it feels like a half-baked project for Ruby in a lot of ways. Of course, that's a subjective feeling on my part but I think I've linked enough issues and discussions to at least show I'm not alone. 😆 

I really don't know why this is, I guess. My own personal thought is that it's a bad idea to just implement client libraries as thin wrappers around C extensions without really trying to implement things in a language-first fashion.

---

Anyways: is it impossible to do this as a GRPC service? No, but I'd really prefer not to. Setting aside the general annoyances, it's just not a good _demo_ for Ruby. The community is heavily entrenched in HTTP, and that's what is shown off the best. We don't even have GRPC instrumentation - and precious few people have even _asked_. So even if we did this in GRPC, it's just not going to be as nice as an HTTP version.

src/emailservice/Dockerfile
```diff
@@ -12,28 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM python:3.10-slim
+FROM ruby:3.1.2
```
mic-max, line 15: Could a slim version of this image be used?

mic-max (approved):
LGTM!

julianocosta89:
Awesome @ahayworth!
Thank you very much for the explanation.

Discussed with everyone at the SIG meeting and this is good to go! 🚀

julianocosta89 (approved):

src/emailservice/Dockerfile
ahayworth: @mic-max I had problems installing some gems (`puma` and `thin` both, I think) when I used the `-slim` image. I could either:
- Use a different webserver that's pure-ruby (`webrick`, maybe 😆 ), the tradeoff being they're not super performant. Not sure that it matters though.
- Manually add whatever packages are needed to get `puma` or `thin` to install.

Do you have any preference for which option is better here? I can definitely switch to a `-slim` image either way, though.

ahayworth: Oh whoops, we merged. We can fix it in post. 😂

cartersocha: Yeah my b haha. Please open a pr & I’ll review asap

src/emailservice/email_server.rb
```diff
@@ -0,0 +1,22 @@
+require "ostruct"
+require "pony"
+require "sinatra"
+
+require "opentelemetry/sdk"
+require "opentelemetry/exporter/otlp"
+require "opentelemetry/instrumentation/sinatra"
+
+OpenTelemetry::SDK.configure do |c|
+  c.use "OpenTelemetry::Instrumentation::Sinatra"
```
arielvalentin, line 10: I think `c.use_all` is the use that we want to steer folks towards.

src/emailservice/email_server.rb
```diff
@@ -0,0 +1,22 @@
...
+end
+
+post "/send_order_confirmation" do
+  data = JSON.parse(request.body.read, object_class: OpenStruct)
```
arielvalentin, line 14: Any change we can use a Struct or Object here instead to mitigate memory leaks?

cartersocha: Feel free to open a pr with the proposed changes. I know @ahayworth would love the help!

src/emailservice/email_server.rb
```diff
@@ -0,0 +1,22 @@
+require "ostruct"
+require "pony"
+require "sinatra"
+
+require "opentelemetry/sdk"
+require "opentelemetry/exporter/otlp"
+require "opentelemetry/instrumentation/sinatra"
```
arielvalentin, line 7: Can we use Bundler here to load dependencies and use `Bundler.require` instead of requiring each dependency separately?

src/emailservice/email_server.rb
ahayworth: Ah, good point.

src/emailservice/email_server.rb
ahayworth: I believe so, yes.

Issue #102, opened 2022-06-03
Give a simple example to use this client demo for AdService


laziobird:
**Is your feature request related to a problem?**

@mic-max  @reyang   

I found a client sample for this Java Project called "AdService" , and it is able to some steps for running  , like this 

1. build a image for AdService , for example:
```shell
cd  /${project}/opentelemetry-demo-webstore/src/adservice
docker build -t  adservice:v1  . 
```
2. run a instance for AdService by Docker, for example:
```shell
docker run -d -p 9555:9555 adservice:v1 
```
3.  enter into the docker instance and start a client demo, for example:
```shell
docker exec -it 3d6a8db7322a /bin/sh
sh /app/build/install/hipstershop/bin/AdServiceClient
```
4.  check  if  AdServer is  successful to start  by the log of this client  , for example :   
```shell
2022-06-03 17:57:11 - hipstershop.AdServiceClient - Get Ads with context camera ... trace_id= span_id= trace_flags= 
2022-06-03 17:57:12 - hipstershop.AdServiceClient - Ads: Hairdryer for sale. 50% off. trace_id= span_id= trace_flags= 
2022-06-03 17:57:12 - hipstershop.AdServiceClient - Ads: Bamboo glass jar for sale. 10% off. trace_id= span_id= trace_flags= 
```

**Describe the solution you'd like:**

So, i want to add more details for how to run a client and check the  AdService . Maybe there ? https://github.com/open-telemetry/opentelemetry-demo-webstore/blob/main/src/adservice/README.md 

**Describe alternatives you've considered.**

A suggestion is just for more people to use this Java demo easily .   What do you think? 😊

mic-max:
That's great!!!

I think these instructions also belong in `/src/adservice/README.md` and also think we should still use the `docker compose up` command for running the service though to keep it simple and not give lots of ways to run these containers, especially if (not in this case) the service depends on another service or resource.

julianocosta89:
AdService client was removed as wasn't being used in the demo.

PR #113, opened 2022-06-05
Give a simple example to use this client demo for AdService for issue #102

laziobird:
Fixes #.
[#102](https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/102)
Changes
I found a client sample for this Java Project called "AdService", and it is able to some steps for running .
 So, i start to add more details for how to run a client and check the AdService .


src/adservice/src/main/java/hipstershop/AdService.java
```diff
@@ -49,7 +52,7 @@ public final class AdService {
   private static final AdService service = new AdService();
 
   private void start() throws IOException {
-    int port = Integer.parseInt(System.getenv().getOrDefault("PORT", "9555"));
+    int port = Integer.parseInt(System.getenv().getOrDefault("PORT", AD_SERVER_PORT+""));
```
wph95, line 55: I personally find this code rather strange.
Suggest default service port use string 9555.
Thus avoiding int conversion. `AD_SERVER_PORT+""`

```suggestion
    int port = Integer.parseInt(System.getenv().getOrDefault("PORT", AD_SERVER_PORT));
```

src/adservice/src/main/java/hipstershop/AdService.java
```diff
@@ -40,6 +40,9 @@ public final class AdService {
 
   private static final Logger logger = LogManager.getLogger(AdService.class);
 
+  /** Ads Server Default Port */
+  final static int AD_SERVER_PORT = 9555;
```
wph95, line 44:
```suggestion
  final static int DEFAULT_AD_SERVER_PORT = 9555;
```

src/adservice/src/main/java/hipstershop/AdService.java
laziobird: @wph95   
Do you really complie this code ? 
```
int port = Integer.parseInt(System.getenv().getOrDefault("PORT", AD_SERVER_PORT));
```
It will cause some errors 
```
Description	                                                        
The method getOrDefault(Object, String) in the type Map<String,String> is not applicable for the arguments (String, int)	
Resource  Path       Location                                   Type
AdService.java	/ad/src/main/java/hipstershop	line 55	   Java Problem
```
Maybe you are not familiar with Java Language ☺
Note :  The method ` System.getenv().getOrDefault ` requires  a String parameter ,  but The type `AD_SERVER_PORT` is int and need to transfer

mic-max (approved):
Hey Tim, ya the instructions look good thanks! I was wondering if we could use docker compose build and docker compose up commands instead though, then the port will be loaded from the .env and then the client could also use that same variable

cartersocha:
What's the benefit of running the ad service locally?

mic-max:
> sample for this Java Project called "AdService", and it is able to some steps for running .
>  So, i start to add more details for how to run a client and check the AdService

This PR is adding instructions to use the ad service's client that was included from the google repository.
It's used to make direct requests to the service without setting up the whole project.

We have other similar clients

- https://github.com/open-telemetry/opentelemetry-demo-webstore/blob/main/src/currencyservice/client.js
- https://github.com/open-telemetry/opentelemetry-demo-webstore/blob/main/src/recommendationservice/client.py

mic-max:
@laziobird Could you please fix the blocking CI checks

cartersocha:
@laziobird could you update the branch & rerun the checks? We should be ready to merge if they pass

PR #119, opened 2022-06-06
Cleanup Load Generator Service

mic-max:
## Changes

Cleanup load generator `Dockerfile` and `requirements.txt` rename


src/loadgenerator/Dockerfile
```diff
@@ -12,23 +12,12 @@
...
+FROM python:3.10-slim AS base
 
-RUN apt-get -qq update \
-    && apt-get install -y --no-install-recommends \
-        g++ \
-        python3-dev
+WORKDIR /usr/src/app/
 
-COPY requirements.txt .
+COPY ./src/loadgenerator/ ./
```
julianocosta89, line 19: This line doesn't work with the current `compose.yaml` file.

The following change is necessary:
```yaml
  loadgenerator:
    build:
      context: ./
      dockerfile: ./src/loadgenerator/Dockerfile
```

julianocosta89 (changes requested):

src/loadgenerator/Dockerfile
mic-max: Thanks, good catch! I guess I forgot to build this one before publishing 😓

julianocosta89 (approved):
LGTM👍🏽

PR #120, opened 2022-06-06
Move `otelcol-config.yml` to root of repo

mic-max:
I think this is a better place for this file, let me know if you have any concerns or better suggestions

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -22,7 +22,7 @@ services:
     image: otel/opentelemetry-collector
     command: [ "--config=/etc/otelcol-config.yml" ]
     volumes:
-      - ./src/otelcollector/otelcol-config.yml:/etc/otelcol-config.yml
+      - ./otelcol-config.yml:/etc/otelcol-config.yml
     ports:
       - "4317"
     depends_on:
--- a/otelcol-config.yml
+++ b/otelcol-config.yml
@@ -16,6 +16,6 @@ processors:
 service:
   pipelines:
     traces:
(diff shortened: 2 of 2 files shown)


cijothomas (approved):
the otelcollector directory can be deleted now right?

compose.yml
```diff
@@ -22,7 +22,7 @@ services:
     image: otel/opentelemetry-collector
     command: [ "--config=/etc/otelcol-config.yml" ]
     volumes:
-      - ./src/otelcollector/otelcol-config.yml:/etc/otelcol-config.yml
+      - ./otelcol-config.yml:/etc/otelcol-config.yml
```
cijothomas, line 25: i think you can keep ./src part.

cijothomas: to avoid too many files in the root

mic-max:
> the otelcollector directory can be deleted now right?

Ya it's empty, no real notion of deleting folders in Git though, empty folders unless noted in a `.gitkeep` are just ignored

compose.yml
julianocosta89: I'm with @cijothomas in this one.
I'd rather keep the otelcol-config within the otelcol folder.

Any reason you would like to move to the root @mic-max ?

mic-max: It just felt like a more top level config similar to the docker compose file. I'm fine to close this PR though and leave as is too - just wanted to spur a little discussion :)

PR #128, opened 2022-06-11
github_action checks add job  markdown-link-check

### LINKED ISSUE #90: Improve the CI to block broken links
https://github.com/open-telemetry/opentelemetry-demo-webstore/pull/86#issuecomment-1145270046

It seems the current CI job doesn't prevent broken links. At minimum we should have gated CI job to prevent any relative link that is broken. Whether we should check absolute links or not is still to be decided due to the fragile nature.

One option is docfx, as been used in https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/.github/workflows/docfx.yml.
Another option is https://github.com/open-telemetry/opentelemetry-specification/blob/9abbdd39d0b35f635f833f072013431da419894e/.github/workflows/checks.yaml#L36, which can be configured to cover relative/absolute/both.

wph95:
Fixes #90

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -54,7 +54,7 @@ jobs:
     - name: run docfx
       run: docfx build --dry-run
 
-  markdown-link-check:
+  markdownlinkcheck:
     name: Broken Links
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -58,10 +58,10 @@ jobs:
     name: Broken Links
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


cartersocha:
Small nit but lets lowercase broken links like the other check names. @reyang is there a particular naming convention we should follow?

cartersocha:
But LGTM. Seems like the check is already showing on this pr

.github/workflows/checks.yml
```diff
@@ -53,3 +53,15 @@ jobs:
 
     - name: run docfx
       run: docfx build --dry-run
+
+  markdown-link-check:
```
wph95, line 57:
```suggestion
  markdownlinkcheck:
```

cartersocha (approved):
🦑

PR #129, opened 2022-06-13
Rust shipping service

### LINKED ISSUE #35: Reimplement 'shippingservice' in Rust
**Objective**

Re-implement the 'shippingservice' in Rust.

GaryPWhite:
Fixes #35 

## Changes
ShippingService is now written in Rust -- providing a demo / reference implementation for rustaceans. 


GaryPWhite:
I have some admin work for Wayfair this week that means I may not make much progress. I plan to write the docker config and OTel integrations now that this is minimally functional, then button up the implementation like the Go implementation to actually make quotes / tracking ID's.

GaryPWhite:
ok I will just let the pr lapse if it's going to worry about it

cartersocha:
@reyang can we update the bot to be less aggressive on prs or is this our preferred process?

cartersocha:
Keeping a shortlist with the ability to re-open whenever sounds good to me. I’d prefer to just follow normal best practices in general but wanted to check the existing reason

GaryPWhite:
totally reasonable @reyang . Wonder how you might feel about "draft" pr's being treated differently.

GaryPWhite:
> If the goal is to allow "draft" PR sitting there forever, I'll need to be convinced 😄

Definitely not the goal to have something sit forever :) It seems like you're looking at the PR queue as something that needs attention immediately when the PR is opened -- I'm keeping this open while I'm working on it so folks can, at their absolute leisure, check in if they're curious, or if they think i'm missing something. It seems to me that it makes more sense to open PR's when they're ready for review here -- and drafts are not necessarily preferred.

I'll open a full PR when it's ready.

PR #141, opened 2022-06-15
add metric links to frontend readme

cartersocha:
Updated standard feature tables to have more metric information with links

### DIFF
--- a/src/frontend/README.md
+++ b/src/frontend/README.md
@@ -23,7 +23,7 @@ else.
   Libraries](https://opentelemetry.io/docs/concepts/instrumenting-library/):
   :100:
 - [Manual Span
-  Creation](https://opentelemetry.io/docs/concepts/instrumenting/#manual-instrumentation):
+  Creation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md#manual-instrumentation)):
   :red_circle:
 - [Span Data
   Enrichment](https://opentelemetry.io/docs/instrumentation/net/manual/#add-tags-to-an-activity):
@@ -44,11 +44,13 @@ else.
   Operator)](https://github.com/open-telemetry/opentelemetry-operator):
   :red_circle:
 - [Instrumentation
-  Libraries](https://opentelemetry.io/docs/concepts/instrumenting-library/): :red_circle:
-- [Manual Metric
-  Creation](https://opentelemetry.io/docs/concepts/instrumenting/#manual-instrumentation):
+  Libraries](https://opentelemetry.io/docs/concepts/instrumenting-library/):
   :red_circle:
-- [Collector Agent Metric Transformation](https://opentelemetry.io/docs/collector/deployment/#agent)::red_circle:
+- [Manual Metric
+  Creation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md#manual-instrumentation)):
+  :red_circle:
+- [Collector Agent Metric
+  Transformation](https://opentelemetry.io/docs/collector/deployment/#agent)::red_circle:
 - [Push
   Metrics](https://opentelemetry.io/docs/reference/specification/metrics/sdk/#push-metric-exporter):
   :red_circle:
--- a/src/frontend/README.md
+++ b/src/frontend/README.md
@@ -44,14 +44,25 @@ else.
 
 ### Metrics
 
-- Automatic Instrumentation (Agent): :large_blue_diamond:
-- Automatic Instrumentation (OTel Operator): :red_circle:
-- Instrumentation Libraries  : :red_circle:
-- Manual Metric Creation:  :red_circle:
-- Collector Metric Transformation::red_circle:
-- Push Metrics: :red_circle:
-- SLO Metrics: :red_circle:
-- Multiple Metric Instruments: :red_circle:
+- [Automatic Instrumentation (Language
+  Agent)](https://github.com/open-telemetry/opentelemetry-java-instrumentation#about):
+  :large_blue_diamond:
+- [Automatic Instrumentation (OTel
+  Operator)](https://github.com/open-telemetry/opentelemetry-operator):
+  :red_circle:
+- [Instrumentation
+  Libraries](https://opentelemetry.io/docs/concepts/instrumenting-library/): :red_circle:
+- [Manual Metric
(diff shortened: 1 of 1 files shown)


src/frontend/README.md
```diff
@@ -44,14 +44,25 @@ else.
...
-- Automatic Instrumentation (Agent): :large_blue_diamond:
-- Automatic Instrumentation (OTel Operator): :red_circle:
-- Instrumentation Libraries  : :red_circle:
-- Manual Metric Creation:  :red_circle:
-- Collector Metric Transformation::red_circle:
-- Push Metrics: :red_circle:
-- SLO Metrics: :red_circle:
-- Multiple Metric Instruments: :red_circle:
+- [Automatic Instrumentation (Language
+  Agent)](https://github.com/open-telemetry/opentelemetry-java-instrumentation#about):
```
reyang, line 48: It's a bit strange to have the frontend (which is implemented in Golang) `README.md` link to a Java related doc.

reyang (approved):
We need to revisit some of the links, but in general the direction looks good to me.

src/frontend/README.md
cartersocha: Go doesn't have the concept of auto instrumentation using a go agent so I figured it made more sense to link to the concept in general. Feel free to propose a better link

src/frontend/README.md
```diff
@@ -44,14 +44,25 @@ else.
...
+  :large_blue_diamond:
+- [Automatic Instrumentation (OTel
+  Operator)](https://github.com/open-telemetry/opentelemetry-operator):
+  :red_circle:
+- [Instrumentation
+  Libraries](https://opentelemetry.io/docs/concepts/instrumenting-library/): :red_circle:
+- [Manual Metric
+  Creation](https://github.com/open-telemetry/opentelemetry-dotnet/tree/main/docs/metrics/getting-started#getting-started-with-opentelemetry-net-in-5-minutes):
```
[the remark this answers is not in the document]
cartersocha: Go's documentation isn't very comprehensive. I can link to straight code samples but there's no readme markdowns documenting what is happening

src/frontend/README.md
puckpuck: This is just passing the buck down the line, but I feel like this is a concept that should be defined within the core OpenTelemetry docs themselves.  Perhaps in this section: https://opentelemetry.io/docs/concepts/instrumenting/#automatic-instrumentation

src/frontend/README.md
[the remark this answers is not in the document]
puckpuck: OpenTelemetry docs has Instrumenting Library well defined here.  Can we use this https://opentelemetry.io/docs/concepts/instrumenting-library/

src/frontend/README.md
cartersocha: Yeah it'd be nice to point to generic definitions of items. It's easier for things like instrumentation libraries

reyang: And I would suggest that we remove "(Language Agent)" because 1) it's not defined 2) it is super confusing.

cartersocha: I personally find just Agent more confusing. Is the agent the collector, is it a language agent like Java etc.

reyang: My understanding is that there is no such concept called "language agent" in OpenTelemetry.

The language agnostic concept is called "automatic instrumentation" and is defined here https://opentelemetry.io/docs/concepts/instrumenting/#automatic-instrumentation.
For Java, a special technology is used based on JVM agent, and some vendors call it "Java Agent" due to historical or whatever reasons.

src/frontend/README.md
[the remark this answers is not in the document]
reyang: I would suggest that we change this line to:

```[Manual Instrumentation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md#manual-instrumentation)```

File: src/frontend/README.md (hunk 1 above)
Comment (cartersocha) [reply src/frontend/README.md]: Updated the links to be more generic. I see the java agent blurb under that doc so we can reference that other places. I removed features that weren't relevant in general.

File: src/frontend/README.md (hunk 2 above)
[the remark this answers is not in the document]
Comment (cartersocha) [reply src/frontend/README.md]: updated

**Comment by cartersocha** [comment]
Wasn't sure what to put for interprocess context prop but I think I have generic links for everything else. @puckpuck, @reyang. I'll merge tomorrow morning unless we have other suggestions

File: src/frontend/README.md [hunk 3]
```diff
@@ -44,14 +40,24 @@ else.
...
+  Creation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md#manual-instrumentation)):
+  :red_circle:
+- [Collector Agent Metric
+  Transformation](https://opentelemetry.io/docs/collector/deployment/#agent)::red_circle:
+- [Push
+  Metrics](https://opentelemetry.io/docs/reference/specification/metrics/sdk/#push-metric-exporter):
+  :red_circle:
+- [SLO Metrics](https://github.com/openslo/openslo#slo): :red_circle:
+- [Multiple Metric
+  Instruments](https://opentelemetry.io/docs/reference/specification/metrics/api/#synchronous-and-asynchronous-instruments):
```
reyang, line 59: Is this an expansion of Ln49 (Manual Metric Creation)? If yes, consider rewording it to "Manual instrumentation using multiple instruments"

src/frontend/README.md
```diff
@@ -44,14 +40,24 @@ else.
...
 
-- Automatic Instrumentation (Agent): :large_blue_diamond:
-- Automatic Instrumentation (OTel Operator): :red_circle:
-- Instrumentation Libraries  : :red_circle:
-- Manual Metric Creation:  :red_circle:
-- Collector Metric Transformation::red_circle:
-- Push Metrics: :red_circle:
-- SLO Metrics: :red_circle:
-- Multiple Metric Instruments: :red_circle:
+- [Automatic Instrumentation (OTel
```
reyang, line 43: If this is based on the OTel Operator, maybe it should be moved to the top level of the project (because individual project from the developer perspective should have no knowledge of the operator).

src/frontend/README.md
cartersocha: Updated to Multiple Manual Metric Instruments

PR #147, opened 2022-06-17
update feature table for checkout service

cartersocha:
adding default feature table to checkout service readme

### DIFF
--- a/src/checkoutservice/README.md
+++ b/src/checkoutservice/README.md
@@ -5,3 +5,49 @@ Run the following command to restore dependencies to `vendor/` directory:
 ```sh
 dep ensure --vendor-only
 ```
+
+## OpenTelemetry features
+
+### Emoji Legend
+
+- Completed: :100:
+- Not Present (Yet): :red_circle:
+
+### Traces
+
+- [Instrumentation
+  Libraries](https://opentelemetry.io/docs/concepts/instrumenting-library/):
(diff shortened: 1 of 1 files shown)


src/checkoutservice/README.md
```diff
@@ -5,3 +5,49 @@ Run the following command to restore dependencies to `vendor/` directory:
...
+
+### Emoji Legend
+
+- Completed: :100:
+- Not Present (Yet): :red_circle:
+
+### Traces
+
+- [Instrumentation
+  Libraries](https://opentelemetry.io/docs/concepts/instrumenting-library/):
```
mviitane, line 19: Not blocking, but can't really see why this line is split into two lines. It makes it more difficult to read. The same split is repeated in most of the bullets. A single line would be better:
"- [Instrumentation  Libraries](https://opentelemetry.io/docs/concepts/instrumenting-library/):"

reyang (approved):

src/checkoutservice/README.md
austinlparker: I believe the split is due to markdown lint rules.

cartersocha: Yep on the rules! It’s just raw markdown formatting but doesn’t affect how it looks online 

I use the rewrap & markdownlint vs code extensions to do the heavy lifting.

PR #157, opened 2022-06-18
[payment] - add manual instrumentation

### LINKED ISSUE #53: Review and enhance tracing support for payment service (node.js)
The following is a list of requirements that we need to evaluate before declaring v1 for trace telemetry. These requirements are across the entire application; Not all services will meet all requirements. Determine the relevant features for this service. 

- [x] Automatic Instrumentation is being used where appropriate.
- [x] Library instrumentation is being used if automatic instrumentation is unavailable.
- [ ] Automatic and Manual Context Propagation are being demonstrated.
- [ ] Telemetry is not being sampled upfront.
- [ ] Telemetry is not being filtered upfront.
- [ ] Baggage is being set and read appropriately (i.e., baggage must be explicitly set as attributes on spans)
- [ ] Update the [centralized tracing doc](https://github.com/open-telemetry/opentelemetry-demo/blob/main/docs/trace_service_features.md)

Referencing: https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/42

Service: https://github.com/open-telemetry/opentelemetry-demo-webstore/blob/main/src/paymentservice/README.md

puckpuck:
Fixes #53 

## Changes

Adds attributes to auto instrumented spans, and creates a new span with attributes.
If a bad card is used, `span.recordException()` is called from the auto-instrumented span, which ultimately creates a Span event.


src/paymentservice/charge.js
```diff
@@ -15,6 +15,8 @@
 const cardValidator = require('simple-card-validator');
 const { v4: uuidv4 } = require('uuid');
 const pino = require('pino');
+const opentelemetry = require('@opentelemetry/api');
+const tracer = opentelemetry.trace.getTracer("");
```
julianocosta89, line 19: Is it a good practice to leave the trace unnamed?

julianocosta89 (approved):
Just a small comment but not a blocker.

I've also tested locally with and without errors on the charge function.
LGTM!

src/paymentservice/charge.js
puckpuck: I'm actually not certain about this. I know you can have more than 1 tracer, though I'm not clear on the use cases to have more than 1 tracer. 

I believe if you pass in an empty name you will get the default tracer. If I'm always going to get the default tracer with a blank string, I would prefer to promote that as a practice, rather than a name, which could be prone to a typo.

julianocosta89: > if you pass in an empty name you will get the default tracer

If that's true, I'm ALL IN!

I have never seen that info in any documentation though. Who could confirm that for us?

puckpuck: I did some more digging into this, and I found that we should probably specify a name. Though it is true a default tracer will be returned by all SDKs, some will also log an error if you pass an empty name, and I don't think that's an experience we want people to have.

I think we should change this to use the service name for each one instead. I may need to go back and update some of the services we already did, so this may be better served as a separate Issue and PRs to track against.

julianocosta89: Awesome @puckpuck!

cartersocha:
Let’s make sure the feature table & service tracing issue reflect the changes

Issue #55, opened 2022-05-30
Review and enhance tracing support for email service (Ruby) 


cartersocha:
The following is a list of requirements that we need to evaluate before declaring v1 for trace telemetry. These requirements are across the entire application; Not all services will meet all requirements. Determine the relevant features for this service. 

- [x] Automatic Instrumentation is being used where appropriate.
- [x] Library instrumentation is being used if automatic instrumentation is unavailable.
- [x] Services extend automatic instrumentation.
  - [x] New attributes/events attached to existing spans.
  - [x] New spans are being created from existing spans.
- [ ] Automatic and Manual Context Propagation are being demonstrated.
- [x] Telemetry is not being sampled upfront.
- [x] Telemetry is not being filtered upfront.
- [ ] Baggage is being set and read appropriately (i.e., baggage must be explicitly set as attributes on spans)
- [ ] Update the [centralized tracing doc](https://github.com/open-telemetry/opentelemetry-demo/blob/main/docs/trace_service_features.md)

Referencing: https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/42

Service: https://github.com/open-telemetry/opentelemetry-demo-webstore/blob/main/src/emailservice/README.md

Ruby dependent on this item: https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/37

ahayworth:
@cartersocha since I can't check off the boxes myself, here's what could be updated right away:
- ✅ `Automatic Instrumentation is being used where appropriate.`
  - The demo installs automatic instrumentation for Sinatra in [the configuration block](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/emailservice/email_server.rb#L9-L11). Because the ruby app is so simple, there's really nothing else to instrument automatically. 😄 
- ✅ `Library instrumentation is being used if automatic instrumentation is unavailable.`
  - We have automatic instrumentation available, so we should satisfy this condition. I'm not sure what "Library instrumentation" is in this case anyways.
- ❓ `Telemetry is not being sampled upfront.`
  - Can you clarify what you mean by "sampled"? In the ruby SIG, we use the term "sampled" the way the SDK spec uses it - to mean "this span will be exported."
  - The way that the SDK uses this term in combination with this TODO item makes me think the requirement is that the ruby app should **not** be exporting spans by default?
[...]

PR #158, opened 2022-06-19
Add manual instrumentation to emailservice

julianocosta89:
Towards #55 .

## Changes

Adds attributes to auto instrumented spans, and creates a new span with attributes.
If sending the email fails (what doesn't happen at the moment), `span.recordException()` is called from the auto-instrumented span, which ultimately creates a Span event.

### DIFF
--- a/src/emailservice/email_server.rb
+++ b/src/emailservice/email_server.rb
@@ -12,11 +12,40 @@ end
 
 post "/send_order_confirmation" do
   data = JSON.parse(request.body.read, object_class: OpenStruct)
-  Pony.mail(
-    to:       data.email,
-    from:     "noreply@example.com",
-    subject:  "Your confirmation email",
-    body:     erb(:confirmation, locals: { order: data.order }),
-    via:      :logger
-  )
+
+  # get the current auto-instrumented span
+  current_span = OpenTelemetry::Trace.current_span
+  current_span.add_attributes({
+    "app.email.order_id" => data.order.order_id,
+    "app.email.shipping_tracking_id" => data.order.shipping_tracking_id,
+    "app.email.shipping_cost.currency" => data.order.shipping_cost.currency_code,
+    "app.email.shipping_cost" => data.order.shipping_cost.units.to_s + "." + 
+      data.order.shipping_cost.nanos.to_s
+  })
+
+  send_email(data)
+
+  rescue Exception => e
+    # record exception in span (will create a span event)
+    current_span.record_exception(e)
+    raise e
+end
+
+def send_email(data)
+  # create and start a manual span
+  tracer = OpenTelemetry.tracer_provider.tracer('')
+  tracer.in_span("send_email") do |span|
+    Pony.mail(
+      to:       data.email,
+      from:     "noreply@example.com",
+      subject:  "Your confirmation email",
+      body:     erb(:confirmation, locals: { order: data.order }),
+      via:      :logger
+    )
+    span.set_attribute("app.email.sent", true)
+  end
+  # manually created spans need to be ended
+  # in Ruby, the method `in_span` ends it automatically
+  # check out the OpenTelemetry Ruby docs at: 
+  # https://opentelemetry.io/docs/instrumentation/ruby/manual/#creating-new-spans 
 end


julianocosta89:
@ahayworth also added you as a reviewer.
Would you mind taking a look? 👀

src/emailservice/email_server.rb
```diff
@@ -12,11 +12,40 @@
...
-    from:     "noreply@example.com",
-    subject:  "Your confirmation email",
-    body:     erb(:confirmation, locals: { order: data.order }),
-    via:      :logger
-  )
+
+  # get the current auto-instrumented span
+  current_span = OpenTelemetry::Trace.current_span
+  current_span.add_attributes({
+    "app.email.order_id" => data.order.order_id,
```
austinlparker, line 19: What generates the order ID? imo this should be normalized across services; app.order.id perhaps?

julianocosta89: app.order.id sounds good!
That's the name used in the [demo.pb](https://github.com/open-telemetry/opentelemetry-demo-webstore/blob/7c877d652ddb89070c9465b3f99fe17a1ea62a2c/src/shippingservice/genproto/hipstershop/demo.pb.go#L1406)

Will update it.

cartersocha:
LGTM

ahayworth:
@julianocosta89 @cartersocha Yes I will take a look this afternoon - I am busy this morning, but I have a suggestion or two and I'll write those up in a bit. 😄

src/emailservice/email_server.rb
```diff
@@ -12,11 +12,40 @@
...
+      data.order.shipping_cost.nanos.to_s
+  })
+
+  send_email(data)
+
+  rescue Exception => e
+    # record exception in span (will create a span event)
+    current_span.record_exception(e)
+    raise e
+end
```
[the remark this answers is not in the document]
ahayworth: However, if you'd prefer to rescue _just_ this route, I would ask that we align the indentation like so:

```suggestion

rescue Exception => e
  # record exception in span (will create a span event)
  current_span.record_exception(e)
  raise e
end
```

src/emailservice/email_server.rb
```diff
@@ -12,11 +12,40 @@
...
+
+def send_email(data)
+  # create and start a manual span
+  tracer = OpenTelemetry.tracer_provider.tracer('')
```
ahayworth, line 36: Providing the empty string here is equivalent to not passing any arguments at all, and in both cases we log a warning when this happens:

```
WARN -- : calling TracerProvider#tracer without providing a tracer name.
```

(as per spec, if I recall). So we should give our tracer a name:

```suggestion
  tracer = OpenTelemetry.tracer_provider.tracer('emailservice')
```

ahayworth (commented):
Thanks for adding this! After thinking about it, I would want to request a few changes before we merge this in. They're a little bit nit-picky, but since this is a reference application I felt maybe it was appropriate. 😄 

(Of course, these are "strong opinions, weakly held" and I won't be upset if you choose not to take them. We also really appreciate the contribution! ❤️ )

src/emailservice/email_server.rb
```diff
@@ -12,11 +12,40 @@
...
-    via:      :logger
-  )
+
+  # get the current auto-instrumented span
+  current_span = OpenTelemetry::Trace.current_span
+  current_span.add_attributes({
+    "app.order.id" => data.order.order_id,
+    "app.shipping.tracking.id" => data.order.shipping_tracking_id,
+    "app.shipping.cost.currency" => data.order.shipping_cost.currency_code,
+    "app.shipping.cost" => data.order.shipping_cost.units.to_s + "." +
```
mic-max, line 22: We should note down all the custom attributes we add to services somewhere :)

mic-max (approved):
LGTM - I agree with @ahayworth's comments as well

src/emailservice/email_server.rb
puckpuck: > Regarding the tracer with an empty name I want to bring @puckpuck to the thread, because i believe it would be better if we follow the same approach in all services.

I did some digging into this, and came to the conclusion we should use a name for each tracer since some SDKs will log an error or warning when an empty string is used.  I updated the other PRs to use the service name when getting the default tracer. I think we should accept this suggestion and use `emailservice` here.

src/emailservice/email_server.rb
julianocosta89: I agree, any suggestions on where @mic-max ?

julianocosta89:
@ahayworth this one should be good to go now.
Would you mind taking a 2nd look?

src/emailservice/email_server.rb
puckpuck: We can start with a custom-instrumentation.md file in the docs folder. I imagine a table for each type of instrumentation (traces, metrics, logs). The traces table would contain a list of the attributes, a short description, and which service(s) you can find it in.  The metrics table would contain a list of custom metrics with descriptions and services found.

julianocosta89:
@cartersocha this one should be good to go 🤩

PR #160, opened 2022-06-20
add prometheus service into docker compose and add a basic metric to frontend

### LINKED ISSUE #70: Review and enhance metric support for frontend (Go)
The following are requirements for metrics in the project. Not all services will meet all requirements.

- [ ] Service metrics should not be duplicative of trace-derived metrics.
- [ ] Services should emit multiple business metrics where appropriate.
- [ ] Services should emit trace exemplars for their metrics.
- [ ] Services should use push metrics where possible.
- [ ] Services should explicitly read baggage where needed.
- [ ] Collectors should perform metric transforms to normalize resource and service metrics.

Referencing issue: https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/43

fatsheep9146:
add prometheus service into docker compose and add a basic metric to frontend

Fixes #70 

## Changes

1. add prometheus service into docker compose
2. add a basic metric `http.server.request_count` to frontend 

I attempted to add metric support to webstore demo in my own mind.  The implementation may be too simple. I only add one metric for frontend service and add a prometheus service to scrape metric from otel-collector. I really hope the comminuty can give more advises, for examples,

- add more metrics
- add more metric visualization tool, such as grafana


fatsheep9146:
add some screen shot about the metric scraped by prometheus.

image

image

fatsheep9146:
@cartersocha @puckpuck @julianocosta89 Could you help review this ?

julianocosta89:
@fatsheep9146 could you take a look at the yaml lint and sanity check issues?
Regarding the PR itself it looks good, but I do not have much experience with metrics, maybe some of the other approvers have more experience, so I'd rather wait a bit and see if we get any insights from them.

cartersocha:
FYI Austin & I are at OTel community day on Monday & won’t be able to take a look until later this week 

thanks for making progress on this tho @fatsheep9146. It’s an area that needs a lot more attention

reyang:
Regarding the overall direction of metrics: https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/70#issue-1252987147 mentioned that "services should use push metrics where possible" which sounds like OTLP exporter scenario?

Here is my understanding:
```text
                       +-------------------------+
SDK + OTLP Exporter ---> OpenTelemetry Collector <----- Prometheus Scraper ----> TSDB
                       +------------+------------+
                                    |
                                    +---------> Vendor Ingestion
```

fatsheep9146:
> exporter



> Regarding the overall direction of metrics: [#70 (comment)](https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/70#issue-1252987147) mentioned that "services should use push metrics where possible" which sounds like OTLP exporter scenario?
> 
> Here is my understanding:
> 
> ```
>                        +-------------------------+
> SDK + OTLP Exporter ---> OpenTelemetry Collector <----- Prometheus Scraper ----> TSDB
>                        +------------+------------+
>                                     |
>                                     +---------> Vendor Ingestion
> ```


Yes, this is same as my understanding.

src/otelcollector/otelcol-config.yml
```diff
@@ -9,6 +9,9 @@ exporters:
     tls:
       insecure: true
   logging:
+  prometheus:
+    endpoint: 0.0.0.0:8889
```
reyang, line 13: Maybe use the default port? https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#prometheus-exporter

src/frontend/middleware.go
```diff
@@ -35,6 +38,11 @@ type logHandler struct {
 	next http.Handler
 }
 
+var (
+	meter                 = global.MeterProvider().Meter("frontend")
+	httpRequestCounter, _ = meter.SyncInt64().Counter("http.server.request_count")
```
reyang, line 43: Consider duration histogram which is more interesting (and if folks only want the count, they can configure the SDK) https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#http-server.

src/prometheus/prometheus-config.yaml
```diff
@@ -0,0 +1,10 @@
+global:
+  evaluation_interval: 30s
+  scrape_interval: 30s
```
reyang, line 3: Do we think 30s scraping interval is too long for the demo? Maybe 5s to make it more interactive?

src/frontend/main.go
```diff
@@ -96,6 +103,37 @@ func InitTracerProvider() *sdktrace.TracerProvider {
 	return tp
 }
 
+func InitMeterProvider() (metric.MeterProvider, error) {
+	ctx := context.Background()
+
+	client := otlpmetricgrpc.NewClient(
+		otlpmetricgrpc.WithInsecure(),
+		otlpmetricgrpc.WithEndpoint("otelcol:4317"))
```
reyang, line 111: Does the Go SDK provide a default port (4317)?

src/otelcollector/otelcol-config.yml
fatsheep9146: Done, I set endpoint to 0.0.0.0:9464 now, since in the configuration of prometheus exporter, the endpoint info must be specified, otherwise it will not start

src/prometheus/prometheus-config.yaml
fatsheep9146: I changed to 5s.

src/frontend/main.go
fatsheep9146: I added the environment `OTEL_EXPORTER_OTLP_ENDPOINT` to service frontend, the metrics sdk will use this environment by default.

src/frontend/middleware.go
fatsheep9146: I add `http.server.duration` metric, but I think this version is literally simple one, much more optimization can be added in the successive prs, for examples,

- import net/http related instrumentation library. https://github.com/open-telemetry/opentelemetry-go-contrib/tree/main/instrumentation/net/http/otelhttp
- add more app related attributes

fatsheep9146: here is a screenshot of raw bucket metric in prometheus, `http_server_duration_bucket`

http://127.0.0.1:9090/graph?g0.expr=http_server_duration_bucket&g0.tab=1&g0.stacked=0&g0.show_exemplars=0&g0.range_input=1h

image

src/otelcollector/otelcol-config.yml
```diff
@@ -3,12 +3,16 @@ receivers:
...
       http:
+
 exporters:
   jaeger:
     endpoint: "jaeger:14250"
     tls:
       insecure: true
   logging:
+  prometheus:
+    endpoint: 0.0.0.0:9464
```
mic-max, line 14:
```suggestion
    endpoint: "otelcol:9464"
```
nit: To use the same notation as above (if this works)

docker-compose.yml
```diff
@@ -251,3 +253,19 @@ services:
...
+      - --web.console.libraries=/etc/prometheus/console_libraries
+      - --storage.tsdb.retention.time=1h
+      - --config.file=/etc/prometheus/prometheus-config.yaml
+      - --storage.tsdb.path=/prometheus
+      - --web.enable-lifecycle
+      - --web.route-prefix=/
+    volumes:
+      - ./src/prometheus/prometheus-config.yaml:/etc/prometheus/prometheus-config.yaml
+    ports:
+      - "9090:9090"
```
mic-max, line 271: Do we want to define a port for this and use it from the `.env` file?

mic-max (approved):

wph95 (approved):
LGTM

austinlparker:
I think prometheus is probably ok -- realistically, what are our options here in terms of a sane default? I think the req behind the req is that we need to demo both pull and push metrics; the actual place they get sent isn't really that important from our perspective. An end-user should be able to swap out the prom exporter with any other OTLP metrics endpoint and it'd just work, right?

reyang:
> I think prometheus is probably ok -- realistically, what are our options here in terms of a sane default? I think the req behind the req is that we need to demo both pull and push metrics; the actual place they get sent isn't really that important from our perspective. An end-user should be able to swap out the prom exporter with any other OTLP metrics endpoint and it'd just work, right?

I would suggest Prometheus as the TSDB (Timeseries Database) and Grafana as the UI. Here is one example in OpenTelemetry C++ and I found Grafana UI to be much more usable than Prometheus UI https://github.com/open-telemetry/opentelemetry-cpp/blob/a847d0ce42209e458f8f98a5751626a33adfad95/examples/prometheus/README.md.

src/otelcollector/otelcol-config.yml
fatsheep9146: Done, I tested, it works.

image

src/otelcollector/otelcol-config.yml
```diff
@@ -19,3 +23,7 @@ service:
       receivers: [otlp]
       processors: [batch]
       exporters: [logging, jaeger]
+    metrics:
+      receivers: [otlp]
+      processors: []
```
puckpuck, line 28: Should the `batch` processor be specified here as a good practice for all pipelines?

puckpuck (approved):
1 small nit pic but LGTM with the histogram metric.

src/otelcollector/otelcol-config.yml
fatsheep9146: I think you are right, the [doc](https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/batchprocessor) of batch processor said that:

> It is highly recommended to configure the batch processor on every collector.

fatsheep9146: Done

fatsheep9146:
I think all conversations are resolved, and ready to be merged. :) @cartersocha

fatsheep9146:
> Ok merging now. Let’s get a basic grafana dashboard running soon to consume this. I think this is using Prometheus’s front end still but no need to block.

I will work on this right now

cartersocha:
> > @fatsheep9146 could you separately update the change log to reflect metrics being added. I think this qualifies as significant enough
> 
> Absolutely, except changelog , I think I should also update the ReadMe of the Webstore and the frontend service?

Yeah let’s flesh out the documentation. Maybe we set up dedicated signal docs that holds all the links to various implementations & the consumption / storage details too? 

In each service’s readme we can keep the generic feature table then some code highlights

cartersocha:
No need to start this structure now but something for us to consider in the next couple weeks as we clean stuff up

aengusrooneygrafana:
@fatsheep9146 the dashboard provider is missing.  Will I raise an issue or a PR to fix this?  You need to add a block for your dashboard provider, in addition to providing your dashboard.json.

PR #165, opened 2022-06-23
update emoji / add payment feature table

cartersocha:
change emoji to improve readability. 

Add payment feature table


reyang (approved):
LGTM. I have a general suggestion - could we move the emoji to the beginning of the line?

Currently it looks like:

image

I personally find it more readable if it looks like:

image

PR #167, opened 2022-06-23
Clean up FeatureFlagService

maciej-szlosarczyk:
This is a cleanup of feature flag service from unused code, unwanted dependencies and some minor UI/code standard fixes.

For now the UI looks like this, I plan to update it to use the same styles as `frontend` in a separate PR.

Screenshot 2022-06-23 at 08 38 07

## Changes

* Remove unused deps (mailer, live dashboard)
* Clean up compilation warnings
* Remove unused code
* Rename function to more Elixir-like `Featureflagservice.FeatureFlags.list_feature_flags/0`
* Remove phoenix branding from UI
* Add buttons and nicer links


tsloughter (approved):
Looks great, thanks!

tsloughter:
For anyone else reading this, I believe there is still more that can be cleaned up, so pls send PRs if you see anything :). Main stuff I'm thinking of that can probably go is stuff out of `FeatureflagserviceWeb` related to liveview and maybe some of the Plug's used in the router. For example, `fetch_session` sounds like one we don't need.

maciej-szlosarczyk:
@tsloughter I think both have to stay - LiveView is used to render .heex templates, and session plug is required to use flash messages (`put_flash(conn, :info, "Feature flag created successfully.")`).

cartersocha:
One thing to note is we're rebranding the front end. So maybe hold off on making this too aligned or detailed with the existing front end

tsloughter:
@maciej-szlosarczyk ah! Thanks, I would have broken that :)

PR #172, opened 2022-06-27
Fixing an emailservice build problem.

mviitane:
Fixing an emailservice build problem.

The build is currently failing when running
`docker-compose up`

Problem:
`Step 3/8 : COPY Gemfile* .
When using COPY with more than one source file, the destination must be a directory and end with a /
ERROR: Service 'emailservice' failed to build : Build failed`

## Changes

Dockerfile updated to fix the build.

### DIFF
--- a/src/emailservice/Dockerfile
+++ b/src/emailservice/Dockerfile
@@ -16,7 +16,7 @@ FROM ruby:3.1.2-slim
 
 RUN apt-get update -y && apt-get install -y build-essential
 
-COPY Gemfile* .
+COPY Gemfile* ./
 RUN bundle install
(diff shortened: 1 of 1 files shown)


cartersocha:
Just read through the docker files and this does look like a bug

cartersocha (approved):

cartersocha:
Thanks for verifying this @proffalken. Just merged the fix

ahayworth:
Huh - thanks for fixing everyone. Never saw this on my machine when implementing it, but I suppose that's why "works on my machine" is dangerous. 😄 ❤️

Issue #178, opened 2022-06-30
Loadgenerator failing to start in Mac M1


julianocosta89:
# Bug Report

Which version of the demo you are using?
[f8db21d7ce011c965d5edd6f058429eae7e2d4e3](https://github.com/open-telemetry/opentelemetry-demo-webstore/commit/f8db21d7ce011c965d5edd6f058429eae7e2d4e3)

## Symptom

**What is the expected behavior?**

We expect the demo to start and have load automatically.

**What is the actual behavior?**

Loadgen failing to start.

## Reproduce

`docker compose up` in a Mac with M1.

## Additional Context

Log:

```shell
> [ghcr.io/open-telemetry/demo:0.1.0-loadgenerator 4/4] RUN pip install --prefix="/usr/local/" -r ./requirements.txt:
#0 3.333 Collecting locust==2.6.1
#0 5.937   Downloading locust-2.6.1-py3-none-any.whl (800 kB)
#0 6.224      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 800.3/800.3 KB 2.9 MB/s eta 0:00:00
#0 6.400 Collecting opentelemetry-api==1.9.1
#0 6.469   Downloading opentelemetry_api-1.9.1-py3-none-any.whl (47 kB)
#0 6.478      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 47.8/47.8 KB 7.7 MB/s eta 0:00:00
#0 6.550 Collecting opentelemetry-exporter-otlp-proto-grpc==1.9.1
...
#0 51.45 Collecting zope.event
#0 51.51   Downloading zope.event-4.5.0-py2.py3-none-any.whl (6.8 kB)
#0 51.61 Collecting brotli
#0 51.84   Downloading Brotli-1.0.9-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl (2.6 MB)
#0 52.79      ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 2.6/2.6 MB 2.8 MB/s eta 0:00:00
#0 53.00 Collecting MarkupSafe>=2.0
#0 53.07   Downloading MarkupSafe-2.1.1-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl (26 kB)
[...]

PR #179, opened 2022-06-30
Shippingservice rust

### LINKED ISSUE #35: Reimplement 'shippingservice' in Rust

GaryPWhite:
Fixes #35

## Changes

remove golang implementation of shippingservice, written in rust now.

* [x] Design discussion issue #35


GaryPWhite:
@TommyCPP

GaryPWhite:
I'm aware that `genproto.sh` was removed from most projects in #94 -- I'm not sure if that's the right move here, since generating the files automatically with docker is not possible -- the build context required is not permitted by the fact the dockerfile sits two directories above where the original `proto` files live -- and generating them appropriately with `build.rs` necessitates the installation of cargo and mounting of a target folder (potentially from a different OS) into the container's build context. To build with docker, one would have to have `cargo` installed, which may defeat the purpose of using docker. Happy to discuss in detail as needed.

GaryPWhite:
tl;dr, it generates code in the same directory as the target executables -- and the filenames are pseudo-random. if we could generate protos to some directory instead, that would be great -- but as I understand that's functionally similar to what `genproto` is doing anyways.

GaryPWhite:
oh, wait, I see. We've moved everything up a couple layers so the build context works.. ok let me make some edits :)

mic-max:
> I'm aware that `genproto.sh` was removed from most projects in #94 -- I'm not sure if that's the right move here, since generating the files automatically with docker is not possible -- the build context required is not permitted by the fact the dockerfile sits two directories above where the original `proto` files live -- and generating them appropriately with `build.rs` necessitates the installation of cargo and mounting of a target folder (potentially from a different OS) into the container's build context. To build with docker, one would have to have `cargo` installed, which may defeat the purpose of using docker. Happy to discuss in detail as needed.

To generate files in the Dockerfile as other services currently do, the build context .can be set in the `docker-compose.yml` file, check how the other services are setup in that file for examples. I'm not familiar with Rust so don't know what `build.rs` is. Seems like a multi-stage Dockerfile would be best for the build. First stage to get all the tools needed and compile the proto and code. Then the second (which would be published) is just the executable essentially.

mic-max (approved):
LGTM overall, thanks for doing this!!

Did as thourough of a review as I could without knowing Rust, mostly just had some comments about the protocol buffers which I'd be happy to have addressed in a follow up PR

src/shippingservice/src/shipping_service.rs
```diff
@@ -0,0 +1,173 @@
...
+        shop::{CartItem, ShipOrderRequest},
+        ShippingServer, NANOS_MULTIPLE,
+    };
+    use tonic::Request;
+    use uuid::Uuid;
+
+    fn make_quote_request_with_items(items: Vec<i32>) -> Request<GetQuoteRequest> {
+        let cart_items: Vec<CartItem> = items.into_iter().fold(Vec::new(), |mut accum, count| {
+            accum.push(CartItem {
+                product_id: "fake-item".to_string(),
```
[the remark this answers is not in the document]
julianocosta89: in Rust we need to convert strings to string 😅 
Maybe @GaryPWhite has a better explanation, but I know from experience that this is required.

src/shippingservice/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+    let health = HealthCheckServer::default();
+
+    Server::builder()
+        .add_service(ShippingServiceServer::new(shipper))
+        .add_service(HealthServer::new(health))
+        .serve(addr)
+        .await?;
+    
+    Ok(())
+}
```
julianocosta89, line 62: QQ, aren't we required to add the following at the end of the `main` method?
`global::shutdown_tracer_provider();`

src/shippingservice/src/shipping_service.rs
```diff
@@ -0,0 +1,173 @@
...
+        let itemct: u32 = request
+            .into_inner()
+            .items
+            .into_iter()
+            .fold(0, |accum, cart_item| accum + (cart_item.quantity as u32));
+
+        // We may want to ask another service for product pricing / info
+        // (although now everything is assumed to be the same price)
+        // check out the create_quote_from_count method to see how we use the span created here
+        let q = global::tracer("shippingservice/get-quote")
```
julianocosta89, line 42: I've tested the changes locally and I was only able to see this trace in Jaeger.
Also, the span was orphan.

Do we need to manually set the context propagation in Rust?

src/shippingservice/src/shipping_service/quote.rs
```diff
@@ -0,0 +1,32 @@
...
+    pub dollars: i64,
+    pub cents: i32,
+}
+
+// TODO: Check product catalog for price on each item (will likley need item ID)
+pub fn create_quote_from_count(count: u32) -> Quote {
+    let f = if count == 0 { 0.0 } else { 8.99*(count as f64) };
+    get_active_span(|span| {
+        let q = create_quote_from_float(f);
+        span.add_event("Quote Issued", vec![KeyValue::new(format!("{}", q), true)]);
```
julianocosta89, line 16: I've re-built the image and ran the demo, but this event never showed up in Jaeger.

julianocosta89 (changes requested):
Added some comments, but the main blocker for me is the Context Propagation.
When running this version I just got a single orphan span for the shipping service.

src/shippingservice/src/shipping_service.rs
TommyCpp: Yeah we can probably extract using [`TraceContextPropagator`](https://docs.rs/opentelemetry/latest/opentelemetry/sdk/propagation/struct.TraceContextPropagator.html)

src/shippingservice/src/main.rs
TommyCpp: For services, it's optional in my mind. Calling shutdown makes sure the buffered spans exports but services are generally long-running and don't restart that often. This is more useful with something like a serverless function.

src/shippingservice/src/shipping_service.rs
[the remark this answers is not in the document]
TommyCpp: This is more of converting `&'static str` to `String` 😃

src/shippingservice/src/shipping_service.rs
TommyCpp: https://github.com/open-telemetry/opentelemetry-rust/blob/main/examples/grpc/src/server.rs#L19-L37 gives an example of how to extract propagation information from the tonic header. 

Given different grpc frameworks may store the metadata/header differently users will have to implement `Extractor` based on their grpc library.

src/shippingservice/src/shipping_service/quote.rs
GaryPWhite: i'll look into it

src/shippingservice/src/health_service.rs
```diff
@@ -0,0 +1,24 @@
+use health::health_server::{Health};
```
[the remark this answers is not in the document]
GaryPWhite: this was a mirror of the golang version -- what you're saying makes sense and I can just import the crate / use the server directly instead. Thanks for the callout :)

src/shippingservice/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+    )
+    .install_batch(opentelemetry::runtime::Tokio)
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    init_logger()?;
+    init_tracer()?;
+    info!("OTel pipeline created");
+    let port = env::var("PORT").unwrap_or_else( |_|{"50051".to_string()});
```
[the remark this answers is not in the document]
GaryPWhite: I left this in mostly so we don't have a weird experience when running / developing the server locally. env var propogation to a debug console sometimes requires fancy wiring instead of a simple grab 'n' go. Let me know what you think.

GaryPWhite:
@mic-max Re: Docker and protos: I assumed that the dockerfile should work without the docker-compose -- but it seems like the `genproto.sh` removal/change prefers that we use `docker-compose` from the base to build dockerfiles -- almost all dockerfiles are copying `./pb/` and using `context: ./` -- I hadn't considered it :) I'll switch it over now.

> To generate files in the Dockerfile as other services currently do, the build context .can be set in the `docker-compose.yml` file, check how the other services are setup in that file for examples. I'm not familiar with Rust so don't know what `build.rs` is. 
`build.rs` is what `cargo` (think maven, gradle, npm) would use pre-compilation, or to set parameters during compilation. As I was thinking about this over the long weekend I think there might be a (slightly gross) solution, so long as the `pb` directory doesn't go anywhere :)

> Seems like a multi-stage Dockerfile would be best for the build. First stage to get all the tools needed and compile the proto and code. Then the second (which would be published) is just the executable essentially.
We are using a multi-stage dockerfile -- I think the main issue here is that editing the code and actually running it without the protos in a place was a brain block for me -- I can use a solution as stated above

src/shippingservice/src/main.rs
```diff
@@ -0,0 +1,62 @@
...
+    // debug is used on lower level apis and not used here.
+}
+
+fn init_tracer() -> Result<sdktrace::Tracer, TraceError> {
+    opentelemetry_otlp::new_pipeline()
+    .tracing()
+    .with_exporter(
+        opentelemetry_otlp::new_exporter()
+            .tonic()    
+            .with_endpoint("http://otelcol:4317/v1/traces")
```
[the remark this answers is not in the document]
GaryPWhite: got it!

GaryPWhite:
noting that I've included a change in the next commit that updates `loadgenerator`'s dockerfile to use the not-`slim` image for python. M1's apparently don't like the slim version! #178 describes the issue, thanks for being patient @julianocosta89 !

src/shippingservice/src/main.rs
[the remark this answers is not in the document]
mic-max: I'm not fully following. I was suggesting the following changes.

```rust
let port = env::var("SHIPPING_SERVICE_PORT")
if port == nil { panic() }
```

Then in the `docker-compose.yml` change a line from
`- PORT=${SHIPPING_SERVICE_PORT}` to `- SHIPPING_SERVICE_PORT`

This is just a nit comment, feel free to resolve too

GaryPWhite:
@TommyCpp I used your sample (thanks!) for context propagation, I think I'm still missing something sending events over.. would really appreciate a wise set of eyes on if I'm doing this right :)

julianocosta89:
@GaryPWhite I've played around a bit with and got the Context Propagation working:
image

I'm adding the changes that I've done in the code review.

src/shippingservice/src/shipping_service.rs
```diff
@@ -0,0 +1,198 @@
+use opentelemetry::trace::TraceContextExt;
+use opentelemetry::trace::Tracer;
```
julianocosta89, line 2:
```suggestion
use opentelemetry::trace::{Span, Tracer};
```

src/shippingservice/src/shipping_service.rs
```diff
@@ -0,0 +1,198 @@
...
+        let parent_cx =
+            global::get_text_map_propagator(|prop| prop.extract(&MetadataMap(request.metadata())));
+        // in this case, generating a tracking ID is trivial
+        // we'll create a span and associated events all in this function.
+        global::tracer("shippingservice/ship-order").start_with_context("ship-order", &parent_cx);
+
+        let tid = create_tracking_id();
+        parent_cx
+            .span()
+            .add_event("Tracking ID issued", vec![KeyValue::new(tid.clone(), true)]);
```
julianocosta89, line 93:
```suggestion
        let mut span = global::tracer("shippingservice/ship-order").start_with_context("ship-order", &parent_cx);

        let tid = create_tracking_id();
        span.add_event("Tracking ID issued", vec![KeyValue::new(tid.clone(), true)]);
```

julianocosta89 (changes requested):
Suggested changes on `main .rs` are for the Context Propagation.
Suggested changes on `shipping_service.rs` are to make the events to work.

TommyCpp:
> @TommyCpp I used your sample (thanks!) for context propagation, I think I'm still missing something sending events over.. would really appreciate a wise set of eyes on if I'm doing this right :)

Sorry didn't get a chance to take a look at it during the day. But I think @julianocosta89 is right on point :smiley: 

`parent_cx` contains remote span so we want to use it as a parent of our local span

src/shippingservice/README.md
```diff
@@ -10,11 +10,10 @@ rust [here](https://www.rust-lang.org/tools/install).
 
 ## Build
 
-From `src/shippingservice`, run:
+From `../../`, run:
 
 ```sh
-./genproto.sh # ensures up-to-date protos are used
-docker build ./
+docker-compose build shippingservice
```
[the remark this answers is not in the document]
GaryPWhite: old habits :)

julianocosta89:
@GaryPWhite it seems that some files that were deleted previously got back in your last commit.
Was that expected, or they just sneaked in 😅 ?

GaryPWhite:
sneaked in. My bad.

GaryPWhite:
@julianocosta89 hey wait -- which files are we talking about here? I don't know which ones I put in 😅

julianocosta89:
> @julianocosta89 hey wait -- which files are we talking about here? I don't know which ones I put in 😅

`src/shippingservice/proto/demo.proto`
`src/shippingservice/proto/grpc/health/v1/health.proto`
`src/shippingservice/src/health_service.rs`

And I also saw that some items that you had resolved before were back.
Line the blank line at the end of file: `src/shippingservice/src/shipping_service/tracking.rs`

julianocosta89:
> > @julianocosta89 hey wait -- which files are we talking about here? I don't know which ones I put in 😅
> 
> `src/shippingservice/proto/demo.proto` `src/shippingservice/proto/grpc/health/v1/health.proto` `src/shippingservice/src/health_service.rs`
> 
> And I also saw that some items that you had resolved before were back. Like the blank line at the end of the files: `src/shippingservice/src/shipping_service/tracking.rs` `src/shippingservice/src/shipping_service/quote.rs`

Looks like I'm seeing stuff 👀

julianocosta89:
@GaryPWhite yeah...

The `getQuote` event is not showing up 😢 

image

julianocosta89 (changes requested):

src/shippingservice/src/shipping_service/quote.rs
GaryPWhite: I think I tracked something down. I noticed that `get_active_span` does... expect the span to be active :)

I specifically marked the span as active on local -- building and will update if it works!

GaryPWhite: I think that fixed it? Let me push and you see for yourself :) 
image

src/shippingservice/src/shipping_service.rs
```diff
@@ -0,0 +1,198 @@
...
+            .items
+            .into_iter()
+            .fold(0, |accum, cart_item| accum + (cart_item.quantity as u32));
+
+        // We may want to ask another service for product pricing / info
+        // (although now everything is assumed to be the same price)
+        // check out the create_quote_from_count method to see how we use the span created here
+        let tracer = global::tracer("shippingservice/get-quote");
+        let span = tracer.start_with_context("get-quote", &parent_cx);
+        let _guard = mark_span_as_active(span);
```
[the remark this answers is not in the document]
julianocosta89: awesome @GaryPWhite!!!! 🥳

julianocosta89 (commented):
LGTM 🚀 !

julianocosta89 (approved):

julianocosta89:
Thank you very much for the contribution @GaryPWhite !

TommyCpp:
Nice work and thanks for making this happens! @GaryPWhite

cartersocha:
> woohoo! Just need someone with permission to merge and we're officially in rust! Thanks for all the help everyone :)

Could you update your branch? then I'll merge

julianocosta89:
@GaryPWhite looks like there are some merge conflicts there.
Would you mind taking a look?

I think after that we are good to merge it!

GaryPWhite:
agh -- sorry for the delay! Fixed the conflict, should be good to go now!

PR #182, opened 2022-07-02
Add list of all manual attributes

julianocosta89:
With all the services being instrumented manually, it would be great if we could keep the span attributes documented.
I've went through all services and the opened PRs to create the initial  file.
From this, we can keep it updated whenever new PRs arrive, and we can follow a naming convention across all services.

### DIFF
--- a/docs/manual_span_attributes.md
+++ b/docs/manual_span_attributes.md
@@ -11,7 +11,9 @@ This document contains the list of manual Span Attributes used throughout the de
 
 ## CartService
 
-* None yet
+* `app.product.id`
+* `app.product.quantity`
+* `app.user.id`
 
 ## CheckoutService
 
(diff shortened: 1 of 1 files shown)


cartersocha:
Lgtm. Thanks for doing this

cartersocha (approved):
We could put these in the individual service readmes too but I see a benefit in centralizing too. It’s easy to go either way here.  

Centralizing makes sense for now to me. I’d prefer to check one doc than 11 readmes personally

julianocosta89:
> Centralizing makes sense for now to me. I’d prefer to check one doc than 11 readmes personally

Me too 😅

cartersocha:
Should we shift the generic feature tables to separate centralized docs too? Align it by traces, metrics, logs? Then just update in one place

julianocosta89:
Which generic feature tables?

austinlparker:
We can work on consolidating more in a separate PR -- this seems good to go, merging.

PR #188, opened 2022-07-06
feat(jaeger): Enable jaeger SPM

LinkMaq:
hey guys,  we might to be enable SPM on jaeger dashboard.
https://www.jaegertracing.io/docs/1.36/spm/

I've tested it and it works with just change a little configuration.
image

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,10 +16,12 @@ services:
       - "14269:14269"
     environment:
       - COLLECTOR_ZIPKIN_HOST_PORT
+      - METRICS_STORAGE_TYPE=prometheus
+      - PROMETHEUS_SERVER_URL=http://prometheus:9090
 
   # Collector
   otelcol:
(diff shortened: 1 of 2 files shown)


cartersocha:
Thanks for sharing! This looks interesting. 

I'm guessing these are tailbased metrics? I'm a bit worried when we add sampling it may provide a misleading picture 

.NET did just add runtime standard metrics & other languages could provide similar data through manual / library instrumentation. It might be better to add the actual instrumentation vs. a specific tool feature. 

@puckpuck, @mic-max any thoughts on this?

julianocosta89:
I personally don't like the idea of having it, that's why I didn't comment before.

We are already using Grafana and Prometheus, I believe those are the projects that most part of the community uses, so I would stick to it.

cartersocha:
Closing as this doesn’t align with our architecture. Thanks for suggesting tho!

Issue #195, opened 2022-07-08
Add E2E service tests


mic-max:
Add tests that send requests to a service and verify that a response is expected.
Planning on implementing all tests in node.js for simplicity.
This will help us make modifications to a service and know without running the entire project that the service changed still operates in as intended.
There might be other changes needed like opening the ports of the service to localhost within the `docker-compose.yml` file and adding some instruction on how to setup and run the tests, which will be part of the first PR that adds tests for one service (paymentservice).

mic-max:
Running `npm test` could be part of the CI if it also started all services for the demo...

PR #196, opened 2022-07-10
Add End-to-end testing (payment service)

mic-max:
Towards #195

## Changes

- add tests for payment service, valid and all different types of invalid credit cards (bad number, expired, not a mastercard or visa)
- open payment service port to localhost


cartersocha (approved):
LGTM for initial test

PR #197, opened 2022-07-10
Payment Service further consolidation

mic-max:
## Changes

- Simplified everything into the one `index.js` file
- Previously the charge service was replying empty objects: fixed so it sends the transaction id now. `transaction_id` -> `transactionId` this was introduced by changing the protoloader options in a previous PR, lookout for a testing PR that will prevent use from making similar regressions to this service and eventually all services.


julianocosta89 (commented):
@mic-max is this change required?

I know that the demo doesn't follow all the principles of development and everything, but the `charge.js` looks like a good example of the single-responsibility principle.

What benefits is the removal bringing to the sample?
If none, I'd rather leave it implemented as it is, just fixing the `transaction id` return.

julianocosta89:
I don't think 2 files is too much of a complexity TBH.

PR #200, opened 2022-07-11
Add container_name s to compose file

mic-max:
## Changes

Mostly to cut down on the size of the name that appears in the logs.

Before
image

After
image

One thing to note is that if you run more than 1 of an image at the [same time this name will be ignored](https://docs.docker.com/compose/compose-file/compose-file-v3/#container_name)


cartersocha:
nice qol change. I like it

cartersocha (approved):

PR #202, opened 2022-07-11
Updating current architecture

cartersocha:
Update architecture mermaid graph

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,6 @@ your backend as well.
 languages that talk to each other over gRPC. Plus one Load Generator which uses
 Locust to fake user traffic.
 
-
 ```mermaid
 graph TD
 
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ your backend as well.
 languages that talk to each other over gRPC. Plus one Load Generator which uses
 Locust to fake user traffic.
 
+
 ```mermaid
 graph TD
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -146,6 +146,7 @@ checkoutservice --> currencyservice
 checkoutservice --> emailservice
 checkoutservice --> paymentservice
 checkoutservice --> shippingservice
+checkoutservice --> |evalFlag| featureflagfeservice
```
mic-max, line 149: In regard to naming the arrows: do we think we'll in the future call out all the request types to and from services?

README.md
```diff
@@ -154,7 +155,15 @@ frontend --> checkoutservice
 frontend --> currencyservice
 frontend --> recommendationservice --> productcatalogservice
 frontend --> shippingservice
+frontend --> |evalFlag| featureflagfeservice
 
+productcatalogservice --> |evalFlag| featureflagfeservice
+
+featureflagbeservice(Flag Server):::erlang
+featureflagfeservice(Flag UI/API):::erlang
+featureflagstore[(Flag Store<br/>&#40Blob/DB&#41)]
```
mic-max, line 164: Do we want to callout the exact technology (Postgres) since we do with Redis just to stay consistent?

mic-max (approved):

README.md
cartersocha: Yeah I think so. Maybe not necessarily in place. We can have multiple versions with various levels of detail. 

I'd like to add in OTel components too

README.md
cartersocha: Yep good call out

Issue #153, opened 2022-06-17
deployment demo webstore to kubernetes


wph95:
## helm chart
- [ ] demo-webstore
- [ ] otelcol
- [ ] jaeger
- [ ] Prometheus
- [ ] document
- [ ] Multi-node (load balancing, replica etc.)
- [ ] documents
- [ ] test

## Kubernetes quick setup (local setup)
- [ ] Kubernetes YAML file
- [ ] documents

---

The demo store currently only supports docker-compose deployments.
We should support the deployment of the demo store to Kubernetes.
There are two ways to deploy to Kubernetes.

1. Kubernetes YAML
  Excellent for beginners, with a quick start. simply run a command line

  ```
  kubectl apply -f https://raw.githubusercontent.com/....../demo-store.yaml
  ```

2. helm chart
   You can flexibly choose the configuration of the demo store you want to deploy.
   For example, helm/value.yaml

``` 
feature:
    log_collect: true
    epbf: true
    database: false
loadgenerator:
    speed: 10
```

This issue is part of #41

fatsheep9146:
I agree with this issue. With kubernetes, we can also involve opentelemetry operator into this demo which is more close to the way of using otel-collector for users who are also using kubernetes.

dmitryax:
The helm chart should be probably added to https://github.com/open-telemetry/opentelemetry-helm-charts

Frapschen:
I would like to help the helm chart :)

wph95:
We need to have a discussion/vote on whether our helm chart should be called `opentelemetry-demo` or` opentelemetry-demo-webstore` (to be consistent with the repo name)
@austinlparker @cartersocha @open-telemetry/demo-webstore-approvers  @open-telemetry/demo-webstore-maintainers

wph95:
helm chart: opentelemetry-demo

JaredTan95:
> helm chart: opentelemetry-demo-webstore

Vote: I'd like the chart called the same name as the repo

dmitryax:
I'd say we should keep it consistent with other charts which are called the same as repos. What's the benefit of dropping `-webstore`?

puckpuck:
We are investigating changing the repo name, and dropping the `-webstore` suffix.  See #222

cartersocha:
Yall should be unblocked now. Please proceed

cartersocha:
There’s a helm chart you can start with already available @xoscar. Check the main readme

[...]

PR #210, opened 2022-07-15
Add Simple kubernetes setup to test 

dineshg13:
Fixes a part of #153  
Adds kubernetes files to test webstore in a `k8s` cluster. 
## Changes

1. Adds a new directory `k8s_local` with the setup necessary to run the demo webstore on a k8s cluster. 
2. Adds changes to `Makefile` to make easy to setup the services in `k8s`

Cons of the approach:
1. We now a different config files for running services via `docker-compose` and `k8s` but that is inevitable as some of config are different are different in two environments.

### DIFF
--- a/k8s_local/adservice.yaml
+++ b/k8s_local/adservice.yaml
@@ -1,0 +1,78 @@
+
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: adservice
+  labels:
+    app: adservice
+    component: otel-demo-adservice
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: adservice
+      component: otel-demo-adservice
+  template:
+    metadata:
+      labels:
+        team: team-opentelemetry
+        service: adservice
+        app: adservice
+        component: otel-demo-adservice
+    spec:
+      containers:
+      - name: adservice
+        image:  otel/demo:v0.2.0-alpha-adservice
+        imagePullPolicy: Always
+        resources:
+          limits:
+            memory: 200Mi
+            cpu: 200m
+          requests:
+            memory: 200Mi
+            cpu: 200m
+        env:
+        - name: MY_POD_IP
+          valueFrom:
+            fieldRef:
+               apiVersion: v1
+               fieldPath: status.podIP
+        - name: OTEL_RESOURCE
+          value: k8s.pod.ip=$(MY_POD_IP)
+        - name: HOST_IP
+          valueFrom:
+            fieldRef:
+              fieldPath: status.hostIP
+        - name: OTEL_AGENT_ENDPOINT
+          value: "$(HOST_IP)"
+        - name: OTEL_EXPORTER_OTLP_ENDPOINT
+          value: "http://$(OTEL_AGENT_ENDPOINT):4318"
+        - name: OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
+          value: "http://$(OTEL_AGENT_ENDPOINT):4317"
+        - name: OTEL_RESOURCE_ATTRIBUTES
+          value: "service.name=adservice,deployment.environment=staging"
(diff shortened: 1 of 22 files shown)


k8s_local/adservice.yaml
```diff
@@ -0,0 +1,78 @@
...
+    metadata:
+      labels:
+        team: team-opentelemetry
+        service: adservice
+        app: adservice
+        component: otel-demo-adservice
+    spec:
+      containers:
+      - name: adservice
+        image:  otel/demo:v0.2.0-alpha-adservice
```
fatsheep9146, line 26: I think the image tag version should be a variable, and keep up to date.

dineshg13: Made this an env var that we set in `Makefile`. Let me know if there is a better way to achieve same.

wph95 (changes requested):
Communicated with the author via slack. It was agreed that the pr will be continued after the otel demo helm chart is completed.

saurabhdes:
I tested this on my Macbook Pro after creating a local cluster via `kind`. I see deployments getting created (e.g. for PaymentService, frontend etc.  but no pods. I ran the command `make k8s_local_apply` . Not sure if i am missing something. When i do a describe on deployment e.g. `kubectl describe deployment paymentservice`, i noticed the image name seems incorrect `otel/demo:-paymentservice` and export DOCKER_IMAGE_VERSION=v0.2.0-alpha doesn't seem to be applied to the yaml file.

dineshg13:
@saurabhdes thanks for testing. I made a fix to makes sure env var is set properly in `Makefile` . Could you retest this ?

saurabhdes:
thanks @dineshg13 for the fix. I am still having same issues with my kind cluster (could be my machine issue). But i was able to deploy successfully on AKS cluster. 
Screen Shot 2022-07-17 at 6 44 43 PM
20/179434584-b6beac85-acc9-419b-8fac-5c792cdeb445.png">
Not all services were there as i am running into a separate memory issue for the otel daemon set to run. I will try increasing memory size of AKS nodes.

fatsheep9146:
> Not all services were there as i am running into a separate memory issue for the otel daemon set to run. I will try increasing memory size of AKS nodes.

Maybe consider only using limit for resources not request to avoid this kind of problem? @dineshg13

saurabhdes:
I removed the "resources" section from k8s_local/otel-agent-daemonset.yaml and after that the all the otel collector daemonset started. Then i saw more accurate flowmap in Jaeger.

image

fatsheep9146:
> I removed this
> 
> ```
> resources:
>             limits:
>               cpu: 1
>               memory: 2Gi
>             requests:
>               cpu: 100m
>               memory: 400Mi
> ```
> 
> "resources" section from k8s_local/otel-agent-daemonset.yaml and after that the all the otel collector daemonset started. Then i saw more accurate flowmap in Jaeger.
> 
> image

Although this works, but I think it's the yaml with resources section is still important, since the resources spec is also important observation information. I think maybe a flag to switch between with-resource mode and without-resource mode is better. @dineshg13 @wph95

k8s_local/adservice.yaml
```diff
@@ -0,0 +1,75 @@
...
+  template:
+    metadata:
+      labels:
+        service: adservice
+        app: adservice
+        component: otel-demo-adservice
+    spec:
+      containers:
+        - name: adservice
+          image: otel/demo:${DOCKER_IMAGE_VERSION}-adservice
```
JaredTan95, line 24: following https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/156, the package is now public, we could use `ghcr.io/open-telemetry/demo` now.

dineshg13:
Closing this in favor of https://github.com/open-telemetry/opentelemetry-helm-charts/pull/280

PR #215, opened 2022-07-17
Bug fix for sending metrics from frontend

dineshg13:
Bug fix for frontend , to make sure it sends metrics. Before this change , we were initialing `meter` and `httpRequestCounter` before we set `global`  meterProvider. 

## Changes
After this change, when we run `docker compose up` , we should be able to see metrics in `prometheus` .


src/frontend/main.go
```diff
@@ -114,7 +114,7 @@ func InitMeterProvider() (*controller.Controller, error) {
 		return nil, err
 	}
 
-	controller := controller.New(
+	pusher := controller.New(
```
dineshg13, line 117: This change is to avoid overriding controller which package that is being imported.

cartersocha:
@fatsheep9146, can you take a look at this please?

fatsheep9146:
> @fatsheep9146, can you take a look at this please?

Absolutely, I will look at this pr

fatsheep9146:
This pr is lgtm. @cartersocha 
But there exists conflicts to be resolved @dineshg13

cartersocha (approved):

PR #218, opened 2022-07-20
centralize a tracing doc

cartersocha:
updated per previous pr feedback

### DIFF
--- a/docs/trace_service_features.md
+++ b/docs/trace_service_features.md
@@ -45,7 +45,7 @@ tba
 - :construction: [Trace
   Links](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#links-between-spans)
 - :construction: [Baggage](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md#overview)
-  
+
 ### [Currency Service](../src/currencyservice/README.md)
 
 tba
--- a/docs/trace_service_features.md
+++ b/docs/trace_service_features.md
@@ -44,8 +44,7 @@ tba
   Propagation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/README.md#context)
 - :construction: [Trace
   Links](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#links-between-spans)
-- :construction:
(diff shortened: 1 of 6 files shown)


cartersocha:
new links @cijothomas ! good suggestion

docs/trace_service_features.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Traces
+
+- :100: [Instrumentation
+  Libraries](https://opentelemetry.io/docs/concepts/instrumenting-library/)
+- :construction: [Manual Span
+  Creation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md#manual-instrumentation)
+- :construction: [Span Data
+  Enrichment](https://opentelemetry.io/docs/concepts/glossary/#metadata)
+- :construction: Interprocess Context Propagation
```
reyang, line 42: I probably missed something here, what does "Interprocess Context Propagation" mean, and what's the difference compared to "Intra-service Context Propagation"?

Maybe the intention is:
* Context Propagation within a process
* Context Propagation across RPC boundaries

cartersocha: Yes

reyang (approved):

docs/trace_service_features.md
cartersocha: To double check given the approval is the preference to update to your language? I like the service framing but I could see the appeal of the more precise ipc / rpc etc.

reyang: Here is my suggestion:
1. Given this PR is focusing on reorganizing the docs, we don't want to block the PR.
2. These terms (e.g. "Interprocess Context Propagation", "Intra-service Context Propagation" and "Trace Links") are not defined AND counter-intuitive (e.g. in computer world "Inter-process" is frequently used instead of "Interprocess", and is normally used under the context of IPC), we should revisit them in a separate PR. Ideally we should use defined terms and link to the definition (e.g. instead of using "Trace Links", I think we should use "Span Links" and add a link to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#specifying-links). As a summary:
  2.1. We should always use defined terms whenever possible, and we should link to the definition.
  2.2. If 2.1 doesn't work, we may introduce/define our own terms, and the terms need to be well explained.

PR #232, opened 2022-07-26
add metrics to satisfy http server semantic conventions

fatsheep9146:

related #70 

## Changes

Add more metrics to satisfy [http server metrics semantic coventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/http-metrics.md#http-server) 

- http.server.request.size
- http.server.response.size
- http.server.active_requests

http.server.response.size dashboard demo

image

http.server.active_requests dashboard demo

image


wph95 (approved):

cijothomas:
Qn - how does the Go front end export this? Is it done via OTLP exporter? Or via Prometheus directly?

cijothomas:
> Qn - how does the Go front end export this? Is it done via OTLP exporter? Or via Prometheus directly?

never mind, I think i got the answer from https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/frontend/main.go#L108

fatsheep9146:
> > Qn - how does the Go front end export this? Is it done via OTLP exporter? Or via Prometheus directly?
> 
> never mind, I think i got the answer from https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/frontend/main.go#L108

Yes, via OTLP exporter

cartersocha:
@fatsheep9146, could you update the changelog and then I'll merge

cartersocha:
Trying to get better about that upfront 😄. Do any docs need to be updated? I think we have a centralized metric doc but it's a bit immature

puckpuck:
Would this PR be in conflict with #236 which is replacing frontend with a Next.js based service?

fatsheep9146:
> Would this PR be in conflict with #236 which is replacing frontend with a Next.js based service?

Yes, I think @puckpuck is right, since the whole frontend service is replaces with Next.js implementation. Should I close this pr ? @cartersocha

cartersocha:
Yes let’s wait for the new front end

fatsheep9146:
@cartersocha I think #70 , #49 should also be modified or close to avoid confusion

cartersocha:
Roger that. Thanks for calling them out. Closed for now

Issue #73, opened 2022-05-30
Review and enhance metric support for ad service service (Java)


cartersocha:
The following are requirements for metrics in the project. Not all services will meet all requirements.

- [ ] Service metrics should not be duplicative of trace-derived metrics.
- [ ] Services should emit multiple business metrics where appropriate.
- [ ] Services should emit trace exemplars for their metrics.
- [ ] Services should use push metrics where possible.
- [ ] Services should explicitly read baggage where needed.
- [ ] Collectors should perform metric transforms to normalize resource and service metrics.

Referencing issue: https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/43

jlawrienyt:
Proposed Solution:

The Ad Service already has some business logic around whether to [deliver targeted ads or random ads, based on the context and ad availability](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/adservice/src/main/java/hipstershop/AdService.java#L116).  We should capture this decision as a single metric named `app.ads.ad_requests` with two dimensions: `ad_request_type` with possible values `TARGETED` and `NOT_TARGETED`; and `ad_response_type` with possible values `TARGETED` and `RANDOM`.

Currently, the code path that would serve a random ad is not being exercised by the load generator, nor is it possible to exercise it by browsing the demo site, because [all products](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/productcatalogservice/products.json) have categories with [associated ads](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/adservice/src/main/java/hipstershop/AdService.java#L188).  I propose we add a new product in a new category that does have an associated ad in order to demonstrate this behavior.  We should probably do this regardless of the metrics change as the [`getRandomAds` span](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/adservice/src/main/java/hipstershop/AdService.java#L159) is not currently getting produced.

cc @puckpuck

austinlparker:
Closing as basic metrics have been added

PR #237, opened 2022-07-26
Set OTEL_EXPORTER_OTLP_METRICS_ENDPOINT needed for Java agent

dineshg13:
Fixes a part of #73 

## Changes

1. Update javaagent jar to the latest version
2. Set OTEL_EXPORTER_OTLP_METRICS_ENDPOINT as required by java agent . The variables are listed here https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md#otlp-exporter-both-span-and-metric-exporters

Post this change , when we do `docker compose up` , we should see jvm metrics in prometheus

Screen Shot 2022-07-26 at 6 39 06 PM


puckpuck:
Should we be using the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable only? We only have a single collector endpoint for all OTLP data, and shouldn't need to use the TRACES and METRICS specific flavors.

julianocosta89:
@puckpuck, I'd argue that if we keep `_TRACES` and `_METRICS` we are "showcasing" how to use env vars to configure those 2 signals, even though they are pointing to the same collector.

Personally I'd accept this PR.

austinlparker:
I suppose explicit is better than implicit, although I wonder if there are other implications of using the separate env vars for configuration?

dineshg13:
i agree with @julianocosta89  & @austinlparker  . i like variables being explicit in demo application.

austinlparker (approved):
This was discussed at the SIG meeting and we'd like to use explicit variables for each signal.

.env
```diff
@@ -9,6 +9,8 @@ COLLECTOR_ZIPKIN_HOST_PORT=:9411
 OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://otelcol:4317
 OTEL_EXPORTER_OTLP_ENDPOINT=${OTEL_EXPORTER_OTLP_TRACES_ENDPOINT}
```
reyang, line 11: Curious, any reason to leave a blank here? Doesn't seem to be consistent with Ln9-10.

reyang (approved):

cartersocha:
@dineshg13 could you please update the changelog too

.env
dineshg13: no reason. Removing it

PR #238, opened 2022-07-27
Limit size of accumulating logs sensibly

cedricziel:
This change will limit the amount of logs accrueing in docker
to 2 segments à 5M and prevents the demo from flooding the disk
which is useful in long-running demos.


julianocosta89 (approved):
Thanks for spotting that!
LGTM!

puckpuck (approved):
thanks for this

PR #239, opened 2022-07-27
Introduce the LGTM stack as alternative OpenSource analysis tooling

cedricziel:
## Changes

In addition to the default stores for tracing and metrics, this introduces an alternate
scenario with:

* Grafana Loki as a store for OTel logs
* Grafana Tempo as a datastore for traces
* Grafana Mimir as a datastore for Metrics
* OpenTelemetry Collector Contrib preconfigured for the LGTM stack

This alternate scenario is activated when including an additional compose file.


julianocosta89:
I personally wouldn't add extra unnecessary tools to the sample.
We are already using Grafana, Jaeger and Prometheus to visualize Traces and Metrics.
I would rather encourage @cedricziel to create a fork showcasing LGTM and we can add it to the `Demos featuring Online Boutique` in the root the README.

cedricziel:
That's sensible :)

PR #240, opened 2022-07-28
New badges

cartersocha:
Add a couple more badges


reyang (approved):

reyang:
There is a standard place for license, maybe duplicated?

image

puckpuck (approved):

cartersocha:
> There is a standard place for license, maybe duplicated?
> 
> image

Yeah I know but I saw another OTel repo use the badge & I like how in your face it is. With something like the demo I feel like licensing is key so important for it to be bit glaring. But happy to just defer to the single reference too

reyang:
> Yeah I know but I saw another OTel repo use the badge & I like how in your face it is. With something like the demo I feel like licensing is key so important for it to be bit glaring. But happy to just defer to the single reference too

I don't have strong opinion, pick your favorite one 😃

PR #241, opened 2022-07-28
going language first for docs

cartersocha:
* Reducing readme size by eliminating service table and adding new file
* Updating to language first for trace docs


README.md
```diff
@@ -122,7 +122,7 @@ your backend as well.
 
 ## Architecture
 
-**Online Boutique** is composed of 10 microservices written in different
+**Online Boutique** is composed of 11 microservices written in different
 languages that talk to each other over gRPC. Plus one Load Generator which uses
 Locust to fake user traffic.
```
reyang, line 127:
```suggestion
[Locust](https://locust.io/) to fake user traffic.
```

docs/service_table.md
```diff
@@ -0,0 +1,16 @@
...
+| [emailservice](../src/emailservice/README.md)                   | Ruby        | Sends users an order confirmation email (mock).                                                                                   |
+| [checkoutservice](../src/checkoutservice/README.md)             | Go            | Retrieves user cart, prepares order and orchestrates the payment, shipping and the email notification.                            |
+| [recommendationservice](../src/recommendationservice/README.md) | Python        | Recommends other products based on what's given in the cart.                                                                      |
+| [adservice](../src/adservice/README.md)                         | Java          | Provides text ads based on given context words.                                                                                   |
+| [featureflagservice](../src/featureflagservice/README.md)         | Erlang/Elixir | CRUD feature flag service to demonstrate various scenarios like fault injection & how to emit telemetry from a feature flag reliant service.                                             |
```
reyang, line 15: Consider aligning the columns.

docs/trace_service_features.md
```diff
@@ -2,34 +2,38 @@
...
-- [Product Catalog Service](#product-catalog-service)
-- [Recommendation Service](#recommendation-service)
-- [Shipping Service](#shipping-service)
-
-## OpenTelemetry Feature Set by Services
+- [C++](#c)
+- [Erlang/Elxir](#erlangelixir)
+- [Go](#go)
+- [Java](#java)
+- [.NET](#net)
```
reyang, line 9: Perhaps put this on top (`.` < `A`), or consider using `C#`

reyang (approved):

README.md
```diff
@@ -122,7 +122,7 @@ your backend as well.
 
 ## Architecture
 
-**Online Boutique** is composed of 10 microservices written in different
+**Online Boutique** is composed of 11 microservices written in different
```
reyang, line 125: Maybe just use "several"

docs/trace_service_features.md
```diff
@@ -2,34 +2,38 @@
...
-- [Email Service](#email-service)
-- [Feature Flag Service](#feature-flag-service)
-- [Frontend Service](#frontend-service)
-- [Payment Service](#payment-service)
-- [Product Catalog Service](#product-catalog-service)
-- [Recommendation Service](#recommendation-service)
-- [Shipping Service](#shipping-service)
-
-## OpenTelemetry Feature Set by Services
+- [C++](#c)
```
fatsheep9146, line 5: c -> c++ ?

docs/trace_service_features.md
```diff
@@ -2,34 +2,38 @@
...
+- [Erlang/Elxir](#erlangelixir)
+- [Go](#go)
+- [Java](#java)
+- [.NET](#net)
+- [Node.js](#nodejs)
+- [Python](#python)
+- [Ruby](#ruby)
+- [Rust](#rust)
```
fatsheep9146, line 13: c# is missed

docs/trace_service_features.md
cartersocha: C++ is a broken link. It needs to be c for it work

docs/trace_service_features.md
cartersocha: .net should be above node

docs/trace_service_features.md
cartersocha: If we use c# I think it’ll get confused with c++ since linking ignores special characters

reyang: Maybe `#c-1`? (it's hard to maintain though)

docs/trace_service_features.md
fatsheep9146: got it

docs/service_table.md
cartersocha: Just updated it a bit using a vs code extension. Better but not 100% perfect

docs/trace_service_features.md
cartersocha: Could you commit an example of what you mean on the "." or point to an example

cartersocha: Or we could use the trusty dotnet & cpp naming

reyang: > Could you commit an example of what you mean on the "." or point to an example

I guess most developers would treat alphabetical ordering as "ordering based on the ASCII code or UNICODE ordinal".
So the order should be:
* .NET
* Algol
* Basic
* C++
* D
* Erlang
* Fortran
* Go

reyang: In addition (I might be pedantic), I found it confusing when people mix up programming language and runtime. For example, I really think ECMAScript https://en.wikipedia.org/wiki/ECMAScript is the programming language (and the proverb is "JavaScript", which can be used to refer to ECMAScript or the very original language developed by Netscape, JScript developed by Microsoft or many other things).

I guess it's probably fine to use JavaScript as the name of the programming language, most developers would not pick whether it's ECMAScript or TypeScript.

I do think Node.js is NOT a programming language, it's a runtime.

cartersocha: updated to dotnet and javascript

docs/trace_service_features.md
mic-max: or `cpp`

PR #242, opened 2022-07-28
Add e2e tests to remaining services

### LINKED ISSUE #195: Add E2E service tests

mic-max:
Fixes #195

## Changes

- Add CartService e2e Tests
- Product Service e2e Tests
- Shipping Service e2e Tests
- Currency Service e2e Tests
- Add e2e tests to Checkout Service + boilerplate for email service
- Add e2e tests to Ad Service

image


julianocosta89:
hey @mic-max do  we really need to expose the localhost ports?
I've done the same question to @xoscar here: https://github.com/open-telemetry/opentelemetry-demo/pull/236#discussion_r931516594

For testing instead of opening the localhost, we could spin up a container within the same docker network and execute the test.
The test container would be able to call the other containers by `name:port`, and we wouldn't need to worry about modifying all the ports.

Another solution would be to keep the internal ports the apps are already using, and expose a different one within the localhost.
> -p 8080:80 | Map TCP port 80 in the container to port 8080 on the Docker host.

In that case we could have multiple container running on port 5050 (for instance), and just change the localhost exposed port.

I'd rather go for the 1st solution.

mic-max:
> be able to call the other containers

Is there any downside to opening to ports to the localhost? I think it is the simplest way to do it.
The test container option sounds doable, does anyone know if that's a common architecture for testing docker containers?

julianocosta89:
> @julianocosta89 Is there any downside to opening to ports to the localhost? I think it is the simplest way to do it.

No actual downside in a demo perspective, just in a real life app.
The main issue now is that we have 2 PRs changing port numbers.
Someone will have to solve some merging conflicts 😅 .

Another point is that we need to ensure all services all communicating with each other with the new ports. But that is fine, considering that it will be done once.

test/test.js
```diff
@@ -6,75 +6,238 @@ const test = require('ava')
...
+
+  t.truthy(res.order.orderId)
+  t.truthy(res.order.shippingTrackingId)
+  t.truthy(res.order.shippingAddress)
+  t.is(res.order.shippingCost.currencyCode, 'USD')
+})
+
+// --------------- Email Service ---------------
+
+// TODO
```
fatsheep9146, line 137: This line maybe redundant?

src/shippingservice/Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN GRPC_HEALTH_PROBE_VERSION=v0.4.7 && \
     chmod +x /bin/grpc_health_probe
 WORKDIR /app
 COPY --from=builder /app/target/release/shippingservice /shippingservice
-ENV PORT=50051
+ENV PORT=50052
 
-EXPOSE 50051
+EXPOSE 50052
```
fatsheep9146, line 21: Should this line use ${SHIPPING_SERVICE_PORT} instead of 50052?

mic-max: Ya, I will do a followup PR that makes sure we do this for all services. It will also include only use the actual service port env var for example SHIPPING_SERVICE_PORT in this case rather than assigning a new env var named PORT which seems like a pointless mapping IMO.

test/test.js
mic-max: this `TODO` is in reference to the email service below. It is a bit different than all the others since it works with JSON POST requests instead of using gRPC service loaded from the protobuf.

src/shippingservice/Dockerfile
julianocosta89: @mic-max I'm actually not able to test this PR due this.
The port 50052 is already allocated by FeatureFlagService and ShippingService is not able to start.

mic-max: Sorry, I fixed it now

cartersocha:
@fatsheep9146, @julianocosta89 any further comments / suggestions ?

fatsheep9146 (approved):

fatsheep9146:
> @fatsheep9146, @julianocosta89 any further comments / suggestions ?

LGTM

cartersocha:
@mic-max please resolve the conflicts then we'll merge

mic-max:
> @mic-max please resolve the conflicts then we'll merge

Resolved 😊

PR #247, opened 2022-08-02
featureflagservice: add src dir to docker image when building

tsloughter:
Fixes the issue where featureflagservice only had the HTTP interface start successfully on boot in docker compose.


mic-max (approved):

cartersocha:
@puckpuck fyi

puckpuck (approved):

PR #248, opened 2022-08-02
Use service environment vars, not setting new vars

mic-max:
Fixes https://github.com/open-telemetry/opentelemetry-demo/pull/242#discussion_r933603430

## Changes

Services will now read their port environment variables directly rather than having any intermediary $PORT variable for example.
Also, getting any environment variables should be done at startup of the service and if it is not present it can crash the program with a relevant error that would be more helpful than starting the service using a default port number that might not be actually communicating with other services which would be harder to detect problems.


puckpuck (approved):

docker-compose.yml
```diff
@@ -144,7 +144,7 @@ services:
       - "${EMAIL_SERVICE_PORT}"
     environment:
       - APP_ENV=production
-      - PORT=${EMAIL_SERVICE_PORT}
+      - EMAIL_SERVICE_PORT
```
ahayworth, line 147: How was this tested? The `emailservice` relies on the `PORT` variable by convention to configure itself. I am unable to verify this at the moment myself but I am virtually certain that this is now broken.

ahayworth (changes requested):

docker-compose.yml
ahayworth: Codespaces to the rescue! With your changes:

```
Attaching to email-service
email-service            | == Sinatra (v2.2.0) has taken the stage on 4567 for production with backup from Puma
email-service            | I, [2022-08-02T20:13:18.324598 #1]  INFO -- : Instrumentation: OpenTelemetry::Instrumentation::Sinatra was successfully installed with the following options {}
email-service            | Puma starting in single mode...
email-service            | * Puma version: 5.6.4 (ruby 3.1.2-p20) ("Birdie's Version")
email-service            | *  Min threads: 0
email-service            | *  Max threads: 5
email-service            | *  Environment: production
email-service            | *          PID: 1
email-service            | * Listening on http://0.0.0.0:4567
email-service            | Use Ctrl-C to stop
```

Notably that is _not_ the value of `EMAIL_SERVICE_PORT`.

Without your changes:

```
email-service            | Puma starting in single mode...
email-service            | * Puma version: 5.6.4 (ruby 3.1.2-p20) ("Birdie's Version")
email-service            | *  Min threads: 0
email-service            | *  Max threads: 5
email-service            | *  Environment: production
email-service            | *          PID: 1
email-service            | * Listening on http://0.0.0.0:8080
email-service            | Use Ctrl-C to stop
```

mic-max: Yes you're right I knew that `emailservice` worked in a special way but I forgot to fix it. I will undo that change.
I guess it can be changed like [this](https://stackoverflow.com/a/44735521) but I'll forgo that change for now as it looks more complicated than simply doing this env var map.

cartersocha:
@mic-max 🦭

mic-max:
@cartersocha 🦧

docker-compose.yml
ahayworth: @mic-max ruby _is_ special, yes! I think we can just do something simple in the Dockerfile and be okay with that - I'll add a PR that does it quickly. ❤️

PR #259, opened 2022-08-04
[currencyservice] - use environment variables to configure tracer

puckpuck:
## Changes

The currencyservice was using command line arguments to set the OTLP endpoint and resource attributes. This was done by parsing the command line options, and rebuilding the equivalent objects in code when initializing the tracer. The command line options were set using `OTEL_` environment variables, which may have led to some confusion on how things were being initialized.

This PR undoes much of that logic and lets the OpenTelemetry C++ SDK configure the OTLP endpoint and resource attributes using the `OTEL_` environment variables.


lalitb (approved):
LGTM. Thanks for the change

cartersocha (approved):

PR #265, opened 2022-08-05
[currency] - add span attributes

puckpuck:
## Changes

Adds span attributes to currency conversion request


cijothomas (approved):

cartersocha:
Should we update the trace table too or was that already done in a previous pr?

cartersocha:
> Should we update the trace table too or was that already done in a previous pr? 

Looks like you did it already nvm

puckpuck:
> Should we update the trace table too or was that already done in a previous pr?

That is already part of previous efforts since we combine adding span attributes and events into the same task and this service already had span events. Perhaps we should consider breaking those 2 items out separately in the feature docs?

cartersocha (approved):

PR #245, opened 2022-08-01
[productcatalog] - generate error on feature flag

puckpuck:
## Changes

Generates an error on a specific product when a feature flag is enabled.
related issue: #162 

Currently, this PR will not work since we have an issue that needs resolving with the feature flag service. Once that is resolved, we can finalize this PR.

For significant contributions please make sure you have completed the following items:

_I will update the changelog after we resolve feature flag service communication issues_
* [x] Appropriate `CHANGELOG.md` updated for non-trivial changes

PR #271, opened 2022-08-09
[featureflag] - use full demo protos

puckpuck:
Unblocks #245 

## Changes

Uses a full copy of the `demo.proto` to build corresponding gRPC modules. Previously we were using a small subset of `demo.proto` without the `package hipstershop` declaration. There are several new files in this PR because we are now building the Erlang behavior modules for all the services part of the protobuf definition. This can be addressed in another PR to clean up the docker build process for this service.

Does this PR require a changelog update?


tsloughter (approved):
@puckpuck thanks! This looks good. All the extra modules is why I was waiting to do this -- and the fact the `ffs_demo_pb` module is now huge. But looking at this PR it seems that new module is only 7k lines and the old 2k, so not bad and not an issue.

I would suggest using a symlink to link to the top level `demo.proto` if possible -- I think Windows supports those now? -- otherwise I'll focus on making a change to the grpc server generator to support reading files from parent directories.

puckpuck:
I created #279 to track updating Dockerfile for this service

PR #277, opened 2022-08-09
Docs - Table layout for manual span attributes

puckpuck:
## Changes

Uses a table layout to list all span attributes by service
Adds a type and description column for each span attribute

Type is restricted to `number` and `string` only.


cartersocha (approved):

fatsheep9146 (approved):

mviitane (approved):

docs/manual_span_attributes.md
```diff
@@ -4,84 +4,108 @@ This document contains the list of manual Span Attributes used throughout the de
...
 ## ProductCatalogService
 
-* `app.product.id`
-* `app.product.name`
-* `app.products.count`
+| Name                 | Type   | Description                 |
+|----------------------|--------|-----------------------------|
+| `app.product.id`     | string | Product Id                  |
+| `app.product.name`   | string | Product name                |
+| `app.products.count` | number | Number of products returned |
```
reyang, line 95: It might be confusing/misleader what "returned" means here. I guess the intention is "returned by the API"? (or is it "items/products returned by the customer as part of the refunding"?)

docs/manual_span_attributes.md
```diff
@@ -4,84 +4,108 @@ This document contains the list of manual Span Attributes used throughout the de
...
+|-----------|------|-------------|
+| None yet  |      |             |
 
 ## PaymentService
 
-* `app.payment.cost`
-* `app.payment.currency`
+| Name                   | Type   | Description        |
+|------------------------|--------|--------------------|
+| `app.payment.cost`     | number | Total payment cost |
```
reyang, line 86: I'm not expert in payment systems, I wonder if this is normally called "amount (transaction amount)" or "cost"?

reyang (approved):
I've left couple non-blocking comments/suggestions, the direction looks good to me.

docs/manual_span_attributes.md
puckpuck: I agree this should be called `amount` and not `cost`. 

Next week I will propose changes to all attributes so we can remove unnecessary ones, normalize names, as well as make slight name adjustments where required. We can track that in with #280

PR #291, opened 2022-08-11
Add auto instrumentation to frontend service

cedricziel:
## Changes

This traces the nextJS backend part of the frontend service.

Since I didnt want to swap out large amounts of code in an unknown framework,
I just applied the auto-instrumentation from the operator by requiring and starting
a customized version of the SDK.

Customizations based on https://ross-hagan.com/blog/instrument-nextjs-opentelemetry

image

### DIFF
--- a/src/frontend/Dockerfile
+++ b/src/frontend/Dockerfile
@@ -16,6 +16,8 @@ COPY ./src/frontend .
 RUN npm run grpc:generate
 RUN npm run build
 
+FROM ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-nodejs:0.31.0 as instrumentation
+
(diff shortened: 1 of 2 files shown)


puckpuck:
I like where this is going, but using the operator's autoinstrumentation image to pull in the OpenTelemetry dependencies will limit the dev experience. Can we update the frontend's package.json file to pull in the dependencies within the dev environment instead?

xoscar:
Hey @cedricziel I also worked on some instrumentation for the front end (serverside/client side) you can find it here https://github.com/open-telemetry/opentelemetry-demo/pull/293

puckpuck:
I think we should close this and look to finish #293

cedricziel:
Sweet! Closing here..

PR #292, opened 2022-08-11
Fix processid(pid) for currencyservice

lalitb:
Fixes #.

## Changes

`currencyservice` is running under shell with pid>1 inside container. Which makes it difficult to send signals using `docker kill` command. Changed to run it as pid=1 so it can get catch the TERM/INT signal through `docker kill` command.


mviitane (approved):
Looks good. After the change, currency-service in running with pid 1.

reyang (approved):

Issue #284, opened 2022-08-10
[Frontend] Add Tests to Frontend


xoscar:
As a fast follow after the initial FE implementation, we want to ensure the basic flows for the webstore are covered by e2e tests, components, key logic should have unit tests and as a stretch goal we should have a plan to include trace validations. 

The implementation of these suites tests needs to match the proposal:
![Screen Shot 2022-08-10 at 10 39 25 a m](https://user-images.githubusercontent.com/11051031/183947044-2229f0be-d171-494e-acf0-ad0fcee493dd.png)

PR #298, opened 2022-08-12
🧪 [Frontend] 284 adding cypress e2e tests

xoscar:
Fixes https://github.com/open-telemetry/opentelemetry-demo/issues/284.

## Changes

- Adds E2E integration test for the frontend using Cypress.
- Fixes the backend services integration tests by dockerizing them.

https://www.loom.com/share/e6e2cbba1ca84bfb8670805b53090d77

Running e2e tests:
`docker compose run frontendtests`

Running backend integration tests:
`docker compose run integrationTests`


docker-compose.yml
```diff
@@ -183,6 +183,49 @@ services:
       - shippingservice
     logging: *logging
 
+  # Frotend Tests
+  frontendtests:
```
xoscar, line 187: Adding images for both front end cypress tests and integration tests

src/frontend/components/Ad/Ad.tsx
```diff
@@ -7,7 +7,7 @@ const Ad = () => {
   } = useAd();
 
   return (
-    <S.Ad>
+    <S.Ad data-cy="ad">
```
xoscar, line 10: Adding cypress attributes

src/frontend/cypress.config.ts
```diff
@@ -0,0 +1,23 @@
...
+const myEnv = dotEnv.config({
+  path: resolve(__dirname, '../../.env'),
+});
+dotenvExpand.expand(myEnv);
+
+const { FRONTEND_ADDR = '', NODE_ENV, FRONTEND_PORT = '8080' } = process.env;
+
+export default defineConfig({
+  env: {
+    baseUrl: NODE_ENV === 'production' ? FRONTEND_ADDR : `http://localhost:${FRONTEND_PORT}`,
```
xoscar, line 15: Validating the env and the port for local vs CLI execution

src/frontend/cypress/e2e/Checkout.cy.ts
```diff
@@ -0,0 +1,32 @@
+describe('Checkout Flow', () => {
+  beforeEach(() => {
+    cy.visit(Cypress.env('baseUrl'));
+  });
+
+  it('should create an order with two items', () => {
```
xoscar, line 6: Some e2e tests

docker-compose.yml
```diff
@@ -183,6 +183,50 @@ services:
...
+    environment:
+      - AD_SERVICE_ADDR
+      - CART_SERVICE_ADDR
+      - CHECKOUT_SERVICE_ADDR
+      - CURRENCY_SERVICE_ADDR
+      - PAYMENT_SERVICE_ADDR
+      - PRODUCT_CATALOG_SERVICE_ADDR
+      - RECOMMENDATION_SERVICE_ADDR
+      - SHIPPING_SERVICE_ADDR
+      - EMAIL_SERVICE_ADDR
```
mic-max, line 218: nit: email in alphabetical order, same with below `depends_on`

src/frontend/cypress.config.ts
```diff
@@ -0,0 +1,23 @@
+import { defineConfig } from 'cypress';
+import dotEnv from 'dotenv';
+import dotenvExpand from 'dotenv-expand';
+import { resolve } from 'path';
+
+const myEnv = dotEnv.config({
+  path: resolve(__dirname, '../../.env'),
+});
+dotenvExpand.expand(myEnv);
```
mic-max, line 9: Oh, without this a value like `FRONTEND_ADDR=0.0.0.0:${FRONTEND_PORT}` in `.env` will remain as the literal string and not become `0.0.0.0:8080`? I didn't know this

src/frontend/cypress.config.ts
mic-max: I think instead of leaving these types of comments on the PR putting them in the source would be more valuable and they would exist in scenarios other than a code review :)

src/frontend/cypress/e2e/Checkout.cy.ts
```diff
@@ -0,0 +1,32 @@
...
+    cy.get('[data-cy="product-card"]').first().click();
```
mic-max, line 7: nit: could make a function like this to simplify the commands for reading
```javascript
const cyget(name) => cy.get(`[data-cy="${name}"]`)
```

mic-max: maybe also we can define all the `data-cy` values added to the frontend code in a file and pass Enum to this `cyget` method or similar so we can avoid potential typos. ex: 

```javascript
// before
cy.get('[data-cy="cart-item-count"]')

// after
import { CART_ITEM_COUNT } from './CypressFields'
cyget(CART_ITEM_COUNT)
```

just some ideas :)

src/frontend/cypress/support/commands.ts
```diff
@@ -0,0 +1,39 @@
+/// <reference types="cypress" />
```
mic-max, line 1: not needed?

src/frontend/cypress/support/e2e.ts
```diff
@@ -0,0 +1,20 @@
+// ***********************************************************
```
mic-max, line 1: also not needed?

test/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+FROM node:16-alpine
+
+USER node
+WORKDIR /app
+ENV NODE_ENV production
+
+COPY --chown=node:node --from=build /app/node_modules/ ./node_modules/
+COPY ./test ./
+COPY ./pb/demo.proto ./
```
mic-max, line 33:
```suggestion
```

test/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM node:16-alpine AS build
+
+WORKDIR /app
+
+COPY ./test/package*.json ./
+
+RUN apk add --no-cache python3 make g++ && npm ci
```
mic-max, line 21:
```suggestion
RUN apk add --no-cache python3 make g++ && npm ci --omit=dev
```
Is the `apk add --no-cache python3 make g++` needed? if so for what?

test/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+USER node
+WORKDIR /app
+ENV NODE_ENV production
+
+COPY --chown=node:node --from=build /app/node_modules/ ./node_modules/
+COPY ./test ./
+COPY ./pb/demo.proto ./
+
+ENTRYPOINT [ "sh", "-c", "npm test" ]
```
mic-max, line 35:
```suggestion
ENTRYPOINT [ "npm", "test" ]
```
does this not work? what does `sh -c` do?

test/test.js
```diff
@@ -1,243 +1,306 @@
...
 
-  const res = await currencyConvert(req)
-  t.is(res.currencyCode, "CAD")
-  t.is(res.units, 442)
-  t.is(res.nanos, 599380805)
-})
+  const res = await currencyConvert(req);
+  t.is(res.currencyCode, "CAD");
+  t.is(res.units, 442);
+  t.is(res.nanos, 599380804);
```
mic-max, line 179: this is interesting. this might be system dependent because I added this and it was passing...
This might need to be changed to allow some wiggle room on either side
```suggestion
  t.true(res.nanos >= 599380800);
  t.true(res.nanos < 599380809);
```
something like this for example..

test/test.js
```diff
@@ -1,243 +1,306 @@
 // Node
```
mic-max, line 1: Was the formatting of this file causing some linting errors? I'm fine with the change but ideally pure refactors should be a separate PR to easily review and merge quick so less people or possibly yourself maybe be affected by conflicts.

mic-max (changes requested):
LGTM other than some minor questions / feedback I left

docker-compose.yml
```diff
@@ -183,6 +183,50 @@ services:
       - shippingservice
     logging: *logging
 
+  # Frotend Tests
```
jorgeepc, line 186: typo `Frontend`

docker-compose.yml
```diff
@@ -183,6 +183,50 @@ services:
...
+      - tests
+    environment:
+      - CYPRESS_baseUrl=http://${FRONTEND_ADDR}
+      - NODE_ENV=production
+    working_dir: /cypress
+    volumes:
+      - ./src/frontend:/cypress
+
+  # Integration Tests
+  integrationTests:
```
jorgeepc, line 201: We should sync the case for `frontendtests` and `integrationTests`. Maybe `frontendTests`?

src/frontend/cypress.config.ts
xoscar: I tent do disagree, I feel that the code itself should be self-explanatory but for PRs it is good to present it as a story for the reviewer so even if they don't have enough context the basic flow of changes can be understood

src/frontend/cypress/e2e/Checkout.cy.ts
xoscar: Great idea!

test/Dockerfile
xoscar: I think we need to keep this as the tests need the proto file

test/test.js
xoscar: Make sense, I just ran the prettify command and it changed it 😬

xoscar:
Hey @mic-max thanks for the review! I just updated the PR, let me know what you think 😄

mic-max (approved):

cartersocha:
Lgtm. @xoscar could you add a change log entry & I'll merge

xoscar:
@cartersocha done 😄

PR #304, opened 2022-08-15
cache_to inline and cache_from

### LINKED ISSUE #83: Add cache_from command to docker compose
The local build time from docker compose is slow. To improve the build time we can use cache_from in the docker compose file. 

This item is dependent on: https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/78

GaryPWhite:
Fixes #83.

## Changes

Sets up [`cache_to` and `cache_from`](https://docs.docker.com/compose/compose-file/build/#cache_from) for docker-compose builds. Note that the [`cache_to` is set to `inline`](https://github.com/moby/buildkit#export-cache). My best estimation for how we push the cache with a given image up to ghcr is included in the link provided -- there's not much documentation on it from docker directly.

Also worth noting -- builds (once cached) are lightning fast!

* [x] Design discussion issue #83


cartersocha:
@austinlparker this might help with some of the build times

cartersocha:
Rebuilds for sure. 

I'm not an expert but I think it can help on initial builds too if we connect it to a cache source for images in gchr like Gary refers to. 

Ideally only the delta in the repo needs to be rebuilt

GaryPWhite:
I found through some tinkering that we should probably update the .env file when we do releases to ghcr -- otherwise it will not pick up the right version.

Otherwise we could put together a makefile that pulls the most recent version of image and then tries to build -- would automatically use the `cache_from` that way.

although then people would need `make` :)

austinlparker:
> although then people would need `make` :)

I mean... you kinda _do_ need `make` rn, even on Windows, if you want to get the toolchain. That said, would it break entirely if someone skipped make and just did docker-compose up?

mic-max (changes requested):
After merging this a release would be done so to upload the cached version of the images to the docker repository. I think just to be safe here we could try with a single service - I elect the particularly slow `shippingservice` (Rust) which will be easy to see the speedup and know the caching worked.

austinlparker:
@mic-max are there any changes on this that are pending? shows you still have stuff requested but I don't see it. would like to get this PR updated and merged in.

mic-max (approved):

cartersocha:
@GaryPWhite please update the branch so we can merge!

austinlparker (approved):

austinlparker:
anyone know why maintainers cant refresh this branch? weird. waiting on OP to refresh then g2g

cartersocha:
I think it's because some of the upstream repos have branch protections and some sort of access restrictions

cartersocha:
@GaryPWhite just tried this on my mac and it was broken. I'm getting this error: "cache export feature is currently not supported for docker driver. Please switch to a different driver"

Any tips? I made sure docker is updated and couldn't find anything specific on Google

PR #307, opened 2022-08-16
fix diagram: services speak to featureflag server not its ui

tsloughter:
The UI is for a user to create/edit/toggle flags, services like the CheckoutService talk to the feature flag (grpc) server.


reyang (approved):

mic-max (approved):

mic-max:
Is it a fair assumption that eventually almost all of the services will interact with `featureflagbservice` in some way?

PR #308, opened 2022-08-16
Delete client programs

mic-max:
Deletes clients to the services `currencyservice` and `adservice`.
The functionality of these programs is covered by `/test/test.js` 

- TODO: Add baggage functionality added from `client.cpp` to `/test/test.js`

I'm interested in if people have opinions on this removal or if this repository should provide any instructions or examples on how to write a gRPC client or is the focus more-so on the services.


austinlparker:
Uh, I'd expect other services to be using those clients... did the frontend use them before being refactored?

xoscar (approved):

mic-max:
> Uh, I'd expect other services to be using those clients... did the frontend use them before being refactored?

The other services create their own clients from the protocol buffer definition file. 
- Ad Service
  - https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/frontend/gateways/rpc/Ad.gateway.ts
- Currency Service
 - https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/frontend/gateways/rpc/Currency.gateway.ts
 - https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/checkoutservice/main.go#L356

puckpuck:
I'm all for removing the clients from this. They are meant to be used if you are locally developing just a specific service. This is a project to showcase OpenTelemetry data flowing through multiple services.

I think the clients will cause unnecessary confusion, and removing them helps avoid that confusion.

puckpuck (approved):

mic-max:
Converting to draft, have to remove clients from gradle and cmake

reyang:
Adding @DebajitDas @lalitb who originally introduced the C++ client file and reviewed the PR #189.

I think the clients were used by the developers who have quick local test (whether it's functionality or perf or stress). Maybe these developers would find it helpful to keep the files somewhere (e.g. under a test folder)?

lalitb:
> 

Yes. client code is handy for quick local tests, it would be good to have it in test directory, if it is causing any confusion.

puckpuck (approved):

PR #319, opened 2022-08-18
Add jaeger as datasource to Grafana

cedricziel:
This adds jaeger as a datasource for trace data to Grafana when provisioning.


mviitane:
Looks good! I think the grafana bullet point in the main README could also be updated to mention traces.

src/grafana/provisioning/datasources/default.yaml
```diff
@@ -1,7 +1,9 @@
 apiVersion: 1
 
 datasources:
-  - name: webstore
+  - name: webstore-metrics
```
reyang, line 4: Maybe:

```suggestion
  - name: Prometheus
```

src/grafana/provisioning/datasources/jaeger.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: 1
+
+datasources:
+  - name: webstore-traces
```
reyang, line 4: Maybe:

```suggestion
  - name: Jaeger
```

reyang (approved):

julianocosta89:
Any reason why do we need that in this demo?

cedricziel:
@julianocosta89 it's an out of the box feature of Grafana. - It enables viewing traces and metrics side by side and stays in the OpenSource track.

Jaeger is still the main store for the trace data. - Hence I dont think there is an issue adding this. It's lesser a thing we need in the demo, but we get it for free - why not enable it?

cedricziel:
Alright :) 

Sure thing - Grafana can provision dashboards from file as well - makes sense to add a set of curated dashboards to the demo.

cedricziel:
I would consider this one finished for now - dashboards would be a separate PR :)

fatsheep9146 (approved):

puckpuck (approved):

Issue #327, opened 2022-08-22
[Frontend] - Improvement


julianocosta89:
# Feature Request

At the moment there are a couple of things in the Frontend that looks broken, even though they are working as expected.
I'll list it here and maybe we could work on improving it as we go:

- [ ] No logo.
- [ ] Product images are cropped.
- [ ] When an order is completed, the user should get the order ID in the frontend.
- [ ] Whenever more than 1 product is bought, the checkout page has `Shipping Data` listed multiple times. 
- [ ] Also in the checkout page there is a link to `See more` that has different color and it's clickable, but the mouse hover behaves like a text selector, and not like expected for a link.
- [ ] When an item is added to the cart we get 2 scroll bars whenever clicking on the cart icon.

puckpuck:
Logo updated with #340

cartersocha:
@julianocosta89 are the unchecked items still open?

julianocosta89:
yes, they are all still pending

cartersocha:
@julianocosta89 can we add the typo when you click on the shopping cart icon? It says go to shipping cart which is technically kinda correct but really not haha. 
<img width="602" alt="Screenshot 2023-01-02 at 4 20 11 PM" src="https://user-images.githubusercontent.com/43380952/210286903-c9fb010f-0fad-404c-a85f-efbe8817b85a.png">

julianocosta89:
@cartersocha I've just updated it, and also added the page title in the #670.

julianocosta89:
Loved this one!
![Screenshot from 2023-01-19 22-15-27](https://user-images.githubusercontent.com/15364991/213562499-eabce5ed-654c-438f-a21e-12c7949d38f9.png)
Just got a discount 😅

austinlparker:
@julianocosta89 Can you give this a look and make sure it reflects the current reality? I'm going to pull this into 1.4.

cartersocha:
@martinkuba is potentially going to pick up some of these items when he has the time

julianocosta89:
@austinlparker yes, this list is up-to-date.
I've recently updated it.

austinlparker:
Closing this as part of issue cleanup. If there's anything that needs to be done, please create a new issue.

PR #341, opened 2022-08-24
[email] remove email body from log

puckpuck:
## Changes

This PR is meant to help clean up the log output when doing a `docker compose up`. The email service adds an entry for the entire contents of the order confirmation email, creating unnecessary noise in the logs.

- removes email body from log output

### DIFF
--- a/src/emailservice/email_server.rb
+++ b/src/emailservice/email_server.rb
@@ -38,9 +38,10 @@ def send_email(data)
       from:     "noreply@example.com",
       subject:  "Your confirmation email",
       body:     erb(:confirmation, locals: { order: data.order }),
-      via:      :logger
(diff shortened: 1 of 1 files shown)


cartersocha:
Not blocking this pr but we should find a way to integrate the email content into the ui. It feels weird to say we have an email service then never really show an output

reyang (approved):

julianocosta89 (approved):
🚀

julianocosta89:
> Not blocking this pr but we should find a way to integrate the email content into the ui. It feels weird to say we have an email service then never really show an output

I think we could simply add order number 123 and confirmation sent to xyz@example.com to the checkout view.
I'll add that to #327

PR #343, opened 2022-08-24
✨ introducing Tracetest to the project

xoscar:
## Changes

Introduces [Tracetest](https://tracetest.kubeshop.io/) to the project to execute, view, and assert traces and spans.


docker-compose.yml
```diff
@@ -14,16 +14,20 @@ services:
...
+    healthcheck:
+      test: ["CMD", "wget", "--spider", "localhost:16686"]
+      interval: 1s
+      timeout: 3s
+      retries: 60
 
   # Collector
   otelcol:
     image: otel/opentelemetry-collector-contrib:0.56.0
-    container_name: otel-col
```
xoscar, line 26: Having container names being part of each of the services can cause colliding issues with other applications running a the same time. I believe it is better to leave docker-compose to take care of the name.

docker-compose.yml
```diff
@@ -377,10 +365,43 @@ services:
...
-    container_name: grafana
     volumes:
       - ./src/grafana/grafana.ini:/etc/grafana/grafana.ini
       - ./src/grafana/provisioning/:/etc/grafana/provisioning/
     ports:
       - "${GRAFANA_SERVICE_PORT}:${GRAFANA_SERVICE_PORT}"
     logging: *logging
+
+  # Tracetest
+  tracetest:
```
xoscar, line 376: Here's everything that is needed to start using Tracetest in local development.
1. Tracetest.
2. Progress.

austinlparker:
FYI, this should be brought up at the next SIG meeting. There were some informal discussions about it in the past, and we were leaning "no" on including TraceTest upstream as it's not a CNCF project. You would, of course, be free to maintain a fork of this repository in your org with TraceTest integrated and we could track it in the readme.

PR #363, opened 2022-09-13
Fix gradlew to LF line-endings

### LINKED ISSUE #77: Fix `gradlew` commands on Windows
The `adservice` fails to build at the `RUN ./gradlew downloadRepos` step

You must run
- `./gradlew installDist` from `/src/adservice/`
- `./gradlew wrapper --gradle-version 7.4.2`

Maybe only one of the above commands is necessary...

Then the following command should work:
`docker compose build cartservice`

Proposed Fix:

Make the `docker compose up` the only required command to run the whole project, so this might be simply upgrading the jar file: `/src/adservice/gradle/wrapper/gradle-wrapper.jar` to the `7.4.2` version done in the command found above.

image
The other changes to `/src/adservice/gradle/wrapper/gradle-wrapper.properties` and `/src/adservice/gradlew` are only related to line endings:
image

mic-max:
Fixes #77

Now if someone on Windows clones the repository no additional steps need to be taken.

## Changes

- Added `.gitattributes` file with wildcard text files set to auto line ending and specifically just `gradlew` file to `LF`
- `git clone clone-tips https://github.com/mic-max/opentelemetry-demo-webstore.git`
- `docker compose up adservice --build`

Note: `--build` was included here just to make sure it is actually building and not using the cloud hosted image since there's some new work happening with that.

image


reyang (approved):

cartersocha:
🦞

PR #368, opened 2022-09-15
Add traffic to adservice & recommendataion service

### LINKED ISSUE #360: [Demo] Adservice & Recommendation Service not getting any traffic
# Bug Report

Which version of the demo you are using?  Latest version
## Symptom

With changes in frontend service, adservice & recommendation service stopped receiving traffic.

**What is the expected behavior?** Traffic to these service either from frontend or load generator

**What do you expect to see?** Traces from adservice & recommendation service 

**What is the actual behavior?** No traffic & traces from these services

## Reproduce

When we run `docker compose up` .

dineshg13:
Fixes #360 

## Changes

1. Add .venv to gitignore 
2. Format locustfile.py using [black](https://pypi.org/project/black/)  formatter
3. Add additional routes in loadgenerator to hit frontend ads & recommendation endpoints 
4. Fix container name in `docker-compose`
5. Add `PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python` to ensure recommendation-service comes up. 

After above changes , we are able to see traffic to adservice & recommendation service. 

Screen Shot 2022-09-15 at 5 46 52 PM

Screen Shot 2022-09-15 at 5 45 46 PM


puckpuck (approved):
This looks great, thank you for this!

mic-max (approved):
LGTM, why change the `recommendationservice`'s `container_name` though? The main reason why these were added was just for readability of the console logs. Trying to keep consistent hyphenated service name for them

dineshg13:
Because of hyphen in container name, the address of container would change . It needs to be consist with ones defined in `.env` file . https://github.com/open-telemetry/opentelemetry-demo/blob/860a35de9be667ff36788d3df09ed4a96b9290fd/.env#L44

mic-max:
> Because of hyphen in container name, the address of container would change . It needs to be consist with ones defined in `.env` file .
> 
> https://github.com/open-telemetry/opentelemetry-demo/blob/860a35de9be667ff36788d3df09ed4a96b9290fd/.env#L44

Were you seeing an error? All the other services addresses work and they use the unhyphenated version, which is the actual root name of the service, not the container name. I believe we want to use the service name here, if we scale up to multiple recommendation service containers running at the same time, I think this would only provide traffic to one or completely break if docker compose decides to name them recommendationservice-1 and recommendationservice-2

dineshg13:
@mic-max Thanks for catching this. I have revert the container name. Initially i was seeing errors during startup of `frontend`, but looks like they are just transient ones, and few minutes it seems to be working as expected.

cartersocha:
Thanks @dineshg13 🙏🏼

austinlparker:
This PR broke the load generator, primarily through _removing locust from requirements.txt_.

Reverting PR.

PR #370, opened 2022-09-16
add updated versions, async call to quoteservice

### LINKED ISSUE #57: Review and enhance tracing support for admin service (PHP)
The following is a list of requirements that we need to evaluate before declaring v1 for trace telemetry. These requirements are across the entire application; Not all services will meet all requirements. Determine the relevant features for this service. 

- [ ] Automatic Instrumentation is being used where appropriate.
- [x] Library instrumentation is being used if automatic instrumentation is unavailable.
- [x] Automatic and Manual Context Propagation are being demonstrated.
  - [x] Interprocess context propagation is demonstrated.
- [x] Telemetry is not being sampled upfront.
- [x] Telemetry is not being filtered upfront.
- [ ] Baggage is being set and read appropriately (i.e., baggage must be explicitly set as attributes on spans)

Referencing: https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/42

Dependent on this item: https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/34

GaryPWhite:
Fixes (part of) #57 .

## Changes

Added new versions for upstream dependencies, faster builds! (by a bit)

Service now utilizes `quoteservice` to fetch the quote information, sends along the price from quoteservice instead of making it up.

Notably, this is a different trace than what works for generating an ID -- because the async runtime / requests between services, this is a worthwhile difference to give examples of how `ContextGuard` does and does not work for the library. Fun stuff 😅 

* [x] Design discussion issue (N/A)


cartersocha:
Could you update our mermaid charts to reflect which service call the new quote service? The v1 architecture should have a php color and legend example + out of date service name / db

cartersocha:
the php color is already present but unused

classDef php fill:#4f5d95,color:white;

src/shippingservice/src/shipping_service/quote.rs
```diff
@@ -9,18 +13,54 @@ pub struct Quote {
...
+
+    let client = reqwest::Client::new();
+
+    let resp = client
+        .post(quote_service_addr)
+        .json(&reqbody)
+        .send()
+        .await?
+        .text_with_charset("utf-8")
+        .await?;
```
julianocosta89, line 56:
```suggestion
    let req = client.request(Method::POST, quote_service_addr);

    let mut headers = HeaderMap::new();

    let cx = Context::current();
    global::get_text_map_propagator(|propagator| {
        propagator.inject_context(&cx, &mut HeaderInjector(&mut headers))
    });

    let resp = req
        .json(&reqbody)
        .headers(headers)
        .send()
        .await?
        .text_with_charset("utf-8")
        .await?;
```

julianocosta89 (changes requested):
@GaryPWhite thank you very much for tackling this one.
When testing the commit I saw that the span from ShippingService wasn't being linked with QuoteService.
I have added some suggestions to inject the context into the headers of the request, please feel free to modify it if you think that the way I've suggested is not the best one.

If you apply my suggestions, you may need to update the `Cargo.lock` file as well, as I've added a new dependency to manage the context propagation.

I've tested it locally and got everything stitched together 🤩

julianocosta89 (approved):
Awesome! 🚀

PR #372, opened 2022-09-17
[cartservice] GetCart should be async?

cartermp:
The code used `.Result`, which blocks the current thread rather than allowing it to await a result. Since the interface returns a `Task<Cart>` I think this was probably just an omission.


cartersocha:
@mic-max, @cijothomas

cijothomas (approved):

cartersocha (approved):
Thanks for catching this @cartermp

PR #380, opened 2022-09-20
bumping protobuf extension

brettmc:
## Changes

pecl is installing an older version of protobuf by default, that does not work in php8.1. Force it to 3.21.5


bobstrecansky (approved):

cartersocha (approved):
Thanks brett

PR #387, opened 2022-09-27
[payment] document instrumentation

puckpuck:
Adds documentation for payment service, focused on instrumentation code.

This service makes use of Baggage which should get its own section. I added it after Metrics and Logs which are currently marked TBD.  not sure if we go with this or a different approach.

### DIFF
--- a/docs/services/paymentservice.md
+++ b/docs/services/paymentservice.md
@@ -13,7 +13,7 @@ application to initialize the SDK and auto-instrumentation. When initializing
 the OpenTelemetry NodeJS SDK, you optionally specify which auto-instrumentation
 libraries to leverage, or make use of the `getNodeAutoInstrumentations()`
 function which includes most popular frameworks. The `tracing.js` contains all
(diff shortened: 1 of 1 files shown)


docs/services/paymentservice.md
```diff
@@ -0,0 +1,96 @@
...
+
+```javascript
+    span.setAttributes({
+      'app.payment.amount': parseFloat(`${amount.units}.${amount.nanos}`)
+    })
+```
+
+### Span Exceptions and status
+
+You can use the span object's `recordExcpetion` function to create a span event
```
mviitane, line 63: -> recordException

docs/services/paymentservice.md
```diff
@@ -0,0 +1,96 @@
...
+with the full stack trace of a handled error. When recording an exception also
+be sure to the span's status accordingly. You can see this in the
```
mviitane, line 65: 'be sure to **set**'? The last word is missing.

mviitane (approved):
Looks very good! Only few minor comments.

PR #389, opened 2022-09-28
feat(docker-compose): add temporality preference environment variable

### LINKED ISSUE #383: Pass temporality preference environment variable though to services
# Feature Request

**Is your feature request related to a problem?**

At this point in time, the demo cannot be easily adapted to export delta metrics instead of cumulative. Once more metrics are being added this might make the demo more difficult to use for people that want to try out different backends.

It would be great to have the `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` passed through to services on the `docker-compose.yml` file, so that temporality can be more easily changed, but keep cumulative as a default.

**Describe alternatives you've considered.**

My alternative now is to provide more setup instructions with the demo, which basically just pass through the environment variable to all the applicable services. 

# 
I can work on this issue if this is something we'd want in this repo. :slightly_smiling_face:

pichlermarc:
## Changes

This PR adds the `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` environment variable to the `.env` file, and the `adservice` in `docker-compose.yml`.

Only `adservice` was chosen, as this is, as far as I can tell the service that is exporting metrics at the moment. The `spanmetrics` processor on the collector config will not be changed as it is configured to export to prometheus anyway. 

Fixes #383


julianocosta89 (approved):
LGTM!

reyang (approved):

.env
```diff
@@ -7,6 +7,9 @@ OTEL_EXPORTER_OTLP_ENDPOINT=http://otelcol:4317
 OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=${OTEL_EXPORTER_OTLP_ENDPOINT}
 OTEL_EXPORTER_OTLP_METRICS_ENDPOINT=${OTEL_EXPORTER_OTLP_ENDPOINT}
 
+# Metrics Temporality
+OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE=cumulative
```
reyang, line 11: Do we want delta to be the default?

cartersocha:
We're planning on adding metric support to the cart service, checkout service, and the Python / Java services I can't remember off the top of my head too!

.env
puckpuck: The spec default for this var is `cumulative` https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk_exporters/otlp.md#additional-configuration

mviitane:
> We're planning on adding metric support to the cart service, checkout service, and the Python / Java services I can't remember off the top of my head too!

.env
pichlermarc: As @puckpuck said, I'm using `cumulative` as the default since that is what the spec defines. Also since the default setup is exporting to Prometheus, I think we should keep it at cumulative. :slightly_smiling_face: 

The `prometheus` collector exporter used by this demo [does drop delta histograms.](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/d6c3f1acc69eb40161b678ff94afaab2efb50481/exporter/prometheusexporter/accumulator.go#L238-L241) so we might have missing data there if we'd set it to delta by default.

pichlermarc:
@cartersocha I thought of adding it to other services as well, but as they're not exporting metrics right now I did not do it yet. Do you think I should add the env var to the services that you mentioned, or is that something that the authors of follow up PRs should do when they add metrics? :slightly_smiling_face:

mviitane (approved):

cartersocha:
@pichlermarc we can leave it for now! Feel free to add the basic meter provider + instrumentation libraries to any of the above mentioned services tho :)

PR #390, opened 2022-09-28
[cart] document instrumentation

puckpuck:
Adds documentation for cart service, focused on instrumentation code.


mic-max (approved):

cartersocha:
@puckpuck what are your thoughts on adding language to the file names / titles?

I feel like a user would prefer to see a language and go directly there rather than consult our architecture for direction

mviitane:
> I feel like a user would prefer to see a language and go directly there rather than consult our architecture for direction

Seems like a great idea! The language could also be mentioned in the top section of each doc, for example, next to the source code link.

mviitane (approved):
LGTM

puckpuck:
> @puckpuck what are your thoughts on adding language to the file names / titles?
> 
> I feel like a user would prefer to see a language and go directly there rather than consult our architecture for direction

I think this is a great idea.

`<servicename>_<language>.md` make sense for the file names?

I will also add it to the title for each one also.  I need to make a few small changes to this PR so will do that here, and submit another one to clean up all the other docs we wrote thus far.

cartersocha:
Awesome!

docs/services/cartservice.md
```diff
@@ -0,0 +1,69 @@
...
+This service maintains items placed in the shopping cart by users. It interacts
+with a Redis caching service for fast access to shopping cart data.
+
+[Cart service source](../../src/cartservice/)
+
+## SDK Initialization
+
+he OpenTelemetry dotnet SDK should be initialized in your application's
+`Startup.cs` as part of the `ConfigureServices` function. When initializing,
+optionally specify which auto-instrumentation libraries to leverage. The SDK is
```
cijothomas, line 12:
```suggestion
optionally specify which instrumentation libraries to leverage. The SDK is
```

cijothomas: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md#instrumentation-library

docs/services/cartservice.md
```diff
@@ -0,0 +1,69 @@
...
+initialized using a builder pattern, where you add each auto-instrumentation
```
cijothomas, line 13:
```suggestion
initialized using a builder pattern, where you add each instrumentation library
```

cijothomas (approved):
LGTM. Please do replace auto-instrumentation with instrumentation library.

cartersocha:
> LGTM. Please do replace auto-instrumentation with instrumentation library.

We'll want to update this in all service docs but can wait until we finish adding them

PR #405, opened 2022-10-06
Update project images

### LINKED ISSUE #31: Improve the images
Currently the screenshots look bad under dark theme

image

julianocosta89:
Fixes #31 .

## Changes

Updated images to reflect the current state of the project.


reyang:
I have a different suggestion, although I'm not sure if that's the common practice:

Given we don't intend to make modifications on these images, I wonder if they should be kept out of the repo (so we don't need to keep these binaries in the history), and only referenced in the markdown files using some permanent links.

julianocosta89:
thx @cartersocha, I've missed that one.

julianocosta89:
> ifferent suggestion, although I'm not sure if that's the common practice:
> 
> Given we don't intend to make modifications on these images, I wonder if they should be kept out of the repo (so we don't need to keep these binaries in the history), and only referenced in the markdown files using some permanent links.
> 
> For example, all the images in this doc works fine without having to be part of the repo

Sounds good @reyang, let me try this.

julianocosta89:
Added links and removed files. Thx for the suggestion @reyang!

reyang (approved):

mic-max (approved):

PR #408, opened 2022-10-07
try to replicate julianos pr error

cartersocha:
check Julianos error


cartersocha:
@julianocosta89 I think your fork might be the drama 😆 

No account error here

julianocosta89:
Most probably a hiccup from GitHub.
Re-ran the tests this morning and both PRs went through.

PR #413, opened 2022-10-07
add language to doc names and titles

cartersocha:
Add language related info to service docs


puckpuck (approved):

reyang:
I'm on the fence. I can see in the future we might have a single service demonstrating language interop, for example, a Python service calling a C++ library via FFI (foreign function interface). Changing the file names could make it harder to maintain/organize.

cartersocha:
> I'm on the fence. I can see in the future we might have a single service demonstrating language interop, for example, a Python service calling a C++ library via FFI (foreign function interface). Changing the file names could make it harder to maintain/organize.

I wish we could add some sort of tag to the documents to give users a visual signal in the repo on which language is in which docs like issues. 

In general I'd prefer a clear user signal at the cost of potential future conflicts. Excluding a hosted version + a Swift app it's not clear we'll be adding more non telemetry features. Thoughts?

austinlparker:
> I'm on the fence. I can see in the future we might have a single service demonstrating language interop, for example, a Python service calling a C++ library via FFI (foreign function interface). Changing the file names could make it harder to maintain/organize.

I'm kinda with you here; Right now we don't really demo _multiple_ methods of instrumentation (i.e., I don't think there's a python service that demos the otel auto instrumentation, right?) so you can see a future where there's different versions of the same thing, or multiple python services instrumented different ways, etc.

cartersocha:
We can discuss in the sig. closing for now

reyang:
Sounds like we can call it a week 😆

image

PR #417, opened 2022-10-08
Fix recommendation logic not to recommend the selected product

mviitane:
## Changes

- There was a fault in the recommendation logic that it offered also the same products which the user just opened (product page) or added to the cart (shopping cart page). The filtering logic not to recommend the same products was fixed.
- Minor refactoring.


julianocosta89 (approved):
LGTM! 🚀

PR #431, opened 2022-10-11
add language to service docs

cartersocha:
Re-opening this to have a discussion on doc structure. At the sig meeting the consensus was that language was worthwhile to add but in previous PR there were concerns over future proofing. 

@reyang, @austinlparker, @puckpuck, @julianocosta89

### DIFF
--- a/docs/services/adservice-java.md
+++ b/docs/services/adservice-java.md
@@ -1,4 +1,4 @@
-# ad service
+# Ad Service (Java)
 
 This service determines appropriate ads to serve to users based on context
 keys. The ads will be for products available in the store.
--- a/docs/services/cartservice-dotnet.md
+++ b/docs/services/cartservice-dotnet.md
@@ -1,4 +1,4 @@
-# cart service
+# Cart Service (Dotnet)
 
 This service maintains items placed in the shopping cart by users. It interacts
 with a Redis caching service for fast access to shopping cart data.


austinlparker:
I don't really have anything to add other than what I noted last time; We might eventually have different styles of the same language (i.e., `Python` vs. `Python Agent`) which we'd need to consider. I also don't necessarily understand the rationale behind having it in the title for each document, isn't the language kinda implicit by reading the code samples?

Another alternative to consider would be front matter like

---
language: python
Instrumentation: [manual]
---

etc.

reyang:
Linking to the actual related PR #413.

puckpuck:
The rationale of language first is that the user doesn't have domain context on which service is written in which language. They want to understand how the example was done in a specific language (ie: Javascript). Right now we require the user to first look at the service matrix to understand which services are implemented in Javascript then go to that service's file to see what was done.

Javascript and Go in particular each have more than 1 service, with different features implemented across both. So even if you map Javascript to a service, you actually will map it to 2 services, each with slightly different details and OpenTelemetry features.

Going with a language-first approach, we would mention which service has a specific feature implemented and would help solve the situation where we implemented some OpenTelemetry features in one service but not the other of the same language.

I recommend we organize and name the files `<language>.md` and within each file reference the necessary service/file where the code snippets can be found.

reyang:
> The rationale of language first is that the user doesn't have domain context on which service is written in which language. They want to understand how the example was done in a specific language (ie: Javascript). Right now we require the user to first look at the service matrix to understand which services are implemented in Javascript then go to that service's file to see what was done.
> 
> Javascript and Go in particular each have more than 1 service, with different features implemented across both. So even if you map Javascript to a service, you actually will map it to 2 services, each with slightly different details and OpenTelemetry features.
> 
> Going with a language-first approach, we would mention which service has a specific feature implemented and would help solve the situation where we implemented some OpenTelemetry features in one service but not the other of the same language.
> 
> I recommend we organize and name the files `<language>.md` and within each file reference the necessary service/file where the code snippets can be found.

This sounds like a good idea to me. @puckpuck would you suggest having separate docs for manual and auto-instrumentation? I guess no but would like to confirm (since we can always have two sections in a `python.md` doc covering different ways of using OpenTelemetry).

mviitane:
What about solving the problem for the _language first_ need with a table like this (with links to service docs):
image

We would avoid the major docs restructuring. Also, I think there’s a need for the existing _service first_ docs (I wouldn’t like to drop them).

puckpuck:
We create sections as appropriate in the docs in a similar fashion we have today to describe what is instrumented automatically, with libraries, or manually.

cartersocha:
A small re-purposing makes sense. I can think of a couple different options. 

1. We could combine the service table, some basic glossary linking, and the proposed language table into a docs folder readme. 

2. We make a stand-alone language table doc and spotlight it through linking in other files

3. We add it to the main readme

4. Other suggestions

cartersocha:
This new readme would be a natural entry or linking point for this info: 

https://github.com/open-telemetry/opentelemetry-demo/pull/438/files

austinlparker:
I think the proposal in https://github.com/open-telemetry/opentelemetry-demo/pull/431#issuecomment-1275769502 makes sense; Can we get a new PR today to add a table to the new `/docs` readme.md, or repurpose this PR to those ends?

cartersocha:
I’ll add the table to the docs readme today. Either a fresh pr or re using this one

PR #432, opened 2022-10-11
Integrate with existing metrics code

### LINKED ISSUE #72: Review and enhance metric support for recommendation service (Python)
The following are requirements for metrics in the project. Not all services will meet all requirements.

- [ ] Service metrics should not be duplicative of trace-derived metrics.
- [ ] Services should emit multiple business metrics where appropriate.
- [ ] Services should emit trace exemplars for their metrics.
- [ ] Services should use push metrics where possible.
- [ ] Services should explicitly read baggage where needed.
- [ ] Collectors should perform metric transforms to normalize resource and service metrics.

Referencing issue: https://github.com/open-telemetry/opentelemetry-demo-webstore/issues/43

avillela:
Fixes #72 

## Changes

* Clean up `requirements.txt`
   * Use `opentelemetry-distro` to install relevant OpenTelemetry packages
   * Since we are installing `psutil`, `opentelemetry-bootstrap` will also install `opentelemetry-instrumentation-system-metrics`
* Update `recommendationservice`'s `Dockerfile` to:
   * Install specific version of `grpcio-tools` to avoid compatibility issues with `protobuf` library versions.
   * Install `opentelemetry-bootstrap` to install auto-instrumentation libraries
* Clean up metrics and tracing init
* Move metrics definitions to `metrics.py`

The following metrics are available through auto-instrumentation, courtesy of the `opentelemetry-instrumentation-system-metrics`, which is installed as part of `opentelemetry-bootstrap` on building the recommendationservice Docker image:

* runtime.cpython.cpu_time
* runtime.cpython.memory
* runtime.cpython.gc_count


cartersocha:
Please update the metric features doc under the docs folder to reflect the added feature too 😃

cartersocha:
I'm getting an odd error. Cursory google searching doesn't show a clear answer.

#0 13.82 × Encountered error while trying to install package.
#0 13.82 ╰─> psutil

I installed the package but still getting an error. Any special dependency requirements?

docker-compose.yml
```diff
@@ -331,10 +331,9 @@ services:
       - RECOMMENDATION_SERVICE_PORT
       - PRODUCT_CATALOG_SERVICE_ADDR
       - OTEL_PYTHON_LOG_CORRELATION=true
-      - OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
-      # OTEL_EXPORTER_OTLP_METRICS_ENDPOINT # Not working for Python OTLP exporter
+      - OTEL_METRICS_EXPORTER=console,otlp
       - OTEL_EXPORTER_OTLP_ENDPOINT
-      - OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE
```
mviitane, line 337: This env variable is needed in order to change the default temporality, to verify 'delta' temporality. This should be kept.

docker-compose.yml
```diff
@@ -331,10 +331,9 @@ services:
...
+      - OTEL_EXPORTER_OTLP_INSECURE=true
```
mviitane, line 336: I don't see any reason to change to insecure.

src/recommendationservice/metrics.py
```diff
@@ -0,0 +1,100 @@
...
+from opentelemetry.metrics import (
+    CallbackOptions,
+    Observation,
+)
+
+# RAM usage
+def ram_usage_callback(options: CallbackOptions) -> Iterable[Observation]:
+    observations = []    
+    ram_percent = psutil.virtual_memory().percent
+    print(f"ram_percent: {ram_percent}")
```
mviitane, line 16: I think this is too much to log.

src/recommendationservice/metrics.py
```diff
@@ -0,0 +1,100 @@
...
+            observations.append(Observation(int(states[1]) // 100, {"cpu": cpu, "state": "nice"}))
+            observations.append(Observation(int(states[2]) // 100, {"cpu": cpu, "state": "system"}))
+            # ... other states
+    return observations
+
+# CPU usage
+def cpu_usage_callback(options: CallbackOptions) -> Iterable[Observation]:
+    observations = []    
+    for (number, percent) in enumerate(psutil.cpu_percent(percpu=True)):
+        print(f"cpu_number: {number}, cpu_percent: {percent}")
```
mviitane, line 40: I think this is too much to log.

src/recommendationservice/metrics.py
```diff
@@ -0,0 +1,100 @@
...
+    
+    rec_svc_metrics = {
+        "list_recommendations_request_counter": list_recommendations_request_counter,
+        "attributes": attributes,
+        "app_recommendations_counter": app_recommendations_counter,
+        "cpu_usage": cpu_usage,
+        "ram_usage": ram_usage
+    }
+    
+    return rec_svc_metrics
```
julianocosta89, line 100: You need a blank line at the end of the file.

src/recommendationservice/metrics.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def observable_gauge_callback(options: CallbackOptions) -> Iterable[Observation]:
+    yield Observation(9, {})
+
+
+def init_metrics(meter):
+
+    # Requests counter
+    list_recommendations_request_counter = meter.create_counter(
```
mviitane, line 63: To align with the other manual counter, I'd change the name to
`app_recommendation_requests_counter`
and
`app.recommendation.requests.counter`
(but I'm fine with the current logic as well)

src/recommendationservice/recommendation_server.py
```diff
@@ -88,6 +83,9 @@ def get_product_list(request_product_ids):
         indices = random.sample(range(num_products), num_return)
         # Fetch product ids from indices
         prod_list = [filtered_products[i] for i in indices]
+        
+        span.set_attribute("app.filtered_products.list", prod_list)
```
julianocosta89, line 87:
```suggestion
        num_products = len(prod_list)
        span.set_attribute("app.filtered_products.count", num_products)
```
According to: https://github.com/open-telemetry/opentelemetry-demo/blob/main/docs/manual_span_attributes.md#recommendationservice
 The attribute is called `app.filtered_products.count` and it is a number.

We can just update the docs if we would like to keep the new name and have it as a list.

src/recommendationservice/requirements.txt
```diff
@@ -1,13 +1,8 @@
...
-opentelemetry-instrumentation-grpc==0.34b0
-opentelemetry-instrumentation-urllib3==0.34b0
-opentelemetry-sdk==1.13.0
-python-dotenv==0.20.0
-python-json-logger==2.0.2
-requests==2.27.1
-urllib3==1.26.8
\ No newline at end of file
+opentelemetry-distro==0.34b0
+opentelemetry-exporter-otlp==1.13.0
```
mviitane, line 5: I'd continue using the previous lib:
`opentelemetry-exporter-otlp-proto-grpc`
just for clarity because we are not planning to export http.

docker-compose.yml
```diff
@@ -331,10 +331,9 @@ services:
       - RECOMMENDATION_SERVICE_PORT
       - PRODUCT_CATALOG_SERVICE_ADDR
       - OTEL_PYTHON_LOG_CORRELATION=true
-      - OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
-      # OTEL_EXPORTER_OTLP_METRICS_ENDPOINT # Not working for Python OTLP exporter
+      - OTEL_METRICS_EXPORTER=console,otlp
```
mviitane, line 334: Using `console` seems unnecessary, it just fills up the log.

julianocosta89 (changes requested):
I'm still not a heavy user of metrics so I'm not sure how it would be better to handle the naming.
I see here that we have `cpu_usage`, `ram_usage`.
Should we add an identifier to those metrics?

Just wondering how that will grow whenever we have 10 services with the same metric name.

mviitane (changes requested):

src/recommendationservice/metrics.py
mviitane: Yeah, I've noticed the same.

mviitane:
I think traces are not working with this PR without adding:
`OTEL_TRACES_EXPORTER` to docker-compose.yml

fatsheep9146:
I'm a little confused that why not using [python instrumentation library](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-system-metrics/src/opentelemetry/instrumentation/system_metrics/__init__.py) to get the system metric such as cpu usage. 

I'm not quite familiar about python, maybe my confusion is wrong, you can correct me. But what I really want to say is, when it come's to metric and trace, we should first consider using the instrumentation library instead of manually creating one.

reyang:
> But what I really want to say is, when it come's to add new metrics or traces, we should first consider using the instrumentation library instead of manually creating one.

💯

docker-compose.yml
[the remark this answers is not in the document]
avillela: My bad - that was an accident. Added it back in!

src/recommendationservice/metrics.py
avillela: Done!

docker-compose.yml
avillela: Fixed. I removed it by accident.

avillela:
> I'm a little confused that why not using [python instrumentation library](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-system-metrics/src/opentelemetry/instrumentation/system_metrics/__init__.py) to get the system metric such as cpu usage.
> 
> I'm not quite familiar about python, maybe my confusion is wrong, you can correct me. But what I really want to say is, when it come's to add new metrics or traces, we should first consider using the instrumentation library instead of manually creating one.

That was me just messing around with metrics and trying to understand them, and I forgot to remove the unnecessary stuff.

docker-compose.yml
avillela: Removed. :)

src/recommendationservice/recommendation_server.py
avillela: Not sure what the issue is here? There's both a `span.set_attribute("app.filtered_products.list", prod_list)` and `span.set_attribute("app.filtered_products.count", num_products)`

src/recommendationservice/metrics.py
avillela: I changed the names of both metrics to use "_" instead of "."

avillela:
> I'm still not a heavy user of metrics so I'm not sure how it would be better to handle the naming. I see here that we have `cpu_usage`, `ram_usage`. Should we add an identifier to those metrics?
> 
> Just wondering how that will grow whenever we have 10 services with the same metric name.

That waste messing around with learning metrics. I'd forgotten to remove these.

src/recommendationservice/recommendation_server.py
avillela: To clarify: I am not replacing `app.filtered_products.count` with  `app.filtered_products.list`. All I did was add another span attribute, `app.filtered_products.list`. If that is causing too much confusion, I can remove it.

julianocosta89 (approved):
Great one!

src/recommendationservice/metrics.py
```diff
@@ -0,0 +1,30 @@
...
+
+# from gc import callbacks
+# from typing import Iterable
+# import psutil
+
+
+def init_metrics(meter):
+
+    # Requests counter
+    list_recommendations_request_counter = meter.create_counter(
```
fatsheep9146, line 11: I think the metric `app_recommendations_request_counter` should also be replaced by instrumentation library. Since rpc server already has mature [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc-metrics.md). 

you can get the total requests amount from `rpc_server_duration_count`. 

But I saw that maybe [python grpc instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/__init__.py) library may not support meter, I think you can first contribute to that. =D

src/recommendationservice/metrics.py
```diff
@@ -0,0 +1,30 @@
...
+        name="app_recommendations_request_counter",
```
fatsheep9146, line 12: I think the metric app_recommendations_request_counter should also be replaced by instrumentation library. Since rpc server already has mature [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/semantic_conventions/rpc-metrics.md).

you can get the total requests amount from rpc_server_duration_count.

But I saw that maybe [python grpc instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-grpc/src/opentelemetry/instrumentation/grpc/__init__.py) library may not support meter, I think you can first contribute to that. =D

fatsheep9146 (changes requested):

src/recommendationservice/metrics.py
fatsheep9146: BTW, when I try to add metrics for checkoutservice, I also encounter this problem too. Now I'm pushing the enhancement of grpc golang instrumentation library to support metrics by PR https://github.com/open-telemetry/opentelemetry-go-contrib/pull/2700

src/recommendationservice/metrics.py
mviitane: I'm also suggesting to change "list_" -> "app_".
When we have all the services implemented with manual metrics, it becomes easy to find them with consistent naming = all manually created metrics start with "app_". We agreed this in the last SIG meeting.

mviitane:
> > I think traces are not working with this PR without adding:
> > `OTEL_TRACES_EXPORTER` to docker-compose.yml
> 
> I was able to send traces without issue...

OK, then it was something in my environment. It seems `otlp` is the default.

I actually think it makes sense to set both of these env variables or then none of them (just for consistency, I believe the end result will be the same):
OTEL_TRACES_EXPORTER=otlp
OTEL_METRICS_EXPORTER=otlp

avillela:
> > > I think traces are not working with this PR without adding:
> > > `OTEL_TRACES_EXPORTER` to docker-compose.yml
> > 
> > 
> > I was able to send traces without issue...
> 
> OK, then it was something in my environment. It seems `otlp` is the default.
> 
> I actually think it makes sense to set both of these env variables or then none of them (just for consistency, I believe the end result will be the same):
> 
> ```
> OTEL_TRACES_EXPORTER=otlp
> OTEL_METRICS_EXPORTER=otlp
> ```

Yes, good point. That was kinda bugging me too. :) Updated.

src/recommendationservice/metrics.py
avillela: Moot point now, since [@fatsheep9146 recommended that metric be removed](https://github.com/open-telemetry/opentelemetry-demo/pull/432#discussion_r994135299), since that should be handled by grace auto-instrumentation metrics (which are not yet in place).

mviitane (approved):
LGTM

src/recommendationservice/recommendation_server.py
fatsheep9146: Why add this attribute, if add all products items to this attributes, will the attributes be too verbosely? And I think this pr is about adding metrics, so the change of trace may better move to another new pr?

fatsheep9146:
https://github.com/open-telemetry/opentelemetry-demo/blob/main/docs/metric_service_features.md

I think this file can also be updated, since Recommendation service auto instrumentation is done.

avillela:
> https://github.com/open-telemetry/opentelemetry-demo/blob/main/docs/metric_service_features.md
> 
> I think this file can also be updated, since Recommendation service auto instrumentation is done.

Added

docs/services/recommendationservice.md
```diff
@@ -61,7 +58,25 @@ block ends execution. This is done in the `get_product_list` function.
...
+endpoints, resource attributes, and service name are automatically set by the
+OpenTelemetry auto instrumentor based on environment variables.
+
+```python
+    meter = metrics.get_meter_provider().get_meter("recommendationservice")
+```
+
+### Auto-instrumented metrics
+
+The following metrics are available through auto-instrumentation, courtesy of the `opentelemetry-instrumentation-system-metrics`, which is installed as part of `opentelemetry-bootstrap` on building the recommendationservice Docker image:
```
julianocosta89, line 74:
```suggestion
The following metrics are available through auto-instrumentation, courtesy of
the `opentelemetry-instrumentation-system-metrics`, which is installed as part
of `opentelemetry-bootstrap` on building the recommendationservice Docker image:
```

docs/services/recommendationservice.md
```diff
@@ -61,7 +58,25 @@ block ends execution. This is done in the `get_product_list` function.
...
+
+* runtime.cpython.cpu_time
+* runtime.cpython.memory
+* runtime.cpython.gc_count
+
```
julianocosta89, line 79:
```suggestion
```

docs/services/recommendationservice.md
julianocosta89: Text wasn't changed, just the line breaking of it.

docs/services/recommendationservice.md
julianocosta89, line 79: It would be great to have the custom metrics documented here as well.

julianocosta89 (changes requested):

docs/services/recommendationservice.md
avillela: Added

julianocosta89 (approved):

austinlparker:
is the requested change actually blocking anything? this needs to merge today to get into 0.6

fatsheep9146 (approved):

PR #436, opened 2022-10-13
Update Docker docs to use config-extras

austinlparker:
This way people won't clobber their config if we change it upstream and they pull


austinlparker:
branch protection :argh:

PR #437, opened 2022-10-13
Add Index to /docs

austinlparker:
Adds an index/table of contents to the /docs path so we can more easily link into it.


austinlparker:
branch protection!!! :argh:

Issue #396, opened 2022-10-02
FeatureFlagService fails to run on Mac M1 Pro


saurabhdes:
# Bug Report

Which version of the demo you are using? (please provide either a specific
Latest `main` or 
[b6e75ee](https://github.com/open-telemetry/opentelemetry-demo/commit/b6e75ee44cc41bb8315e5e116cc1ac4807ef5e1d)

## Symptom

**What is the expected behavior?**

FeatureFlag UI loads on going to http://localhost:8081/ so we can turn on different feature flags

**What is the actual behavior?**

FeatureFlag UI http://localhost:8081/ doesn't load on Mac M1 Pro

## Reproduce

- On Mac M1 Pro , follow the getting started steps for Docker https://github.com/open-telemetry/opentelemetry-demo/blob/main/docs/docker_deployment.md , essentially, `docker compose up --no-build`
- Wait about 5min for containers to start
- Notice featureflag container shows it's running
- Look at the logs of feature flag service, you will see 
`qemu: uncaught target signal 11 (Segmentation fault) - core dumped
Segmentation fault`
My understanding is qemu is the upstream component  used for running Intel (amd64) containers on M1 (arm64)
- Then go to http://localhost:8081/  and notice the page doesn't load as expected

## Additional Context

My best guess is the image available for featureflag service ghcr.io/open-telemetry/demo:v0.4.0-alpha-featureflagservice does not support ARM architecture. The same issue exists for other services but somehow they do seem to run successfully on M1.
refer below screenshot which shows a warning message on M1 for amd64 images.
<img width="909" alt="Screen Shot 2022-10-01 at 6 24 33 PM" src="https://user-images.githubusercontent.com/47896520/193434020-c174804c-7bb7-467f-821f-06c7eeb79997.png">

 I think we need to build multi-arch images for all services with buildx and then push them to ghcr.

janssenlima:
If you run `docker compose up --build` the image will be generated on your architecture and it will work. I went through this too and it's working.

saurabhdes:
@janssenlima That's correct and I am aware of it as well. It's a good workaround until this issue is resolved by building multi arch images that get pushed to ghcr as part of build pipeline.

joshleecreates:
This is resolved with @austinlparker's PR https://github.com/open-telemetry/opentelemetry-demo/pull/536

I built v1.0.0 on amd64 and this error occurs when running the image under emulation with Docker Desktop on apple silicon
[...]

PR #439, opened 2022-10-13
Update cartservice Dockerfile to support ARM64

chancez:
Use multiarch runtime-deps image, and configure the runtime identifier based on the images architecture using uname -m.


puckpuck:
Is the cart service the only one that is problematic with ARM ?

chancez:
So far the rest build fine on arm64 without changes. That said, the currently released images aren't all built for arm64 yet IIRC.

reyang (approved):

mic-max (approved):

saurabhdes:
that's correct @chancez - The currently released images are build to ARM. We have a bug for [ARM images not being build](https://github.com/open-telemetry/opentelemetry-demo/issues/396)   .

puckpuck (approved):

PR #385, opened 2022-09-27
remove docfx check

puckpuck:
docfx is failing on relative links to folders. We require these types of links for our service docs.

Per the SIG meeting on 9/26, we determined that the docfx checks were providing no value-add and should be removed.  This PR comments out the check (we can remove it completely on a subsequent PR)

PR #440, opened 2022-10-14
finish removing docfx check

puckpuck:
## Changes

This finishes removing the docfx check we had. It was commented out in #385 

A bonus side effect will be removing this GitHub beta warning in our PR code views.
Screen Shot 2022-10-13 at 10 19 42 PM


julianocosta89 (approved):
😍

PR #448, opened 2022-10-14
[frontend] - sort currency list

puckpuck:
## Changes

Sorts the list of currencies in the currency selector.
Screen Shot 2022-10-14 at 2 34 20 PM

### DIFF
--- a/src/frontend/providers/Currency.provider.tsx
+++ b/src/frontend/providers/Currency.provider.tsx
@@ -24,7 +24,7 @@ interface IProps {
 export const useCurrency = () => useContext(Context);
 
 const CurrencyProvider = ({ children }: IProps) => {
-  const { data: currencyCodeList = [] } = useQuery('currency', ApiGateway.getSupportedCurrencyList);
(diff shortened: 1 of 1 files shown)


reyang (approved):
Awesome!

src/frontend/providers/Currency.provider.tsx
```diff
@@ -36,13 +36,23 @@ const CurrencyProvider = ({ children }: IProps) => {
     SessionGateway.setSessionValue('currencyCode', currencyCode);
   }, []);
 
+  const currencyCodeList = currencyCodeListUnsorted.sort((n1, n2) => {
```
mic-max, line 39: I'm not sure what `currencyCodeListUnsorted` is seems just like an array of strings though. `Array.sort()` should suffice in that case.

mic-max (approved):
LGTM nice change, 1 comment about implementation detail

src/frontend/providers/Currency.provider.tsx
puckpuck: When I wrote the code, I felt like I was doing something unnatural... and yes `sort()` is all I needed without having to define actual compare logic.

Issue #450, opened 2022-10-15
Review relevants docs before v1 release (services)


cartersocha:
services:
- [ ] ad service
- [x] cart service
- [ ] checkout service
- [x] email service
- [x] feature flag service
- [ ] payment service
- [ ] product catalog service
- [ ] shipping service
- [x] recommendation service
- [x] quote service
- [ ] load gen
- [ ] frontend

cartersocha:
Reviews are mostly done

PR #451, opened 2022-10-15
load gen group review for v1

cartersocha:
Please review and suggest any needed changes in preparation for v1 release. 

We'll try to do this for most docs to get multiple reviewers. 

Contributing issue: https://github.com/open-telemetry/opentelemetry-demo/issues/450


cartersocha:
Lets aim for 3 approvals. We can do more or less based on cadence next week

PR #453, opened 2022-10-15
main readme group review for v1

cartersocha:
Please review and suggest any needed changes in preparation for v1 release.

We'll try to do this for most docs to get multiple reviewers.

Contributing issue: TBA

Original PR: https://github.com/open-telemetry/opentelemetry-demo/pull/451


cartersocha:
We could potentially tighten this up some more. Maybe link only to the docs readme instead of the contents list.

Features and tests could be segmented out as well. Open to suggestions

PR #460, opened 2022-10-17
set memory limits for all services

puckpuck:
## Changes
Sets reasonable memory limits for all services, based on an 80% memory target for each service, running the app for 7 days (minimum of 20M).

Screen Shot 2022-10-17 at 2 18 37 PM

Data used by this PR incorporated the memory limits shown in the chart above which is a 7-day run on EKS. Some exceptions for loadgenerator, email, and quote services where I used data from docker stats after running the app on my local machine for several days.

- loadgenerator - the above didn't incorporate the memory leak fix
- email - I guess we have another small bug in the Helm chart since the email service wasn't listening on the proper port and I didn't notice :(, instead I used data from docker stats. I think we also have an issue with this service since it continues to eat memory erratically after running for several hours.
- quote - I'm not sure why the memory just runs away, but if you put a limit on it this service would stay at 99% usage and be happy, so I picked 30M.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,3 +118,5 @@ significant modifications will be credited to OpenTelemetry Authors.
  [#432](https://github.com/open-telemetry/opentelemetry-demo/pull/432)
 * Replaced the Jaeger exporter to the OTLP exporter in the OTel Collector
 ([#435](https://github.com/open-telemetry/opentelemetry-demo/pull/435))
+* Set resource memory limits for all services
+([#460](https://github.com/open-telemetry/opentelemetry-demo/pull/460))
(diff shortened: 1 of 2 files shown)


mviitane (approved):
Looks good!

austinlparker (approved):

styblope:
feature-flag-service limit of 160M is too strict and causes the container to be Killed in my docker compose deploy. It may be environment specific, but a minimum limit I can get this working is 1600M.

puckpuck:
> feature-flag-service limit of 160M is too strict and causes the container to be Killed in my docker compose deploy. It may be environment specific, but a minimum limit I can get this working is 1600M.

@styblope 1600M is definitely not normal. Which OS is this on?

styblope:
I know, it's weird, but I stepped up piece by piece until the container stopped getting killed. It's on x86_64 Fedora 34 VM (vsphere), 8GB RAM, Docker CE 20.10.17. Maybe it's a docker thing?

PR #461, opened 2022-10-18
[currency] - remove baggage processing

puckpuck:
## Changes

Removes baggage processing from the currency service. Previously it would take all items from baggage and emit them as a span, which is not the desired behavior.


austinlparker (approved):
lgtm

mviitane (approved):

PR #496, opened 2022-10-20
Update .env for 0.6.1 release

austinlparker:

PR #501, opened 2022-10-20
Revert "Update .env for 0.6.1 release"

puckpuck:
Reverts open-telemetry/opentelemetry-demo#496

Until we get the `0.6.1-beta` images up, we need to revert back to the prior version, else `docker compose up --no-build` will fail.


austinlparker:
Instead of this can we revert the build change?

austinlparker:
Mooted by #502

PR #504, opened 2022-10-20
[shippingservice] add resource data

julianocosta89:
Fixes #.

## Changes

Shipping Service has no resource data.
I've used the Resource Detectors in order to add some data to it.
Unfortunately `telemetry.sdk` data is not part of it.

Sending the PR to maybe get some insights from the everyone.


julianocosta89:
This is what I have at the moment:
Screenshot from 2022-10-20 22-20-11

@TommyCpp and @GaryPWhite any ideas how can I add `telemetry.sdk*` to the spans?

I've played with the ones documented here, but no luck:
https://docs.rs/opentelemetry_sdk/0.18.0/opentelemetry_sdk/resource/trait.ResourceDetector.html#tymethod.detect

julianocosta89:
Also asked for some insights in the [otel-rust slack channel](https://cloud-native.slack.com/archives/C03GDP0H023/p1666355320238369).

TommyCpp:
https://github.com/open-telemetry/opentelemetry-rust/pull/899 should do the trick

TommyCpp:
I need to double-check and see if we can release a minor version to include this change but feel free to copy-paste the code in the PR I linked above (or I can do it as a follow-up PR if that's preferable)

austinlparker:
Did the upstream change ever go in? Could we update deps in lieu of this PR?

julianocosta89:
@austinlparker updated.
@TommyCpp 's change cannot be added here because we would need a new release from the OTel Rust, which we don't have.
I don't like the approach of hardcoding the telemetry data here, so I think we could merge this one, and whenever there is a new release from the Rust side, we just update it with the `TelemetryResourceDetector` configuration.

austinlparker (approved):

puckpuck (approved):

Issue #397, opened 2022-10-03
Update the frontend service to use 2 different exporters ( Client & Backend)


henrikrexed:
# Feature Request
Once the Demo is deployed, the browser is trying to send spans to localhost ( or the collector address).
To make it more realistic all the client side spans need to be sent to a specific url.

**Describe the solution you'd like:**

i have tried to resolve this  by creating a new Environment variable for the Frontend code  :
-> FrontendTracer.ts : 
 const provider = new WebTracerProvider({
    resource: new Resource({
      [SemanticResourceAttributes.SERVICE_NAME]: process.env.NEXT_PUBLIC_OTEL_SERVICE_NAME,
    }),
  });

What is expected is that all the spans generated by the backend will be sent to a local collector.
The spans produced by the end-user browser should use another endpoint to export the spans. 

In the context of k8s a specific ingress rule needs to be created to expose the collector out of the cluster.

cartersocha:
@xoscar could you potentially take a look at this?

xoscar:
Yep, this is part of what we discussed two weeks ago about adding new env vars to the helm chart, I'll take a look!

PR #505, opened 2022-10-20
Adding front-end exporter custom URL

xoscar:
Fixes https://github.com/open-telemetry/opentelemetry-demo/issues/397

## Changes
With the Front-end overhaul, we introduced the browser-side instrumentation but we were missing the dynamic URL for the exporter, this PR fixes that so it can work on environments like K8s.


julianocosta89 (approved):
LGTM

austinlparker (approved):

PR #508, opened 2022-10-21
[frontend] [ops golf?] Pass collector URL to frontend via reverse proxy.

austinlparker:
This fixes the current problem of the collector URL not being passed into the frontend, as Next.JS will not pass env vars into the client except ones known at build time.

It accomplishes this by way of establishing a reverse proxy that serves the Next.JS app (best practice!) and uses it to set a cookie that the client code reads on load.


austinlparker:
@puckpuck had an excellent idea to use envoy rather than nginx. there is one small catch here, that we have to build an envoy container in order to get `envsubst` but that's probably not the end of the world.

Someone more enterprising than I can enable tracing/metrics off this post v1.

xoscar (approved):

puckpuck (approved):
Just a few small naming things.

`src/frontendProxy` -> `src/frontend proxy` all lowercase. This is to stay consistent with the other services

cartersocha:
The frontend is broken on my machine? Is port 8080 still the right connection or do I need to navigate to the reverse proxy?

cartersocha (changes requested):
Small pause to prevent merging before localhost question is answered

cartersocha:
If we update the docker file to port 80 we should be good

cartersocha:
+ changelog

austinlparker:
> If we update the docker file to port 80 we should be good

what part of the dockerfile? the compose file or the frontend?

puckpuck:
We need to update the readmes to indicate we run on port 80 now

austinlparker:
Updated the docker readme with the current state

puckpuck:
Not for this PR, but it would be nice if there is a way to force mermaid to not draw a box in the top right corner that gets covered by the github icons
Screen Shot 2022-10-21 at 1 59 39 PM

PR #517, opened 2022-10-22
AdService tuneup including various upgrades and fixes to allow running locally

jack-berg:
Went and tried to run the adservice locally and ran into a number of issues.

## Changes

- Downgrade to latest java LTS version, 17. The adservice is currently on 18, which is awkward to use because tooling like [sdkman](https://sdkman.io/) include LTS versions along with the latest non-LTS version. The latest non-LTS version is java 19, which we could update to eventually but which currently causes issues the gradle build. Its very common for java shops to use the latest LTS.
- Use eclipse-temurin base images instead of openjdk. The openjdk images are [deprecated](https://hub.docker.com/_/openjdk). eclipse-temurin is the successor. 
- Add a gradle task for copying protos into the build directory. This is required to run locally outside of docker. Currently this fails.
- Fix readme instructions for building docker image. Include steps to run locally.
- Upgrade to otel java 1.19.0, and otel java instrumentation (annotations and java agent) 1.19.1. These are the latest versions.
- Upgrade to latest gradle version 7.5.1.


src/adservice/build.gradle
```diff
@@ -14,14 +14,15 @@ description = 'Ad Service'
...
+def opentelemetryInstrumentationAlphaVersion = "1.19.1-alpha"
 def grpcVersion = "1.45.1"
 def jacksonVersion = "2.13.2"
 def protocVersion = "3.20.0"
 
 tasks.withType(JavaCompile) {
-    sourceCompatibility = JavaVersion.VERSION_18
-    targetCompatibility = JavaVersion.VERSION_18
+    sourceCompatibility = JavaVersion.VERSION_17
+    targetCompatibility = JavaVersion.VERSION_17
```
trask, line 25: I agree that it's good for demo to use latest Java LTS👍

src/adservice/build.gradle
```diff
@@ -33,16 +34,18 @@ dependencies {
     if (speed) {
         implementation fileTree(dir: offlineCompile, include: '*.jar')
     } else {
+        implementation platform("io.opentelemetry:opentelemetry-bom:${opentelemetryVersion}")
+        implementation platform("io.opentelemetry.instrumentation:opentelemetry-instrumentation-bom-alpha:${opentelemetryInstrumentationAlphaVersion}")
```
trask, line 38: 👍

src/adservice/build.gradle
```diff
@@ -60,14 +84,20 @@ protobuf {
...
             grpc {}
         }
         ofSourceSet('main')
     }
 }
 
+afterEvaluate {
+    // Ensure protos are copy before classes are generated
+    tasks.getByName('processResources').dependsOn('copyProtos')
+    tasks.getByName('generateProto').dependsOn 'copyProtos'
```
trask, line 98: just to match line above
```suggestion
    tasks.getByName('generateProto').dependsOn('copyProtos')
```

trask (approved):
> The openjdk images are [deprecated](https://hub.docker.com/_/openjdk). eclipse-temurin is the successor.

oh! I hadn't seen that those images had been deprecated, I think we still use them in a couple of places in the java instrumentation repo, I'll look into updating those too 👍

src/adservice/build.gradle
trask: (one could even argue for it being Java 11 based on https://newrelic.com/resources/report/2022-state-of-java-ecosystem#toc-java-11-is-the-new-standard and the low Java 17 adoption rate, but I personally agree with using Java 17)

jack-berg: I think choosing the latest LTS is simple and defensively. Latest non-LTS version will be prone to issues with tooling, and choosing a non-latest LTS (i.e. 11) would seem kind of arbitrary.

mateuszrzeszutek (approved):

mic-max (approved):

reyang (approved):

cartersocha:
Thanks @jack-berg and @trask 🤩

PR #520, opened 2022-10-22
paymentsservice: structure logs

naseemkullah:
Use Pino's [mergingObject](https://github.com/pinojs/pino/blob/master/docs/api.md#mergingobject-object) parameter to add context as JSON
rather than in the message parameter, which is now more generic (like an event name).

Also factor out logger so it is instantiated once for the whole service and upgrade pino.

Also remove unneeded comments.

### DIFF
--- a/src/paymentservice/index.js
+++ b/src/paymentservice/index.js
@@ -12,17 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Npm
 const grpc = require('@grpc/grpc-js')
 const protoLoader = require('@grpc/proto-loader')
 const health = require('grpc-js-health-check')
 const opentelemetry = require('@opentelemetry/api')
-const pino = require('pino')
 
-// Local
 const charge = require('./charge')
+const logger = require('./logger')
 
-// Functions
 function chargeServiceHandler(call, callback) {
   const span = opentelemetry.trace.getActiveSpan();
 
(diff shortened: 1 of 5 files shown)


src/paymentservice/index.js
```diff
@@ -31,13 +28,13 @@ function chargeServiceHandler(call, callback) {
...
     })
-    logger.info(`PaymentService#Charge invoked by: ${JSON.stringify(call.request)}`)
+    logger.info({request: call.request}, "Charge request received.")
 
     const response = charge.charge(call.request)
     callback(null, response)
 
   } catch (err) {
-    logger.warn(err)
+    logger.warn({err})
```
naseemkullah, line 37: Pino applies their error serialization to Error objects named `err` out of the box 👍 

- https://github.com/pinojs/pino-std-serializers#exportserrerror
- https://getpino.io/#/docs/api?id=serializers-object

src/paymentservice/index.js
```diff
@@ -63,9 +57,9 @@ server.addService(health.service, new health.Implementation({
 
 server.addService(hipsterShopPackage.hipstershop.PaymentService.service, { charge: chargeServiceHandler })
 
-server.bindAsync(`0.0.0.0:${process.env['PAYMENT_SERVICE_PORT']}`, grpc.ServerCredentials.createInsecure(), (err, port) => {
+server.bindAsync(`0.0.0.0:${process.env['PAYMENT_SERVICE_PORT'] ?? 50051}`, grpc.ServerCredentials.createInsecure(), (err, port) => {
```
puckpuck, line 60: I don't think we want to hardcode a default port for any of the services outside of the .env file.
```suggestion
server.bindAsync(`0.0.0.0:${process.env['PAYMENT_SERVICE_PORT']}`, grpc.ServerCredentials.createInsecure(), (err, port) => {
```

puckpuck (approved):
please remove the hardcoded default port, the rest looks great.

Thanks!

naseemkullah:
> please remove the hardcoded default port, the rest looks great.
> 
> Thanks!

done, my pleasure

src/paymentservice/logger.js
```diff
@@ -0,0 +1,3 @@
+const pino = require('pino');
+
+module.exports =  pino();
```
mic-max, line 3:
```suggestion
module.exports = pino();
```

mic-max (approved):

mic-max:
LGTM!

reyang (approved):

austinlparker (approved):

PR #538, opened 2022-10-24
bump to v1.0 for release

austinlparker:


xoscar:
🔥

README.md
```diff
@@ -37,6 +37,9 @@ If you'd like to extend this demo or maintain a fork of it, please read our
 You can be up and running with the demo in a few minutes. Check out the docs for
 your preferred deployment method:
 
+> **Note:** If you're running on Apple Silicon, please run `docker compose
```
puckpuck, line 40: This note should be in docker_deployment.md instead.

puckpuck (changes requested):

README.md
austinlparker: Mmmmm, I was worried that people wouldn't read it if it was there since `docker compose up` is pretty intuitive?

puckpuck: Since this note is referencing the `docker compose up` command it should be next to where that command is located.  Being in the main readme, disconnects it from docker as a whole.

austinlparker: moved

puckpuck (approved):

PR #540, opened 2022-10-24
fix dashboard bugs

### LINKED ISSUE #539: Grafana: Demo/OpenTelemetry Collector fails to load
# Bug Report

v1.0.0

## Symptom

When navigating over to the Demo/OpenTelemetry Collector an error of  `Templating Failed to upgrade legacy queries Datasource ${DS_PROMETHEUS} was not found` pops up in the top right.   None of the graphs load.   I could be missing something silly that I should have been doing as well, but I've reproduced this twice in a row locally. 

Graphs would all load without error. 

An error message of `Templating Failed to upgrade legacy queries Datasource ${DS_PROMETHEUS} was not found` is produced in the UI. 

## Reproduce

After starting up the docker compose go to: http://localhost:8080/grafana/d/BKf2sowmj/opentelemetry-collector?orgId=1&from=now-5m&to=now

fatsheep9146:

Fixes #539

### DIFF
--- a/src/grafana/provisioning/dashboards/general/opentelemetry-collector.json
+++ b/src/grafana/provisioning/dashboards/general/opentelemetry-collector.json
@@ -58,7 +58,7 @@
   "panels": [
     {
       "collapsed": false,
-      "datasource": "${DS_PROMETHEUS}",
+      "datasource": "Prometheus",
       "gridPos": {
         "h": 1,
         "w": 24,
@@ -75,7 +75,7 @@
       "bars": false,
       "dashLength": 10,
       "dashes": false,
-      "datasource": "${DS_PROMETHEUS}",
+      "datasource": "Prometheus",
       "description": "Rate of spans successfully accepted vs refused per second",
       "fieldConfig": {
         "defaults": {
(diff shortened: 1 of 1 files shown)


julianocosta89:
I will leave this PR to someone with more experience in Grafana to review.
Personally I'd prefer keeping the environment variable, because that allows us to easily change the datasource, but I'm not sure if it is doable.

Anyways, thanks for taking care of that @fatsheep9146!

fatsheep9146:
> I will leave this PR to someone with more experience in Grafana to review. Personally I'd prefer keeping the environment variable, because that allows us to easily change the datasource, but I'm not sure if it is doable.
> 
> Anyways, thanks for taking care of that @fatsheep9146!

I agree with your opnion about keeping environment variable. But for now, the problem is that the environment variable is not correctly configured which result that the dashboard could not be opened rightly and throwing out an error `emplating Failed to upgrade legacy queries Datasource ${DS_PROMETHEUS} was not found`. 

So I think maybe it's more suitable to use const for now to let user can open the dashboard without failure. And then try to improve the dashboard with configurable datasource variable. =D

julianocosta89:
Agreed

julianocosta89 (approved):

austinlparker (approved):

PR #546, opened 2022-10-31
Improve build times for JS and PHP

austinlparker:
In support of #458, this PR is making an attempt to improve build times by optimizing gRPC libs where possible/easy.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -339,7 +339,7 @@ services:
       - OTEL_EXPORTER_OTLP_ENDPOINT=otelcol:4317
       - OTEL_TRACES_SAMPLER=parentbased_always_on
       - OTEL_TRACES_EXPORTER=otlp
-      - OTEL_EXPORTER_OTLP_TRACES_PROTOCOL=grpc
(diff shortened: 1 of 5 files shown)


reyang (approved):

puckpuck (approved):

julianocosta89:
@austinlparker this PR broke the PHP traces.
No spans are being exported anymore.

austinlparker:
That's weird, it worked locally.

reyang:
Would folks be interested in a CI job which captures traces across all services as part of the integration test? (I think integration test which covers telemetry correctness is an amazing thing to demonstrate, seems a good fit)

austinlparker:
Yeah, that’d be good if we could get build times down… I don’t think it’s reasonable until then though. Sent from my iPhoneOn Nov 2, 2022, at 7:25 PM, Reiley Yang ***@***.***> wrote:﻿
Would folks be interested in a CI job which captures traces across all services as part of the integration test? (I think integration test which covers telemetry correctness is an amazing thing to demonstrate, seems a good fit)

PR #548, opened 2022-11-01
[checkoutservice] add http protocol when posting request to emailservice

yuxiaoba:
The checkoutservice meets one error in its logs 

```
{"message":"failed to send order confirmation to \"someone@example.com\": failed POST to email service: Post \"emailservice:8080/send_order_confirmation\": unsupported protocol scheme \"emailservice\"","severity":"warning","timestamp":"2022-10-31T11:58:16.477407728Z"}
```

I fix this error by adding http protocol when posting request to emailservice


puckpuck:
This is set in the environment variable itself that gets passed into the checkout service.  No need to hardcode this.

https://github.com/open-telemetry/opentelemetry-demo/blob/main/.env#L40

PR #549, opened 2022-11-01
Enhance architecure diagram

### LINKED ISSUE #104: Enhance the architecture diagram with OpenTelemetry elements
The current micro-services is a great overview of our application but is missing OpenTelmetry components. To give users a full perspective of our environment we need to add these OTel components to an enhanced architecture diagram using mermaid. 

https://github.com/open-telemetry/opentelemetry-demo-webstore/blob/main/docs/v1Graph.md

algebrajones:
Fixes #104 

## Changes

Added two alternative architecture diagram enhancements
1. OTEL compoenents shown as part of the main architectural diagram
2. OTEL componenets shown as a seperate diagrm


austinlparker:
Thanks for the contribution! This looks pretty good. Is there a way to tweak the contrast on the background at all? Right now, on the topmost image, going between the grey and black looks a bit odd.

puckpuck:
I'm not a big fan of the variation on the top graph. The top graph should be about the actual application service connections, and not necessarily the telemetry flow.  The bottom chart represents the telemetry flow.

reyang:
Thanks for the contribution @algebrajones!

It's hard to comment on a graph so I'll try a different approach:

1. Would it help if we callout that all the color-coded components are instrumented (any grey/black components are not instrumented)

    image

2. Then in the telemetry flow diagram, we can abstract it out a bit, e.g. "all the instrumented components are sending data to Collector", which helps to make the diagram more readable.

puckpuck:
@reyang's approach makes lots of sense and also allows us to easily document (perhaps optional styling) to indicate where vendors would plug into the Collector to export to their backend

austinlparker:
+1 to @reyang's approach -- less busy, which is nice!

algebrajones:
I really like the proposal @reyang I'll work on putting something together like that. Have to admit I prefer the second option, and as @puckpuck says it gives that info on how other vendors can link into the demo.

I'll link out to the actual documentation on what is instrumented i.e. [metrics](https://github.com/open-telemetry/opentelemetry-demo/blob/main/docs/metric_service_features.md) and [traces](https://github.com/open-telemetry/opentelemetry-demo/blob/main/docs/trace_service_features.md), IIRC the collector isn't setup for log collection currently, I'll double check that as I document this though.

algebrajones:
🤞 this works.
@puckpuck I've not added any custom styling on the exporters, but I have added text and links above the diagram to say where the exporters are configured.

To save people jumping around here's an image of the new version
mermaid-diagram-2022-11-09-222431

cartersocha:
Small nit. The titles of the colored boxes fades in dark mode. 

But otherwise looks excellent! I’m a fan 🚀

reyang (approved):
Looks great to me.

docs/current_architecture.md
```diff
@@ -91,4 +91,80 @@ classDef erlang fill:#b83998,color:white;
...
+            pr-tsdb --> pr-http
+        end
+
+        pr-b{{"Browser<br/>Prometheus UI"}}
+        pr-http ---->|"http://localhost:9090/graph"| pr-b
+    
+        subgraph ja[Jaeger]
+            style ja fill:#60d0e4;
+            ja-col[/"Jaeger Collector<br/>listening on<br/>grpc://jaeger:4317/"/]
+            ja-tsdb[(Jaeger TSDB)]
```
reyang, line 144: Not blocking this PR, might worth double checking if "Jaeger TSDB" is the precise term (I guess it might just be a "DB (database)" instead of "TSDB (timeseries database)").

puckpuck:
These updates are fantastic!

puckpuck (approved):

PR #553, opened 2022-11-03
Remove redundant components from collector config

dmitryax:
`metrics/spanmetrics` pipeline is not being used. Removing it along with unused `otlp/spanmetrics` receiver.

### DIFF
--- a/src/otelcollector/otelcol-config.yml
+++ b/src/otelcollector/otelcol-config.yml
@@ -7,10 +7,6 @@ receivers:
           allowed_origins:
             - "http://*"
             - "https://*"
-  otlp/spanmetrics:
-    protocols:
-      grpc:
-        endpoint: "localhost:65535"
 
 exporters:
   otlp:
@@ -20,7 +16,6 @@ exporters:
   logging:
   prometheus:
     endpoint: "otelcol:9464"
-
 
 processors:
   batch:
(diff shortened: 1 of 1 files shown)


puckpuck:
The spanmetrics processor/receiver combo is used in the demo and exports data into Prometheus.  This is a screenshot with some of the metrics it generates

The spanmetrics processor is absolutely being used in the demo.  This is a screenshot of some of the metrics it generates in Grafana.
Screen Shot 2022-11-02 at 8 27 20 PM

This brings up a new issue we should add, which is putting charts with these metrics into the main demo dashboard we ship in Grafana.  I'll go ahead and close the PR, and create a new Issue to add charts with these metrics.

dmitryax:
@puckpuck can you explain why this pipeline is needed for and make sure the dashboard doesn't work without it before closing the PR?

dmitryax:
cc @demo-approvers

dmitryax:
cc @open-telemetry/demo-approvers

puckpuck:
I think I missed this the first time through.  This removes everything but the processor itself, and uses the prometheus metrics pipeline instead of the otlp/spanmetrics one.

puckpuck (approved):
I tried this config in K8s and it works as expected.

austinlparker (approved):
lgtm on local

PR #554, opened 2022-11-03
Fix PHP Traces

austinlparker:
Addresses regression from https://github.com/open-telemetry/opentelemetry-demo/pull/546


puckpuck:
Did #546 break OTLP export over gRPC?

austinlparker:
It removed it, yes, so we could drop the gRPC dependency. The only thing that used it was grpc otlp export and it was adding to build time for the image.

puckpuck (approved):

PR #566, opened 2022-11-08
Remove setting `http.route` attribute in frontend

AbhiPrasad:
## Changes

As per the [OpenTelemetry HTTP Server semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions), the `http.route` attribute is meant to be a low cardinality route template. The URI cannot substitute it.

This patch remove the setting of the `http.route` semantic attribute by the Instrumentation Middleware in the frontend service.

Trying to get a paramaterized template route in NextJS is incredibly difficult (take a look at https://github.com/getsentry/sentry-javascript/issues/5505 to see what it takes), so I did not replace it with anything. This seems to be fine since as per the semantic convention is the `http.route` attribute is "Conditionally Required: If and only if it's available"

Two notes:
1. Should we change `SemanticAttributes.HTTP_TARGET` to be the url? This seems to match the semantic conventions better.
2. Do we need a changelog entry for this?


puckpuck (approved):
This change makes sense

julianocosta89 (approved):

AbhiPrasad:
Do I need to do anything else to get this merged in? Or is there some release cadence we need to follow here?

cartersocha:
> Do I need to do anything else to get this merged in? Or is there some release cadence we need to follow here?

Just update the branch and for future prs please use a personal fork. We can’t trigger branch updates on an organizational fork. 

Should be good to go after that

PR #567, opened 2022-11-09
Render shipping cost nanos in emailservice

jjatria:
## Changes

The template used for the confirmation email in the emailservice did not render the nanos of the shipping cost. This commit adds them, so the rendering of shipping cost and item cost (lower in the template) is equivalent.

This is a minor fix.

### DIFF
--- a/src/emailservice/views/confirmation.erb
+++ b/src/emailservice/views/confirmation.erb
@@ -32,7 +32,7 @@
     <p><%= order.order_id %></p>
     <h3>Shipping</h3>
     <p><%= order.shipping_tracking_id %></p>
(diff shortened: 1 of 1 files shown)


julianocosta89:
Hello @jjatria, thanks for the contribution.
Is this change actually visible somewhere?
I've rebuilt the service locally, but I couldn't find any difference from the `main` branch and this PR.

julianocosta89 (approved):

ahayworth (approved):

ahayworth:
@jjatria Thanks! I think this was just an oversight from when I initially added the ruby bits (but never noticed because as you pointed out: we don't _really_ send any emails 😆 )

PR #569, opened 2022-11-11
[currencyservice] Create multiple build jobs to optimize build time

mviitane:
## Changes

Create multiple build jobs to run the currency service build in parallel and optimize build time. 

Different commands are needed to find the number of available CPU cores per OS.
`make -j$(nproc || sysctl -n hw.ncpu || echo 1)`
- Uses `nproc` on Linux
- Uses `sysctl -n hw.ncpu` on macOS
- Uses 1 core for others

## Background

Currency service had the longest build time among the Demo App services.

Rough build time improvements with this change:

- Linux, docker (Ryzen 5 5600G): 1000 s -> 400 s.
- macOS, docker desktop (M1 Pro): 750 s -> 450 s.

Measured with:
`docker compose build currencyservice --no-cache`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,3 +134,5 @@ significant modifications will be credited to OpenTelemetry Authors.
 ([#512](https://github.com/open-telemetry/opentelemetry-demo/pull/512))
 * Replaced PHP-CLI to PHP-Apache for a more realistic service
 ([#563](https://github.com/open-telemetry/opentelemetry-demo/pull/563))
+* Optimize currencyservice build time with parallel build jobs
+([#569](https://github.com/open-telemetry/opentelemetry-demo/pull/569))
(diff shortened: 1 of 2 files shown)


julianocosta89 (approved):
Awesome!
```
docker compose build currencyservice ||| Building 553.7s (14/14) FINISHED
```

cartersocha:
No real change on my personal machine. But it's a basic macbook air m2 sku

cartersocha:
it did improve build time by ~36 seconds and any improvement is welcome

mviitane:
> No real change on my personal machine. But it's a basic macbook air m2 sku

@cartersocha, it can be that Docker Desktop allocates only 2 CPUs by default (for MacBook Air). In that case, this PR makes no difference. You can check this from Docker Desktop Preferences/Resources and allocate more. You should get faster build times when you give more than 2 CPUs.

cartersocha:
Ahh of course thanks. I overlooked that. 

Should we document a cpu recommendation in the docker or contributor file? Bumping my cpus up to 5 + your updates dropped my build time to ~830 seconds from ~1150. 

Ideally we still make perf improvements to currency, feature flag, and shipping services regardless

PR #584, opened 2022-11-15
Align Docker Postgres image with Kubernetes

mviitane:
## Changes

Previously, we have been using different Postgres images in Docker and Kubernetes deployments (possibly by mistake).

This PR changes the Docker image from cimg/postgres to official postgres. Reasoning:
- Align Docker and Kubernetes deployments
- The official postgres image is many times smaller than the previously used image
- The official postgres provides native support also for ARM


cartersocha (approved):
Lgtm

julianocosta89 (approved):

PR #587, opened 2022-11-15
[protobuf] - encode zipcode as a string

puckpuck:
This changes the `zipCode` attribute of the Address message to be a string type instead of an int.

Also adds an attribute to the shipping service spans that contain the zipCode.

- [x] Appropriate CHANGELOG.md updated for non-trivial changes


cartersocha:
Could you add a change log item too?

julianocosta89 (commented):
The changes look good to me, but when testing I got a weird behavior.
Even though everything was working when the requests were coming from the loadgen, when I tried to navigate to the page and "buy" a product, the shipping was 0,00 and I got the following error in the console:
```
shipping-service         | thread 'tokio-runtime-worker' panicked at 'called `Option::unwrap()` on a `None` value', src/shipping_service.rs:67:91
shipping-service         | 07:52:13 [INFO] GetQuoteRequest: Request { metadata: MetadataMap { headers: {"traceparent": "00-70c1bde2eddef2088aaceeefe54d15eb-460f45857d9e8886-01", "grpc-accept-encoding": "identity,deflate,gzip", "accept-encoding": "identity", "user-agent": "grpc-node-js/1.6.7", "content-type": "application/grpc", "te": "trailers"} }, message: GetQuoteRequest { address: None, items: [CartItem { product_id: "66VCHSJNUP", quantity: 1 }] }, extensions: Extensions }
frontend                 | Error: 1 CANCELLED: Call cancelled
frontend                 |     at Object.callErrorFromStatus (/app/node_modules/@grpc/grpc-js/build/src/call.js:31:26)
frontend                 |     at Object.onReceiveStatus (/app/node_modules/@grpc/grpc-js/build/src/client.js:189:52)
frontend                 |     at Object.onReceiveStatus (/app/node_modules/@grpc/grpc-js/build/src/client-interceptors.js:365:141)
frontend                 |     at Object.onReceiveStatus (/app/node_modules/@grpc/grpc-js/build/src/client-interceptors.js:328:181)
frontend                 |     at /app/node_modules/@grpc/grpc-js/build/src/call-stream.js:187:78
frontend                 |     at processTicksAndRejections (node:internal/process/task_queues:78:11) {
frontend                 |   code: 1,
frontend                 |   details: 'Call cancelled',
frontend                 |   metadata: Metadata { internalRepr: Map(0) {}, options: {} }
frontend                 | }
```

I'm not getting any error if I don't navigate via browser.

puckpuck:
@julianocosta89 can you try again? I made some changes to the frontend code to pass in the dummy address when it attempts to get the shipping quote.

julianocosta89:
> @julianocosta89 can you try again? I made some changes to the frontend code to pass in the dummy address when it attempts to get the shipping quote.

Fixed! Good one!

julianocosta89 (approved):

PR #599, opened 2022-11-19
[featureflag] - Move protobuf file generation to docker build

### LINKED ISSUE #279: [featureflag] - Dockerfile should build protobuf modules
# Feature Request

Update Dockerfile to build artifacts similar to how other services are built. 

- [ ] Use the top-level build context, updating paths in Dockerfile accordingly
- [ ] Copy the top-level demo.proto file into the appropriate build location
- [ ] Run `rebar3 grpc gen` as part of the build

cc @tsloughter

puckpuck:
Fixes #279 

- [x] Use the top-level build context, updating paths in Dockerfile accordingly
- [x] Copy the top-level demo.proto file into the appropriate build location
- [x] Run rebar3 grpc gen as part of the build

## Changes

This moves the protobuf generated code files to be done in the docker build instead of a prior step. This will increase build time slightly (10-20 seconds) to set up the rebar tooling and generate the code files.


austinlparker (changes requested):
If you're changing the build context you need to update the GHA file for build-and-release with the new context

mic-max (approved):

puckpuck:
> If you're changing the build context you need to update the GHA file for build-and-release with the new context

This is done. Also updated the `make generate-protobuf` command to generate files.

austinlparker (approved):

PR #457, opened 2022-10-17
Implement message queue

secustor:
Fixes #429 

## Changes
- [x] add Kafka to docker-compose
- [x] add Kafka producer to `checkoutservice`
- [x] implement Accountingservice
- [x] implement Frauddetectionservice 

For significant contributions please make sure you have completed the following items:

* [x] Appropriate `CHANGELOG.md` updated for non-trivial changes
* [x] Design discussion issue #429

PR #600, opened 2022-11-20
Fix #586: Pass `categories` as an input for Ad service

### LINKED ISSUE #586: Strange adservice behavior. Is it intentional?
# Bug Report

Version: `v1.1.0` / `9a66c3a7e490dc0a30c738b92d931d04ce42f479` / `main` (just plain `git clone`)

## Symptom

It seems that the request for the ad is crafted incorrectly. It results in random ads being served.

After a quick looking at the docs and scenarios it's not clear whether it's intentional or not.

Appropriate adds are served for every request.

Random ad is served for every request.

## Reproduce

1. Open any product page. E.g. "Optical Tube Assembly": http://localhost:8080/product/9SIQT8TOJO. This `9SIQT8TOJO` is a product id, right?
2. A request is seen in browser's developers console: http://localhost:8080/api/data?productIds=9SIQT8TOJO
3. This `9SIQT8TOJO` is used to query `adservice`: https://github.com/open-telemetry/opentelemetry-demo/blob/9a66c3a7e490dc0a30c738b92d931d04ce42f479/src/frontend/pages/api/data.ts#L12
4. A request arrives in `adservice`:
    ```
    ad-service               | 2022-11-15 20:57:22 - hipstershop.AdService - received ad request (context_words=[9, S, I, Q, T, 8, T, O, J, O]) trace_id=e8c29c9773b16a424d5383acb41a2f15 span_id=ecd925196c0ed6c9 trace_flags=01
    ```
    image
5. So this `9SIQT8TOJO` is treated as ten separate context words.
6. The ads map doesn't contain any ads for any of these letters: https://github.com/open-telemetry/opentelemetry-demo/blob/9a66c3a7e490dc0a30c738b92d931d04ce42f479/src/adservice/src/main/java/hipstershop/AdService.java#L221-L227
7. Thus, no ads found and the ads request falls back to the `getRandomAds` any time.
8. This is also seen in Jaeger (all the requests have a call to the `getRandomAds` at the end):
    image

## Additional Context

N/A

madhead:
Fixes #586.

## Changes

1. Pass `categories` as an input for Ad service
2. Randomize the ad shown, in case there are multiple ads in a response.
3. I had to bump the TS target to ES2015 to be able to use `Set`. I hope this is OK for a demo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -142,3 +142,5 @@ significant modifications will be credited to OpenTelemetry Authors.
 ([#583](https://github.com/open-telemetry/opentelemetry-demo/pull/583))
 * Change ZipCode data type from int to string
 ([#587](https://github.com/open-telemetry/opentelemetry-demo/pull/587))
+* Pass product's `categories` as an input for the Ad service
+([#600](https://github.com/open-telemetry/opentelemetry-demo/pull/600))
--- a/src/frontend/components/Ad/Ad.tsx
+++ b/src/frontend/components/Ad/Ad.tsx
@@ -3,9 +3,8 @@ import { useAd } from '../../providers/Ad.provider';
 import * as S from './Ad.styled';
 
 const Ad = () => {
-  const {
-    adList: [{ text, redirectUrl } = { text: '', redirectUrl: '' }],
-  } = useAd();
+  const { adList } = useAd();
+  const { text, redirectUrl } = adList[Math.floor(Math.random() * adList.length)] || { text: '', redirectUrl: '' };
 
   return (
     <S.Ad data-cy={CypressFields.Ad}>
--- a/src/frontend/gateways/Api.gateway.ts
+++ b/src/frontend/gateways/Api.gateway.ts
@@ -78,11 +78,11 @@ const ApiGateway = () => ({
       },
     });
   },
-  listAds(productIds: string[]) {
+  listAds(contextKeys: string[]) {
     return request<Ad[]>({
       url: `${basePath}/data`,
       queryParams: {
-        productIds,
+        contextKeys,
       },
     });
   },
--- a/src/frontend/gateways/rpc/Ad.gateway.ts
+++ b/src/frontend/gateways/rpc/Ad.gateway.ts
@@ -6,9 +6,9 @@ const { AD_SERVICE_ADDR = '' } = process.env;
 const client = new AdServiceClient(AD_SERVICE_ADDR, ChannelCredentials.createInsecure());
 
 const AdGateway = () => ({
-  listAds(productIds: string[]) {
+  listAds(contextKeys: string[]) {
     return new Promise<AdResponse>((resolve, reject) =>
-      client.getAds({ contextKeys: productIds }, (error, response) => (error ? reject(error) : resolve(response)))
+      client.getAds({ contextKeys: contextKeys }, (error, response) => (error ? reject(error) : resolve(response)))
     );
   },
 });
--- a/src/frontend/pages/api/data.ts
+++ b/src/frontend/pages/api/data.ts
@@ -8,8 +8,8 @@ type TResponse = Ad[] | Empty;
 const handler = async ({ method, query }: NextApiRequest, res: NextApiResponse<TResponse>) => {
   switch (method) {
     case 'GET': {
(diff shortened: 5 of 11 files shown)


julianocosta89 (approved):
LGTM

cartersocha:
@madhead please make a changelog entry and I'll merge 🚀

madhead:
@cartersocha, done.

puckpuck:
When I run this... spans going into the Ad service have empty `app.ads.contextKeys`.

Screen Shot 2022-11-21 at 11 27 40 PM

puckpuck (changes requested):
When I run this I get empty `app.ads.contextKeys`

Screen Shot 2022-11-21 at 11 27 40 PM

julianocosta89:
> When I run this I get empty `app.ads.contextKeys`
> 
> Screen Shot 2022-11-21 at 11 27 40 PM

@puckpuck not all of the requests:
image

mviitane:
I also tested this and most of the time `app.ads.contextKeys` is empty. I'd assume this is not how it should be.

madhead:
@mviitane , @julianocosta89, @puckpuck

I agree, I saw those empty requests too, but didn't address them here.

I'm not an React expert, but the bigger problem here, IMO, is that [`AdProvider`](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/frontend/providers/Ad.provider.tsx) provides not only ads, but recommendations too, violating single responsibilities principle. But wherever it is used, both ads and recommendations are requested.

Looking at three usages of this provider:

1. https://github.com/open-telemetry/opentelemetry-demo/blob/90964205fd37a3dbd8d36b9d7362590ced4bc818/src/frontend/pages/cart/index.tsx#L17-L25
    Cart only uses recommendations.
2. https://github.com/open-telemetry/opentelemetry-demo/blob/32d1ee681042e55cb2c360d918a4928de50325c2/src/frontend/pages/product/%5BproductId%5D/index.tsx#L51-L84
    Product page uses both [ads](https://github.com/open-telemetry/opentelemetry-demo/blob/32d1ee681042e55cb2c360d918a4928de50325c2/src/frontend/pages/product/%5BproductId%5D/index.tsx#L81) and [recommendations](https://github.com/open-telemetry/opentelemetry-demo/blob/32d1ee681042e55cb2c360d918a4928de50325c2/src/frontend/pages/product/%5BproductId%5D/index.tsx#L79).
3. https://github.com/open-telemetry/opentelemetry-demo/blob/90964205fd37a3dbd8d36b9d7362590ced4bc818/src/frontend/pages/cart/checkout/%5BorderId%5D/index.tsx#L19-L43
    Order page uses both [ads](https://github.com/open-telemetry/opentelemetry-demo/blob/90964205fd37a3dbd8d36b9d7362590ced4bc818/src/frontend/pages/cart/checkout/%5BorderId%5D/index.tsx#LL40) and [recommendations](https://github.com/open-telemetry/opentelemetry-demo/blob/90964205fd37a3dbd8d36b9d7362590ced4bc818/src/frontend/pages/cart/checkout/%5BorderId%5D/index.tsx#L38).

In some of this contexts the categories are inherently empty.

I would say, this should be fixed and split into two different provider components.

But, I could just "short-circuit" this call, what do you think? Like put a check, a guard clause, "if the categories are empty, then just return empty list of adds". Or something similar to remote this unneeded call. A dirty hack, but should work.

Would it be ok?

madhead:
I totally forgot about the load generator. It might be sending incorrect requests too. I've added both fixes, please check the PR now (beware of the cached images!).

cartersocha:
@austinlparker we should test the build time increase from 3 new containers. One thing to keep in mind is that go doesn’t have an operator instrumentation option. We’d need to implement that with the Java service or replace Go

cartersocha:
For the architecture docs and services table my bias would be to label the language Java not Kotlin for the fraud detection service

julianocosta89:
@cartersocha I think your comments are regarding #457, and not this one😅

cartersocha:
Lol the perils of mobile bulk reviews @julianocosta89. Thanks

julianocosta89 (approved):
After the update we always get a context.

puckpuck (approved):
Confirmed this provides proper context keys on all requests types.

PR #610, opened 2022-11-29
shippingservice: add http client request span

styblope:
# Changes
I propose to add a dedicated http client span for the outgoing calls to `/getquote`. This change will provide tracing separation between the server part and the downstream requests client part inside shippingservice. A client span can also better match with the peer quoteservice server span. The client middleware-based instrumentation uses the available `reqwests-tracing` and `tracing-opentelementry` crates.

## Merge Requirements

essential items:

### DIFF
--- a/docs/services/shippingservice.md
+++ b/docs/services/shippingservice.md
@@ -152,6 +152,21 @@ Adding a span event:
     info!("Tracking ID Created: {}", tid);
 ```
 
+### Reqwest HTTP client instrumentation
+
+A child *client* span is also produced for the outoing HTTP call to
+`quoteservice` via the `reqwest` client. This span pairs up with the
+corresponding `quoteservice` *server* span. The tracing instrumentation is
+implemented in the client middleware making use of the available
+`reqwest-middleware`, `reqwest-tracing` and `tracing-opentelementry` libraries:
+
+```rust
+    let reqwest_client = reqwest::Client::new();
+    let client = ClientBuilder::new(reqwest_client)
+        .with(TracingMiddleware::<SpanBackendWithUrl>::new())
+        .build();
+```
+
 ## Metrics
 
 TBD
--- a/docs/services/shippingservice.md
+++ b/docs/services/shippingservice.md
@@ -57,7 +57,7 @@ Be mindful of async runtime, [context
 guards](https://docs.rs/opentelemetry/latest/opentelemetry/struct.ContextGuard.html),
 and inability to move and clone `spans` when replicating from these samples.
 
-### Adding gRPC/HTTP instrumentation
+### Adding gRPC instrumentation
 
 This service receives gRPC requests, which are instrumented in the middleware.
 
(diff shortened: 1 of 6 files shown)


julianocosta89:
Hello @styblope could you please sign the CLA?

julianocosta89 (changes requested):
Thanks for the contribution @styblope!
A lot of new attributes and events were added automatically. Good one!

Could you also update the doc for the Shipping Service to reflect your changes?
https://github.com/open-telemetry/opentelemetry-demo/blob/main/docs/services/shippingservice.md

docs/services/shippingservice.md
```diff
@@ -152,6 +152,21 @@ Adding a span event:
     info!("Tracking ID Created: {}", tid);
 ```
 
+### Reqwest HTTP client instrumentation
```
julianocosta89, line 155: Just a nit pick 😅 

Could you move that to the following section?
https://github.com/styblope/opentelemetry-demo/blob/shippingservice-client-span/docs/services/shippingservice.md#adding-grpchttp-instrumentation

Maybe we could even split the section in 2:
`Adding gRPC instrumentation` with what is there.

and: `Adding HTTP instrumentation` with what you have added.

styblope: yup makes sense, 1 sec

julianocosta89:
Hey @GaryPWhite, as you were the main responsible for the `shippingservice` implementation, would you mind taking a look at this one?

src/shippingservice/src/shipping_service/quote.rs
```diff
@@ -48,7 +50,10 @@ async fn request_quote(count: u32) -> Result<f64, Box<dyn std::error::Error>> {
     let mut reqbody = HashMap::new();
     reqbody.insert("numberOfItems", count);
 
-    let client = reqwest::Client::new();
+    let reqwest_client = reqwest::Client::new();
+    let client = ClientBuilder::new(reqwest_client)
```
GaryPWhite, line 54: personal style choice, I would rewrite this as just:

```rust
let client = ClientBuilder::new(reqwest::Client::new())
```

Up to you.

src/shippingservice/src/main.rs
```diff
@@ -82,7 +91,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         .await;
 
     init_logger()?;
-    init_tracer()?;
+    let tracer = init_tracer()?;
+    init_reqwest_tracing(tracer)?;
```
GaryPWhite, line 95: same as other comment -- I think this one makes a little less sense with the `?` but still think this reads better as 

```
init_reqest_tracing(init_tracer()?)?;
```

my 2c.

GaryPWhite (approved):
this is really cool. I want to point out that this is a way of using even _downstream_ dependencies in a really adept way and that's a massive contribution of SME to this project. Thanks much for your input here.

left some comments but by no means blocking.

src/shippingservice/src/shipping_service/quote.rs
styblope: Thanks for the kind words. I guess I wanted to be explicit and transparent with the new dependency, but I also agree with the concise Rusty syntax. Will rewrite the bits as you propose.

PR #613, opened 2022-11-30
[FrontendProxy] Add tracing to Envoy

julianocosta89:
# Changes

As Envoy supports OpenTelemetry, I believe it would be great if we could use it.
https://www.envoyproxy.io/docs/envoy/latest/start/sandboxes/opentelemetry

This PR adds the configuration required to add tracing to Envoy.

jaeger-proxy

## Merge Requirements

essential items:


cartersocha:
We should add envoy to the current services architecture doc. Especially if there's a span coming from it

Should we add a simple envoy service doc with config instructions too?

julianocosta89:
@cartersocha Not sure what exactly to write, but I've added the "how to" make Envoy produce spans.

cartersocha:
I'm seeing errors on the frontend and checkout service

cartersocha:
Not sure why the shipping service is throwing issues. 

"13 INTERNAL: shipping quote failure: failed to get shipping quote: rpc error: code = Unknown desc = error sending request for url (http://quoteservice:8090/getquote): error trying to connect: dns error: failed to lookup address information: Name does not resolve"
--

cartersocha:
A system prune cleared up the errors but I'm not seeing an envoy span in any trace. I see envoy as an orphan span currently and the main trace doesn't have it present

julianocosta89:
@cartersocha true!
Something that I forgot to mention is that this is generating spans only when you navigate to the OTel-Demo page.
The loadgen doesn't send requests through the envoy, but to the frontend container directly.

cartersocha:
I see! lgtm. 

This is where it would be nice to have non synthetic requests set the attribute to filter on.

cartersocha (approved):
Please add a note in the envoy services doc at the top saying only non synthetic requests will trigger the envoy tracing. 

Lgtm otherwise

docker-compose.yml
```diff
@@ -262,6 +262,8 @@ services:
       - GRAFANA_SERVICE_HOST
       - JAEGER_SERVICE_PORT
       - JAEGER_SERVICE_HOST
+      - OTEL_COLLECTOR_HOST
```
puckpuck, line 265: I think these environment variables belong in the `# Collector` section instead.  Though only used by Envoy, we should keep these vars with the other similar ones.

puckpuck (approved):
One small nit. everything works as expected. 👍

docker-compose.yml
julianocosta89: @puckpuck I think you meant in the `.env` file, right?

puckpuck: yes I meant in the .env file

PR #614, opened 2022-12-05
Add Sentry to Python/recommendationservice

antonpirker:
Adds Sentry intstrumentation to the Python recommendationservice.

For local development I added a `docker-compose.override.yml` to set the Sentry DSN and use the local Sentyr Python SDK for better debugging. The override file is not checked into the repo to not break stuff.


antonpirker:
Wrong repo. Sorry for the noise!

PR #622, opened 2022-12-09
Reduce rust trace verbosity.

austinlparker:
# Changes

In 1.2, the new tracing crates for reqwest used the default subscriber, which picks up everything from `trace` level on up. Thanks to the tireless efforts of the rust community, this means we get spans and span events for quite literally everything! While great for detailed inspection of performance, not so great for our purposes. This PR reduces the verbosity at build time to something reasonable (`info`), giving us our http client spans without murdering our poor jaeger instances.


julianocosta89 (approved):
This excludes all events from the `reqwest` instrumentation, but it LGTM.

puckpuck (approved):

PR #626, opened 2022-12-11
[frontendproxy] - remove set-cookie

puckpuck:
In October we added code to the frontend proxy to set a cookie used by the frontend in order to specify the browser reachable otel-collector.  This cookie is no longer used by the frontend, and we can safely remove it.

_red diffs are the best diffs_


austinlparker (approved):

austinlparker:
I think there's more cleanup here? Do we need to remove the env var set in the compose file as well?

Also, just to make sure, the thing that's in-place now works at runtime not compile-time right?

fatsheep9146 (approved):

puckpuck:
@austinlparker I removed the env var from docker-compose.yml.  Is anything else missing?

austinlparker:
lgtm

PR #640, opened 2022-12-15
[chore] fix generate protobuf for TypeScript

puckpuck:
Previously `make generate-protobuf` only copied the demo.proto file but did not generate the protobuf classes required for IDEs to resolve imports. This adds the command to generate the protobuf classes.

### DIFF
--- a/ide-gen-proto.sh
+++ b/ide-gen-proto.sh
@@ -57,6 +57,8 @@ gen_proto_ts() {
   echo "Generating Typescript protobuf files for $1"
   cd "$base_dir"/src/"$1" || return
   cp -r "$base_dir"/pb .
+  mkdir -p ./protos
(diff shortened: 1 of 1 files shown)


fatsheep9146 (approved):

julianocosta89:
@puckpuck do we need an extra tool for this?
I keep getting:
```
Generating Typescript protobuf files for frontend
./node_modules/.bin/protoc-gen-ts_proto: program not found or is not executable
Please specify a program using absolute path or make sure the program is available in your PATH system variable
--ts_proto_out: protoc-gen-ts_proto: Plugin failed with status code 1.
```

I've tried installing https://www.npmjs.com/package/protoc-gen-ts, but it didn't solve.

puckpuck:
@julianocosta89 you need to run `npm install` in the src/frontend folder first.

I can add that to the gen proto script if you feel we should have it there as well.

julianocosta89:
> @julianocosta89 you need to run `npm install` in the src/frontend folder first.
> 
> I can add that to the gen proto script if you feel we should have it there as well.
> 

Either that or we could mention that step in the development doc

puckpuck:
Let’s go with doc updates. I’ll get a PR together shortly.

PR #641, opened 2022-12-15
[frontend] add address to IProductCheckout type

puckpuck:
This change benefits IDEs so they can properly detect and find type declarations. It adds `shippingAddress` to the `IProductType` type so that code in the cart checkout order page can properly detect the declaration.


julianocosta89 (approved):
LGTM

PR #645, opened 2022-12-19
Update sentry sdk for python

antonpirker:
Using the productino Sentry SDK for Python, because it now supports OTel.


antonpirker:
wrong.

PR #654, opened 2022-12-21
Add OTel java agent with JMX Metric Insights to kafka

svrnm:
# Changes

This PR adds the java agent to the kafka broker. While this will not provide tracing, it will use [JMX Metric Insights](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/jmx-metrics/javaagent) generate metrics and export them to prometheus:

Screenshot 2022-12-21 at 15 27 02

Screenshot 2022-12-21 at 19 03 26

## Additional info

We plan to have a blog post about JMX Metric Insights on opentelemetry.io eventually, so calling out that the demo is making use of it would be a nice feature:-)

## Merge Requirements


src/kafka/Dockerfile
```diff
@@ -1,5 +1,10 @@
 FROM confluentinc/cp-kafka:latest-ubi8
 
+USER root
```
fatsheep9146, line 3: Should we allow using the `root` user? I remember that we modify all service user to non-root user in previous pr.

@open-telemetry/demo-approvers @open-telemetry/demo-maintainers

svrnm: The root permissions are needed for the chmod of the agent. They are dropped again later down in the Dockerfile.

mviitane (approved):
Runs fine. Nice to get the additional Kafka metrics. Kafka seems to consume a massive amount of memory, but I guess there's no easy solution for that.

puckpuck (approved):
This is fantastic. TYVM!

PR #657, opened 2022-12-23
Add build for arm64 arch

cedricziel:
# Changes

This change introduces another architecture for the images created: arm64

## Merge Requirements

essential items:


puckpuck:
@austinlparker is this still needed? I thought we do multi-arch builds already.

austinlparker:
We do not do multi arch builds because it takes like 4+ hours.

cedricziel:
I have added it in particular because the featureflagservice keeps segfaulting on my M1 device with and without emulation.

I think it makes a lot of sense to have this available for multiple architectures. If the build time is a problem, it's probably a question of optimization

TylerHelmuth:
I also thought we fixed this but I just tried installing the helm chart locally again and featureflag service is crashing again (https://github.com/open-telemetry/opentelemetry-helm-charts/issues/580).  

Is there a way that we can do the existing build and publish process and separately build a specific arm64 image either after the existing build finishes or in parallel?

cedricziel:
> Is there a way that we can do the existing build and publish process and separately build a specific arm64 image either after the existing build finishes or in parallel?

We sure could. Not sure if this benefits us. Currently all images are built in parallel and in each build sequentially builds and pushed amd64 and arm64. I would argue there is a fair amount of parallelism already in place.

Current duration for the `Build & Push` job is `[19m 47s]` - so overall pretty far away from the projected "4hrs" :)

austinlparker:
No, it takes about 4 hours: https://github.com/open-telemetry/opentelemetry-demo/actions/runs/3313405848

It might be better now with some of the optimizations we've made to build time but Rust and C++ images take _forever_ to build under emulation. 

I would suggest modifying the GHA runner file here to allow builds for PR's (you need to modify the `release.yaml` file to allow the `build_and_push_images` action to run on PR _and_ add logic to disable pushing images on non-release builds) in order to see how long it takes.

austinlparker:
I went ahead and tested this out on our runner config -- everything builds OK. Shipping service takes an hour which isn't great, but we can work on getting that down.

austinlparker (approved):

PR #664, opened 2022-12-28
Add OTEL_RESOURCE_ATTRIBUTES to docker compose

svrnm:
# Changes

This adds OTEL_RESOURCE_ATTRIBUTES to the `.env` file and also makes use of it across all services. This allows to easily introduce `service.namespace`


docker-compose.yml
```diff
@@ -290,7 +296,7 @@ services:
       - RECOMMENDATION_SERVICE_ADDR
       - SHIPPING_SERVICE_ADDR
       - OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
-      - OTEL_RESOURCE_ATTRIBUTES=service.name=frontend
+      - OTEL_RESOURCE_ATTRIBUTES=${OTEL_RESOURCE_ATTRIBUTES},service.name=frontend
```
julianocosta89, line 299:
```suggestion
      - OTEL_RESOURCE_ATTRIBUTES=${OTEL_RESOURCE_ATTRIBUTES}
```

I think `service-name` was a leftover here.
We have `OTEL_SERVICE_NAME=frontend` defined below, so we are good to remove it from the `OTEL_RESOURCE_ATTRIBUTES`.

julianocosta89 (changes requested):

puckpuck (approved):

fatsheep9146 (approved):

julianocosta89 (approved):

puckpuck:
Do we want to add this to the Helm chat as well? Is this different from what we call `k8s.namespace.name` we already export as a resource attribute when deployed with Helm?

PR #666, opened 2022-12-29
Add Coralogix as vendor

oded-dd:
# Changes

Adding Coralogix as vendor.

As it is only updating the README.md file with new vendor contribution, I believe no need for other changes before merge (unless stated otherwise).


cartersocha:
Lgtm. 

I'll probably open a small PR to move the companies to a table instead of list now that it's getting long

cartersocha (approved):

fatsheep9146 (approved):

PR #667, opened 2022-12-30
update vendor list to table to reduce space

cartersocha:
Small changes to reduce the increasing amount of space the vendor list is taking. 

Any thoughts on leaving opening slots for vendors to easily add their links to or a different table config?

In my experience markdown tables are always a bit of a hassle but I just used a generator to save time


mviitane (commented):
I think you could also use 3 columns to have even more space for additional demos.

cartersocha:
> I think you could also use 3 columns to have even more space for additional demos.

Good thought! It adds a blank space at the end too. Updated.

fatsheep9146 (approved):

mviitane (approved):
One more though. I'd remove the "Vendors" header line completely from the table. I think the scope is described well enough in the description above. Just a though.

PR #615, opened 2022-12-06
support non-root containers

styblope:
## Ensure/enable containers to run as non-root

Many of the demo containers currently run as root by default. We should support running the containers under non-root UIDs to align with security best-practices and support running the demo app in restrictive environments (e.g. OpenShift).

There are two steps to reach the goal:

1. **Enable** containers to run as non-root user.
Make sure the containers can execute as non-root. Some of the containers that now launch under root are happy to run also as non-root. There are, however, a couple of containers (emailservice, adservice, frontend-proxy, postgres) which fail to run under a non-root user and need to be fixed. 

2. **Ensure** the containers run as non-root by default.
This step actually applies non-root UID to all container deployments. This can either be done at launch-time by setting the user/UID in Compose/Helm, or it can be left up to the orchestrator (e.g. [OpenShift UIDs](https://cloud.redhat.com/blog/a-guide-to-openshift-and-uids)). Some containers require to be run under a specific UID defined in `/etc/passwd` (e.g. envoy, postgres, apache).

The PR implements the first step by fixing the services that fail to run under non-root.

* adservice - add read permission to `opentelemetry-javaagent.jar`. Can run as any non-root UID.
* emailservice - read/write permission to `Gemfile.lock` + Dockerfile cleanup. Can run as any non-root UID.
* frontendproxy - build image with built-in `envoy` user.   
* quoteservice -  configure and run apache under built-in `www-data` user + Dockerfile [Apache envvar substitution](https://httpd.apache.org/docs/2.2/configuring.html#syntax) and cleanup.
* ffspostgres - set `postgres` user in Compose, no image modification.
* redis - set `redis` user in Compose, no image modification.

Services with with their default users (after the PR changes) and working non-root alternatives:

| container_name          | default UID          | non-root UID |
|-------------------------|----------------------|--------------|
| ad-service              | root                 | anyuid       |
| cart-service            | root                 | anyuid       |
| checkout-service        | root                 | anyuid       |
| currency-service        | root                 | anyuid       |
| email-service           | root                 | anyuid       |
| feature-flag-service    | nobody (65534)       | anyuid       |
| frontend                | nextjs (1001)        | anyuid       |
| frontend-proxy          | ~~root~~ **envoy** (101) | envoy        |
| load-generator          | root                 | anyuid       |
| payment-service         | node (1000)          | anyuid       |
| product-catalog-service | root                 | anyuid       |
| quote-service           | ~~root~~ **www-data**    | www-data     |
[...]

PR #675, opened 2023-01-09
No root users

svrnm:
# Changes

This fixes all services which are currently still running as root. The solution might not be perfect, that's why I start this of with a draft PR. A few alternatives:

* Instead of creating a new user/new group an existing user like "nobody" could be used
* Giving full permissions to the WORKDIR might not be necessary for all services, but testing that out requires a little bit more effort, since I don't know all the details of all services.

let me know what you think

## Merge Requirements

## Note

Here's a script I was using to find the services that run as root:

```
for IMAGE in $(docker images ghcr.io/open-telemetry/demo --format '{{.Repository}}:{{.Tag}}'); do echo ${IMAGE}; docker run --entrypoint whoami ${IMAGE}; done;
```


styblope:
Ref. #615 where the idea was that we don't enforce non-root unless where necessary. We rather just *enable* the rest of containers so that they can execute either as root or non-root. The actual security context can then be set in via helm parameters, or it can be left up to the orchestrator's security policy enforcement (e.g. OpenShift likes to apply own non-root UIDs from a preset range)

svrnm:
> Ref. #615 where the idea was that we don't enforce non-root unless where necessary. We rather just _enable_ the rest of containers so that they can execute either as root or non-root. The actual security context can then be set in via helm parameters, or it can be left up to the orchestrator's security policy enforcement (e.g. OpenShift likes to apply own non-root UIDs from a preset range)

missed that one, thanks for clarification :)

Issue #679, opened 2023-01-09
Rust and C++ - Can we remove gRPC dependency on exporter?


austinlparker:
In Rust and C++, we have hard requirements on gRPC due to OTLP exporters even if we aren't using gRPC export. Can this be changed upstream?

austinlparker:
https://github.com/open-telemetry/opentelemetry-cpp/issues/1826

This might fix the C++ issue

svrnm:
For c++ the exporter is not the problem alone, as far as I see it: the service itself serves via gRPC, so also that needs to be refactored.

styblope:
Same as above for Rust + [http exporter](https://github.com/open-telemetry/opentelemetry-rust/blob/c3b1056e0e947f1d0e44109dd401cdf3132bacd4/opentelemetry-otlp/src/lib.rs#L251) currently only supports traces, no metrics (in case we'd like to implement some)

cartersocha:
Probably a moot point with build times reduced. Closing for now.

Issue #574, opened 2022-11-14
Switch currency service from grpc to http


cartersocha:
To improve build times we would like to refactor from using grpc in the currency service to http

cartersocha:
fixed for now

PR #687, opened 2023-01-14
Rebuild currency service Dockerfile with alpine

### LINKED ISSUE #681: Optimize c++ currency service build time
# Feature Request

The currency service build time is drastically slowing our release process and a general pain. 

Please help.

svrnm:
# Changes

This should fix #681, #679 partially and #574: Alpine linux provides a [grpc-dev package](https://pkgs.alpinelinux.org/package/edge/community/x86/grpc-dev) that can be used to skip all the dependencies for building opentelemetry-cpp

This brings down the build time to **2m 30s** on an intel macbook pro (2015)

## Merge Requirements


cartersocha:
This made a HUGE difference on my machine. 43 vs.  407 seconds on my m2 mac book air. Thank you so much @svrnm 

@austinlparker we should release a new version immediately after we merge this and test the new build performance. Then maybe we can add another quick release for arm64 support unless other services continue to block us. 

I'd be curious if we could do something similar for the other slow services? fyi @GaryPWhite

cartersocha (approved):

austinlparker:
Wouldn't this also work for the rust image since its the same libs? Worth a shot.

src/currencyservice/src/server.cpp
```diff
@@ -176,10 +176,10 @@ class CurrencyService final : public hipstershop::CurrencyService::Service
     std::string span_name = "CurrencyService/Convert";
     auto span =
         get_tracer("currencyservice")->StartSpan(span_name,
```
julianocosta89, line 178: Could you also update this in here?
https://github.com/open-telemetry/opentelemetry-demo/blob/main/docs/services/currencyservice.md#create-new-spans

julianocosta89 (changes requested):

julianocosta89 (approved):

lalitb (approved):
Nicely done. Thanks.

src/currencyservice/src/server.cpp
svrnm: ah yes, thx, done!

svrnm:
Additional update: To reduce the final image size by ~300 MB (down to 65 on arm64) I turned it into a multi-stage build.

fatsheep9146 (approved):

mviitane:
Thanks @svrnm, this was an amazing improvement!

PR #694, opened 2023-01-17
Add random errors in AdService

jack-berg:
Generate random errors that exercise error handling login in the AdService.  Currently, the error handling logic which adds an event and sets the span status isn't reachable, as far as I can tell.


mviitane (commented):
What would be the reason for creating this error in the normal flow? Should this logic be behind a feature flag (disabled by default)? I’d expect the demo app to run without errors by default unless specifically breaking it.

julianocosta89:
I agree with @mviitane, if we want to add errors to `AdService` we should add it behind a FeatureFlag as we have for other services.

Don't get me wrong @jack-berg I think the PR is valid and it adds a scenario to the demo.
We just need to make it possible to enable/disable the errors from the `FeatureFlagService`.

You can either add this to the current PR, or if you prefer, we can close this one and you can send a new PR with the implementation.

Here is an example of how that is handled in the `ProductCatalogService`: https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/productcatalogservice/main.go#L207

julianocosta89:
> Adding this as a feature flag sounds reasonable to me. I tried to make progress on this today but was [having some trouble](https://github.com/open-telemetry/opentelemetry-demo/issues/696) running the featureflagservice locally, so I'm blocked for the moment.

@jack-berg the featureflagservice was fixed yesterday 🥳 
You should be able to build it locally now.

src/featureflagservice/priv/repo/migrations/20220524172636_create_featureflags.exs
```diff
@@ -25,6 +25,11 @@ defmodule Featureflagservice.Repo.Migrations.CreateFeatureflags do
       name: "recommendationCache",
       description: "Cache recommendations",
       enabled: false})
+
+    repo().insert(%Featureflagservice.FeatureFlags.FeatureFlag{
+      name: "adServiceFailure",
+      description: "Fail ad service requests sporadically",
+      enabled: false})
```
jack-berg, line 32: Copy and pasted. Do I need to do anything else to ensure that migration runs and applies this change?

docker-compose.yml
```diff
@@ -74,6 +74,7 @@ services:
       - "${AD_SERVICE_PORT}"
     environment:
       - AD_SERVICE_PORT
+      - FEATURE_FLAG_GRPC_SERVICE_ADDR
```
jack-berg, line 77: Need to make a corresponding update to the [helm charts](https://github.com/open-telemetry/opentelemetry-helm-charts).

src/featureflagservice/priv/repo/migrations/20220524172636_create_featureflags.exs
julianocosta89: Great question,  I think @tsloughter would be able to help us out.
If I'm not mistaken @puckpuck also added a FeatureFlag previously.

mviitane: @jack-berg, see the last 5 lines of this file. I think the new flag should be added also there.

src/adservice/src/main/java/hipstershop/AdService.java
```diff
@@ -156,7 +182,15 @@ public void getAds(AdRequest req, StreamObserver<AdResponse> responseObserver) {
...
         span.setAttribute("app.ads.ad_response_type", adResponseType.name());
 
-        adRequestsCounter.add(1, Attributes.of(adRequestTypeKey, adRequestType.name(), adResponseTypeKey, adResponseType.name()));
+        adRequestsCounter.add(
+            1,
+            Attributes.of(
+                adRequestTypeKey, adRequestType.name(), adResponseTypeKey, adResponseType.name()));
+
+        if (checkAdFailure()) {
+          logger.warn(ADSERVICE_FAIL_FEATURE_FALG + " fail feature flag enabled");
```
mviitane, line 191: FALG->FLAG

src/adservice/src/main/java/hipstershop/AdService.java
```diff
@@ -105,15 +121,25 @@ private void stop() {
...
 
   private enum AdResponseType {
-    TARGETED, RANDOM
+    TARGETED,
+    RANDOM
   }
 
   private static class AdServiceImpl extends hipstershop.AdServiceGrpc.AdServiceImplBase {
 
+    private static final String ADSERVICE_FAIL_FEATURE_FALG = "adServiceFailure";
```
mviitane, line 135: FALG->FLAG

src/adservice/src/main/java/hipstershop/AdService.java
```diff
@@ -169,6 +203,20 @@ public void getAds(AdRequest req, StreamObserver<AdResponse> responseObserver) {
...
+    boolean checkAdFailure() {
+      // Flip a coin and fail 1/10th of the time if feature flag is enabled
+      if (random.nextInt(10) != 1) {
+        return false;
+      }
+
+      GetFlagResponse response =
+          featureFlagServiceStub.getFlag(
+              hipstershop.Demo.GetFlagRequest.newBuilder()
+                  .setName(ADSERVICE_FAIL_FEATURE_FALG)
```
mviitane, line 216: FALG->FLAG

mviitane (commented):
Running fine on my docker environment. Added few minor comments.

src/adservice/settings.gradle
```diff
@@ -1 +1 @@
-rootProject.name = 'hipstershop'
+rootProject.name = 'opentelemetry-demo-ad-service'
```
jack-berg, line 1: Giving the adservice module a more appropriate name. This property is used by IDEs as the name of the project, so this makes it easier to find in a list. `Dockerfile` needed a small update as well.

src/featureflagservice/priv/repo/migrations/20220524172636_create_featureflags.exs
jack-berg: Thanks!

mviitane (approved):
Changelog conflict needs to be resolved. Otherwise, LGTM.

austinlparker (approved):

julianocosta89 (approved):
LGTM

PR #700, opened 2023-01-19
[recommendationservice] - base image, licensing and diagram update

julianocosta89:
# Changes

- Updated base image of recommendationservice to use a smaller one.
  - Image size reduced from 991MB to 190MB.
- As we have changed the files enough, I've updated the licensing to OTel authors.
- Updated demo diagram, as the call from recommendationservice to featureflagservice was missing.

## Merge Requirements

For new features contributions please make sure you have completed the following
essential items:

* [x] `CHANGELOG.md` updated to document new feature additions
* [x] Appropriate documentation updates in the [docs](../docs/) folder

Maintainers will not merge until the above have been completed. If you're unsure
which docs need to be changed ping the
[@open-telemetry/demo-approvers](https://github.com/orgs/open-telemetry/teams/demo-approvers).

PR #703, opened 2023-01-20
Update emailservice Dockerfile to use alpine and multistage build

svrnm:
# Changes

Once again some updates to one of the Dockerfiles, this time for emailservice (ruby):

* Do a multistage build
* Use alpine

This brings down the image size to 90MB (from 447MB) on arm64 and the build time improves (from 20 down to 12s)


cartersocha:
447 mb and 20s build time to 90 mb and 13s. Nice!

cartersocha (approved):

src/emailservice/.dockerignore
```diff
@@ -0,0 +1,2 @@
+Dockerfile
```
julianocosta89, line 1: Maybe we do not need the `.dockerignore` here.
I've added the `Dockerfile`s and the `README`s in the root `.dockerignore`, as all the docker compose build commands should be executed from there.
The PR where I do this should be good to go and merged soon:
https://github.com/open-telemetry/opentelemetry-demo/pull/700/files#diff-2f754321d62f08ba8392b9b168b83e24ea2852bb5d815d63e767f6c3d23c6ac5

This is not a blocker though.

julianocosta89 (approved):

src/emailservice/.dockerignore
svrnm: Yeah, I saw that one, I think it's a good practice in general, because having the dockerfile added to the image will make all caches invalid, doing a full build everytime. 👍

PR #705, opened 2023-01-20
Update dockerfile for adservice to use different base images

svrnm:
# Changes

Minor change for the adservice: Similar to frauddetectionservice this makes use of the gradle image to save some time downloading gradle (since both use the same gradle image then this even saves some extra time downloading the image). It also replaces the jdk with a jre image for the final image, saves some ~190MB (518 down to 328)

### DIFF
--- a/src/adservice/Dockerfile
+++ b/src/adservice/Dockerfile
@@ -12,14 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM gradle:7-jdk17 AS builder
+FROM eclipse-temurin:17-jdk AS builder
 
 WORKDIR /usr/src/app/
 
+COPY ./src/adservice/gradlew* ./src/adservice/settings.gradle* ./src/adservice/build.gradle .
+COPY ./src/adservice/gradle ./gradle
+
+RUN ./gradlew
+RUN ./gradlew downloadRepos
+
 COPY ./src/adservice/ ./
 COPY ./pb/ ./proto
-RUN gradle downloadRepos
-RUN gradle installDist -PprotoSourceDir=./proto
+RUN ./gradlew installDist -PprotoSourceDir=./proto
 
 # -----------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


cartersocha:
10s build time improvement and same size reduction. lgtm

cartersocha (approved):
🚀

src/adservice/Dockerfile
```diff
@@ -12,26 +12,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM eclipse-temurin:17-jdk AS builder
+FROM gradle:7-jdk17 AS builder
```
jack-berg, line 15: I don't recommend using a the gradle wrapper for local builds / development and using a different version of gradle for the docker build.

Instead, you can speed up the docker build by ordering the commands such that things that change infrequently like the gradle wrapper and the dependencies are cached in intermediate layers:
```
FROM eclipse-temurin:17-jdk AS builder

WORKDIR /usr/src/app/

COPY ./src/adservice/gradlew* ./src/adservice/settings.gradle* ./src/adservice/build.gradle .
COPY ./src/adservice/gradle ./gradle

RUN ./gradlew
RUN ./gradlew downloadRepos

COPY ./src/adservice/ ./
COPY ./pb/ ./proto
RUN ./gradlew installDist -PprotoSourceDir=./proto

# -----------------------------------------------------------------------------

FROM eclipse-temurin:17-jre
```

svrnm: Makes sense, let me fix this here (and with the frauddetectionservice, where I did the same)

austinlparker (approved):

src/adservice/Dockerfile
svrnm: Updated it accordingly

svrnm:
followed @jack-berg's comment and changed the builder once again

cartersocha:
Thanks @svrnm and @jack-berg 🔥

Issue #709, opened 2023-01-22
Improve arm64 build times through cross-compilation


svrnm:
# Feature Request

**Is your feature request related to a problem?**

As stated in [this comment](https://github.com/open-telemetry/opentelemetry-demo/pull/657#issuecomment-1375798475) by @austinlparker the build for arm64 takes ~4hrs right now. This time hopefully should improve due to the plenty improvements that now exist for the Dockerfiles, but will probably still be significant (e.g. the rust image took 1+hrs and is not at all optimized)

**Describe the solution you'd like:**

I suggest making use of cross compilation. I did a tiny experiment with the checkoutservice and it improved the build times from ~7mins to ~2mins:

https://github.com/svrnm/playground/actions/runs/3981194659
https://github.com/svrnm/playground/actions/runs/3981102723

The dockerfiles need to be changed for that as outlined in this [blog post](https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/), e.g. here's my checkout service Dockerfile for cross compilation:

https://github.com/svrnm/playground/blob/main/src/checkoutservice/Dockerfile

The good thing is, that those changes don't need to be introduced at once and also not for all services. My suggestion is that we pick the ones which take the longest:

- [ ] Currency 
- [ ] Payment
- [ ] Quote
- [ ] Shipping

What do you all think?

cartersocha:
FYI @open-telemetry/demo-approvers 

super interesting and worth pursuing I think

austinlparker:
I'm not entirely sure this is going to help with the worst-case stuff. It's a nice reduction in time, but since builds happen monolithically (every image is built on release, we don't do builds for specific changed containers during PRs, etc) saving time on go doesn't really help unless we get the really slow stuff broken up and we can't cross-compile the dependency chains for grpc afaik?

svrnm:
@austinlparker this also works for C++/Rust/..., I just picked Go as an example because it was kind of easy. I played with the rust stuff a little bit, but have no results yet. In theory you should be able to cross-compile everything (including the grpc dependency chain), so the ideal setup is to have a build stage that does 99% of the heavy lifting and a release stage that just copies files out of that

austinlparker:
Hm, when I tried doing the cross-compilation stuff I wasn't able to get it to work on Rust (and the C++ side was really frustrating), maybe you'll have better luck.

svrnm:
[...]

PR #715, opened 2023-01-24
Add cross-compilation for shipping service

svrnm:
# Changes

Related to #709, this enables cross-compilation (arm64 target on amd64 build) for the github build workflow. It brings down the build time for the shipping service on both platforms to ~6min (from 1h14min). There are 2 disadvantages two that:

* The Dockerfile becomes complicated (there is probably a way to optimize that, but look at that mess...)
* The final image is not based on alpine anymore but debian slim, because cross-compiling with musl (and without glibc) didn't work for a variety of reasons, so the final image is bigger (97.8MB)

I raise this as a draft PR because this only becomes valuable when other long running builds are reduced accordingly, so merging it is not yet reasonable. The purpose of the PR right now is to discuss if the disadvantages (especially the first one) are worth the effort (assuming something similar can be done for the Quote, Payment & Currency-Services)


svrnm:
taking this out from Draft state for now. As a note: there is an alternative ways on how to structure this: Conditional stages using build arguments.

By introducing a build argument (e.g. BUILDER_TYPE) one can have 2 different stages for native & cross compilation, looks like this:

```
ARG BUILDER_TYPE=native
FROM rust:1.61 as builder-native
# native compilation
...
FROM --platform=${BUILDPLATFORM} rust:1.61 as builder-aarch64_on_x86
# cross compile
...
FROM builder-${BUILDER_TYPE} as builder
...
FROM debian:bullseye-slim as release
COPY --from=builder /app/target/release/shippingservice /app/shippingservice
...
```

The advantage is a clearer set of docker instructions per stage
The build argument needs to be set in the github workflow

Which one do people prefer?

cartersocha:
131s vs. 105s for me. 

When you do a full demo build the docker log goes a bit nuts but I think that's intended with the cross compilation

austinlparker:
> taking this out from Draft state for now. As a note: there is an alternative ways on how to structure this: Conditional stages using build arguments.
> 
> By introducing a build argument (e.g. BUILDER_TYPE) one can have 2 different stages for native & cross compilation, looks like this:
> 
> ```
> ARG BUILDER_TYPE=native
> FROM rust:1.61 as builder-native
> # native compilation
> ...
> FROM --platform=${BUILDPLATFORM} rust:1.61 as builder-aarch64_on_x86
> # cross compile
> ...
> FROM builder-${BUILDER_TYPE} as builder
> ...
> FROM debian:bullseye-slim as release
> COPY --from=builder /app/target/release/shippingservice /app/shippingservice
> ...
> ```
> 
> The advantage is a clearer set of docker instructions per stage The build argument needs to be set in the github workflow
> 
> Which one do people prefer?

Honestly, I don't feel like the Dockerfile is that confusing as-is; I'd lean towards accepting it as stands in order to avoid adding a special case for the build scripts.

julianocosta89 (approved):
In my case it went from 990s to 255s (WSL - Ubuntu 20.04)
Great improvement!

PR #726, opened 2023-01-31
[chore] - fix image version

puckpuck:
Remove the `v` prefix from the image version. We now publish images without this prefix.

Without this `docker compose build` will not work.

### DIFF
--- a/.env
+++ b/.env
@@ -1,5 +1,5 @@
 # Images
-IMAGE_VERSION=v1.3.0
+IMAGE_VERSION=1.3.0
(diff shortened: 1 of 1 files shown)


julianocosta89 (approved):

fatsheep9146 (approved):

austinlparker:
Don't we need 1.3.1 then since this went out in 1.3?

puckpuck:
> Don't we need 1.3.1 then since this went out in 1.3?

No. This only affects people that want to build from source using `docker compose build`

Issue #758, opened 2023-02-24
Frontend span attribute `app.session.id` gone


julianocosta89:
# Bug Report

Which version of the demo you are using? (please provide either a specific
603af3652b6bcfd383c6b61e73ca5302f7289253

## Symptom

It seems that the span attribute `app.session.id` is not available anymore in the demo.

**What is the expected behavior?**

Either update the [docs](https://opentelemetry.io/docs/demo/manual-span-attributes/#frontend) to remove the `app.session.id` from the list of custom attributes, or ideally, add the span attribute back.

disfluxly:
I'd definitely prefer that it be added back instead of removing it from the docs, since the Session ID is displayed on the bottom of the GUI pages. Made it really easy to walk someone through a demo and see all of the traces they made.

PR #795, opened 2023-03-17
Add session ID attribute to frontend spans

martinkuba:
# Changes

This adds back the `app.session.id` span attribute that was originally handled in the Go frontend server. The attribute is added for browser spans as well as some spans from the frontend server.

Resolves https://github.com/open-telemetry/opentelemetry-demo/issues/758


cartersocha:
Please add a changelog item too when you get a chance

cartersocha (approved):
lgtm

cartersocha:
Should we note this only appears on non-synthetic requests in our docs? @open-telemetry/demo-approvers 

Also maybe it's just me but I find front-end, front-end-web, and front-end-proxy confusing. Might be worth re-naming one of the user facing services or the synthetic pair. I'd expect to see my user generated traces in the general front-end bucket but you have to go to another service that starts reporting post website interaction

julianocosta89 (changes requested):
Where is the `app.session.id` supposed to be?
I've re-built the sample, started it, stopped the loadgen, and accessed the frontend manually.
When navigating through the generated spans, I could only find the `id` as part of the `http.url` and there was no `app.session.id` attribute, as we can see in the screenshot below:
Screenshot from 2023-03-17 07-24-23

Am I missing something?

cartersocha:
@julianocosta89 it only shows on user generated traces rn. See my comments above

julianocosta89:
@cartersocha I stopped the loadgen and navigated manually.

martinkuba:
@julianocosta89 I have pulled down the codebase to a different directory, rebuilt from my branch, and I am seeing the `app.session.id` attribute there.

image

@cartersocha Are you seeing this on your end?

julianocosta89 (approved):
Thanks @martinkuba!
Looks like I did something wrong. Re-built it again and now it worked.

LGTM!

cartersocha:
Please add the changelog item and we're ready to merge!

PR #796, opened 2023-03-17
Add logs for Ad service and Recommendation service

mviitane:
# What’s included?

- Enable logs for Ad service and Recommendation service.
- Add OTLP exporters for logs
- Add the filter processor to prevent an error from the Prometheus exporter for duplicate queueSize metric, see https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/18194. The filter processor can be removed when the fault gets fixed.

This PR doesn’t introduce any logs backend. Instead, logs are output only to Logging exporter and can be seen in the console (otelcol).

```otel-col  | 2023-03-17T11:40:22.662Z	info	LogsExporter	{"kind": "exporter", "data_type": "logs", "name": "logging", "#logs": 2}```

After this PR, different logging backends can be easily tested by configuring additional exporters.

### CODE REVIEW

File: docker-compose.yml [hunk 1]
```diff
@@ -78,8 +78,10 @@ services:
       - OTEL_EXPORTER_OTLP_TRACES_ENDPOINT
       - OTEL_EXPORTER_OTLP_METRICS_ENDPOINT
       - OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE
+      - OTEL_EXPORTER_OTLP_LOGS_ENDPOINT
       - OTEL_RESOURCE_ATTRIBUTES
       - OTEL_SERVICE_NAME=adservice
+      - OTEL_LOGS_EXPORTER=otlp
```
julianocosta89, line 84: Just a nit picky, maybe for better maintainability. 
Could we move `OTEL_LOGS_EXPORTER=otlp` to `.env` and leave it in here just the `OTEL_LOGS_EXPORTER`?

julianocosta89 (approved):
LGTM!
We also need to add that to the docs. 
I've created the following issue: https://github.com/open-telemetry/opentelemetry.io/issues/2521

austinlparker (approved):

PR #799, opened 2023-03-21
Fix shipping cost update when new item is added to cart 

martinkuba:
# Changes

This fixes the shipping cost update when any parameters (like list of products) for the shipping API call change.

The `react-query` component needs a unique key to know whether to make a network call or use cached data. The query parameters are used as a unique key. This is documented [here](https://react-query-v3.tanstack.com/guides/query-keys#if-your-query-function-depends-on-a-variable-include-it-in-your-query-key).

The original bug is listed in https://github.com/open-telemetry/opentelemetry-demo/issues/327


cartersocha (approved):
nice

julianocosta89 (approved):
Thanks @martinkuba! 🤩

PR #791, opened 2023-03-11
[chore] add kubernetes manifest

puckpuck:
Finishes what is needed for #545 

Adds the actual Kubernetes manifest after running `make generate-kubernetes-manifests`.

PR #807, opened 2023-03-24
[chore] use yamllint 1.3.0

puckpuck:
In the never-ending quest to get #791 to pass yamllint checks

Turns out that locally on my system, I have `yamllint` installed with brew, which grabs the latest version (currently 1.30.0). The GH check (via our Makefile) installs this using pip, which specifies version 1.26.3.  We use a yamllint config option of `ignore-from-file`, which was introduced in 1.28.0, and hence that's why it wasn't working.

This PR updates the yamllint version to the latest version (1.30.0)

### DIFF
--- a/.yamlignore
+++ b/.yamlignore
@@ -1,1 +1,1 @@
-/kubernetes/opentelemetry-demo.yaml
+kubernetes/opentelemetry-demo.yaml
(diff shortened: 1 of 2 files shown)


julianocosta89 (approved):
🤞🏽

PR #808, opened 2023-03-24
fix(featureflagservice): use proper root path

mat-rumian:
# Changes

`/feature` was used as `root path` which was causing service dysfunction.

## Merge Requirements

essential items:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ release.
   ([#797](https://github.com/open-telemetry/opentelemetry-demo/pull/797))
 * Fixed shipping update in the frontend UI when number of products in cart changes
   ([#799](https://github.com/open-telemetry/opentelemetry-demo/pull/799))
+* [featureflag] fix root path
+  ([#808](https://github.com/open-telemetry/opentelemetry-demo/pull/799))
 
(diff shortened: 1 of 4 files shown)


austinlparker:
Does this work? I seem to recall having to change the root path because we're fronting the FF service with Envoy and it wouldn't respond correctly without the prefixing.

mat-rumian:
> Does this work? I seem to recall having to change the root path because we're fronting the FF service with Envoy and it wouldn't respond correctly without the prefixing.

I've built local image with changes and finally `featureflagservice` webserver was correctly displayed. All CRUD functionalities were working. I've also tested existing flags, and their are working too (product failing, recommendation cache and adservice failure).

puckpuck:
The `/feature` root path was specified so that you can access the featureflagservice via the frontend proxy. The entire demo is exposed through this proxy, and each UI has their own endpoint. This PR as it stands will break the demo UI.

Screenshot 2023-03-24 at 5 16 49 PM


Noted on the `FEATURE_FLAG_SERVICE_PATH_ROOT` environment variable though. It is passed in but doesn't seem to be used anywhere in the code. We should look to remove it if possible.

PR #809, opened 2023-03-24
Otlp env variables

### LINKED ISSUE #773: Clean up OTEL_EXPORTER_OTLP_(*_)ENDPOINT environment variables
Per the [specification](https://opentelemetry.io/docs/reference/specification/protocol/exporter/#endpoint-urls-for-otlphttp) for the `OTEL_EXPORTER_OTLP_ENDPOINT` and we should not set the path when used (i.e.: `/v1/traces`). If setting the path, this needs to be done with the telemetry-specific variable i.e.: `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`. Some SDKs or use cases require the use of telemetry-specific variables. When that is the case, the generic one should not be included in that service's container.

The telemetry-specific environment variables are used in nearly all services, but the default value specified in the `.env` file is without the path specified, violating the specification. A few services do override the default and specify the path. We should make the path part of the defaults in `.env`. 

The frontend demo component (perhaps others) specifies both the telemetry-specific and the non-telemetry-specific environment variables. This should be cleaned up as well.

puckpuck:
Fixes #773 

Updates `OTEL_*` environment variable usage to conform with the [specification](https://opentelemetry.io/docs/reference/specification/sdk-environment-variables/).

For the most part, this affected the `OTEL_EXPORTER_OTLP(_*)_ENDPOINT` variables. The spec calls for telemetry-specific endpoints to have the path specified, while the non-telemetry ones should not have a path set. We have a small handful of services that need to specify the path or can't use the default gRPC protocol due to SDK and/or deployment limitations. We should use the telemetry-specific environment variables for those services with the proper protocol port and path specified.

The `OTEL_LOGS_EXPORTER` environment variable is set only for the adservice. The [Java agent](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md#otlp-exporter-span-metric-and-log-exporters) does not set a default of `otlp` for this variable, even though the [spec](https://opentelemetry.io/docs/reference/specification/sdk-environment-variables/#exporter-selection) calls for `otlp` to be set by default. This could be because Logs are not yet considered stable in Java.

As part of this effort, I also removed the `FEATURE_FLAG_SERVICE_PATH_ROOT="/feature"` environment variable that was unused by the featureflagservice.


julianocosta89 (approved):
Good one!

Issue #749, opened 2023-02-19
Frontend requests correct span status  but  wrong http.status_code and two independent traces produced instead of one


devrimdemiroz:
# Bug Report
Version:
[b2f729a56884e3b55e1e203897c9644d8d92a8a9](https://github.com/open-telemetry/opentelemetry-demo/pull/747/commits/b2f729a56884e3b55e1e203897c9644d8d92a8a9)
v 1.3.0 release

## Symptom
See the screenshot attached.

**1-** Frontend request returns an error but sets http.status_code to 200 always. Following is the code segment.

[frontend/pages/api/cart.ts](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/frontend/pages/api/cart.ts)
case 'POST': {
      const { userId, item } = body as AddItemRequest;

      await CartGateway.addItem(userId, item!);
      const cart = await CartGateway.getCart(userId);

      return res.status(200).json(cart);
    }

**2-** Two separate trace is observed, the second trace has reference spanid as FOLLOWS_FROM. Better (even async) to have one trace that starts from the loadgenerator until the end of the transaction.

## Reproduce

When load generator runs, you can search "http.status_code=200" in traces and look into ones with "status=error" 

## Additional Context

<img width="3347" alt="Screenshot 2023-02-19 at 19 42 36" src="https://user-images.githubusercontent.com/62701796/219969203-aad57a77-6916-4e47-9a5d-e0716b33e2ea.png">

PR #810, opened 2023-03-26
[frontend] fix http.status_code on error

puckpuck:
Fixes: #749 

The `http.status_code` for synthetic requests to the frontend, was being set at span creation time. This changes that to set it just before the span ends. Looks like the `app.frontend.requests` metric was also affected by this issue.

Also removed some duplicated error handling code.


julianocosta89:
Hello @puckpuck 👋🏽 

Not sure what is going on here, maybe @martinkuba can help us out, but I've tested this change locally and I didn't get the expected result.

I've built the new frontend and manually killed the cart-service.
The error spans were still with status `200`:
image

I do get some `500`, but not all errors are correctly mapped:
image

Any thoughts?

julianocosta89:
When testing it I started getting some errors in the logs like this:
```
frontend                 | Can not execute the operation on ended Span {traceId: 9a72857f7ceef2b0e3a7856e0856e4e7, spanId: 214218836dea381f}
frontend                 | You can only call end() on a span once.
```

Besides that, it looks good to me.

martinkuba:
I am looking at this as well, and I am seeing the `Can not execute the operation on ended Span` error as well. It looks like it is this [line](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/frontend/utils/telemetry/InstrumentationMiddleware.ts#L67). It is happening on the main branch for me as well. I think it makes sense since the span is from the http instrumentation when the site is NOT running in synthetic mode. I am guessing we just never noticed this before? The fix would be to end the span manually only when it is also created in this same file.

puckpuck:
Oh, I see. It only happens when you access the app from a browser UI (non-synthetic requests).

Yeah, I can plug in some logic to only close if we created it. Will update the PR later today.

puckpuck:
Updated the PR to only end the span when it's the one that we create to split synthetic requests.

martinkuba (approved):

julianocosta89 (approved):
Awesome! 🤩

cartersocha:
Can you add a changelog item please

PR #814, opened 2023-03-30
Fix to shipping calculation

martinkuba:
# Changes

This fixes shipping calculation when the Comet Book product is added to the shopping cart. The logical expression was evaluated incorrectly.

Fixes one bug in https://github.com/open-telemetry/opentelemetry-demo/issues/327


cartersocha:
Please add a changelog item too! 😎

puckpuck (approved):

Issue #771, opened 2023-03-07
Long Term Testing: Memory Leaks


puckpuck:
After running a 4-week test on a stable EKS cluster using m6i.large nodes I have found the following services to be experiencing memory leaks:
- [ ] `emailservice` (severe)
- [ ] `loadgenerator` (moderate)
- [x] `redis` (moderate)

This chart shows the 3 services' memory consumption and pod restarts over 3 weeks
![Screenshot 2023-03-06 at 9 47 32 PM](https://user-images.githubusercontent.com/1296118/223307442-0e33bbb9-1b77-49bf-96fd-bd48e7989a36.png)

[This PR](https://github.com/protocolbuffers/protobuf/issues/10106) seems to be related to the `emailservice` though I'm unclear on it.
The `redis` leak is likey attributed to a lack of TTL when the `cartservice` is adding items to the cart
The `loadgenerator` may need further investigation to understand how and why the leak is being caused.

I have also found the `frauddetectionservice` to be experiencing a mild leak, though the service has not restarted after 4 weeks of running. I don't think we need to take any action on this at this point, but wanted to report it just in case.
![Screenshot 2023-03-06 at 9 48 18 PM](https://user-images.githubusercontent.com/1296118/223307461-9300472d-ae2f-4c44-8074-9c9b785e118b.png)

cartersocha:
Would basic [garbage collection](https://docs.python.org/3/library/gc.html) potentially help the loadgenerator?

austinlparker:
Closing this as part of issue cleanup. If there's anything that needs to be done, please create a new issue.

PR #816, opened 2023-03-30
Fix python loadgenerator memory leak

cartersocha:
My good friend ChatGPT and I tried a couple different solutions to eliminate the memory leak here. 

The most effective method seemed to be switching to the locustio image as our base image from a generic python option. 

I ran both versions locally for ~40 mins. The locustio image stabilized around 67 MB usage. The regular python image continued to add memory usage overtime and was around 86 MB usage when I stopped it. 
Screenshot 2023-03-30 at 10 41 26 AM
Screenshot 2023-03-30 at 11 20 10 AM


cartersocha:
towards: https://github.com/open-telemetry/opentelemetry-demo/issues/771

puckpuck:
oooh... this looks super interesting.  Will give it a try on a long running tests

Wonder if we can do the same for the emailservice now 😛

puckpuck:
I have this running now. Loadgenerator was OOMing every 55-ish hours before.

julianocosta89 (approved):
Image size grew 100MB, but for the benefit that this is bringing I don't have anything against it.
Nice finding @cartersocha!

cartersocha:
@puckpuck im guessing no news is good news on restarts! Are we good to merge once I resolve the conflicts or have any issues popped up?

puckpuck:
I'll confirm tonite one way or the other.

fatsheep9146 (approved):

puckpuck:
I still see a memory leak, at about the same rate as before :(

Issue #616, opened 2022-12-07
Improve ENV_PLATFORM flag in UI


austinlparker:
<img width="224" alt="Screenshot 2022-12-07 at 10 02 26 AM" src="https://user-images.githubusercontent.com/4140740/206214008-74c20e3e-cff0-4a58-8de2-8d03c420b6bb.png">

This guy here.

It occludes parts of the UI, and I don't think it actually properly updates right now anyway? We should move this to some other part of the UI (maybe in the footer, idk) or just get rid of it entirely.

julianocosta89:
This is populated by the environment variable `ENV_PLATFORM`.
I personally like it, for blue/green deployments it is actually nice to set different values and easily identify which env you are.

Welllll.... Now that I wrote it, not sure if it makes much sense in our scenario 😅

puckpuck:
If we do keep this, it needs to be able to be set to any string, and not be bound to an enum.

austinlparker:
To summarize, the work here is as follows:

- Move the flag so that it doesn't occlude other UI elements
- Allow an arbitrary short string to be passed in rather than selecting from an enum

martinkuba:
Would bottom right corner of the footer work? Like this
<img width="500" alt="image" src="https://user-images.githubusercontent.com/4933147/228684619-aa871246-f766-4043-973e-0dd1b82c7d7f.png">

julianocosta89:
Bottom right is fine to me

PR #818, opened 2023-03-30
Updated ENV_PLATFORM flag

martinkuba:
# Changes

Moved the env flag the the bottom right corner of the footer. I also removed the environment enum, so the text in the flag can be anything. The implication, however, is that the background color will be the same for any value.

Addresses https://github.com/open-telemetry/opentelemetry-demo/issues/616


julianocosta89 (approved):
I hope you don't mind, but I've removed `import PlatformFlag from '../PlatformFlag';` from `src/frontend/components/Layout/Layout.tsx` because that was crashing the build.

After that, I've built the new frontend and the env var `ENV_PLATFORM` worked like a charm!

Thanks for taking care of that @martinkuba!

cartersocha:
Please add a changelog item for this too! 

Looks great overall. Thank you for the help @martinkuba 🙏🏼

PR #820, opened 2023-03-30
Fix email service memory leak

cartersocha:
Updating the docker container version seemed to reduce the memory leak not sure if it solved it. Ran for 30ish mins locally and there was a 10MB memory difference. The updated version seems to stabilize around 46 MB but unclear if it maintains that medium term

Screenshot 2023-03-30 at 1 31 07 PM
Screenshot 2023-03-30 at 12 51 20 PM


cartersocha:
@puckpuck anotha one

puckpuck:
I have been trying the same as well, and noticed it too. 

Ultimately I think the email service leak may be tied to a dependency with the OpenTelemetry tooling that depends on a slightly older version of protobuf
...

Still digging

cartersocha:
Yeah I saw your links in the issue. But just trying different things to see what works. 

Maybe @ahayworth will have found an answer or solution to the OTel dependency issue by now

PR #823, opened 2023-04-04
use async php runtime

brettmc:
# Changes

To better demonstrate PHP's capabilities, use an async runtime (react/http). This means that batch exporters (traces and metrics) are long-lived and more efficient, and they can now use export delays to only send batches after the configured time has elapsed. Update auto-instrumentation extension to install from PECL (the preferred mechanism, which we've just set up), and bump other dependencies to their latest beta versions.

## Merge Requirements

essential items:


julianocosta89 (approved):
Thanks @brettmc!
LGTM!

julianocosta89:
@brettmc 
I was wondering... I saw that OTel PHP got OTel logs merged.
Do we need to update something in the `app/settings.php` in order to get OTel into logs?

brettmc:
> Do we need to update something in the app/settings.php in order to get OTel into logs

Since the logging spec isn't stable yet (there are some proposed changes that I think will lead to minor API changes for us), I'd prefer to wait. Also, we're discouraged from using the logging API directly anyway, I'm actually right now working on an integration for the most-used PHP logging library.
Once those things are done, it's on my list to submit a PR to emit some logs!

julianocosta89:
Awesome @brettmc 🤩,
Thank you!

PR #824, opened 2023-04-05
Generate random errors in cartservice

alanwest:
Similar to #694 but demonstrates errors from the cart service (.NET). I'm unsure if I've done things correctly for adding the new feature flag `cartServiceFailure`. I will submit a corresponding docs PR soon.


puckpuck:
I'm all for adding more scenarios, but I need help understanding how this adds value beyond the other 2 existing error scenarios we already have. This will also add 2 more spans for the feature flag service on 10% (random sample) of all traces for EmptyCart requests.

I don't have any issues with the code and will approve it. I just want to ensure we aren't adding new scenarios "just because."

alanwest:
The motivation is to demonstrate an error recorded using OpenTelemetry .NET. Go and Java are covered with the existing scenarios. When speaking with end users, I have often found it useful to hone in on a particular service in the language they're most familiar with to speak to the features the OpenTelemetry SDKs provide.

alanwest:
@puckpuck thoughts?

julianocosta89:
Hello @alanwest,

We have started some discussion regarding this at one of the SIG meetings, but we didn't go through yet, because of the 1.4 release for KubeCon.
As KubeCon is this week, I don't think we will have time to discuss this through, but please bare with us for a little bit longer.

austinlparker (approved):

alanwest:
@cartersocha Fixed conflicts. After this is merged should I submit a PR to update the docs here https://opentelemetry.io/docs/demo/feature-flags?

cartersocha:
Yes please do @alanwest. Thanks for your patience and following up

PR #829, opened 2023-04-09
[collector] use spanmetrics connector instead of spanmetrics processor

fatsheep9146:
# Changes

I would suggest updating opentelemetry-demo to use the [spanmetrics connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/connector/spanmetricsconnector) instead of the [spanmetrics processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/spanmetricsprocessor), as the latter has been deprecated. 

By doing this we can also let user to take advantage of the new advanced functionality `connector`.

I also modify the spanmetric dashboard to be compatible with breaking changes from the spanmetrics connector, which are listed in (here)[Compatibility with important changes to the Spanmetrics connector
]

## Merge Requirements

essential items:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@ release.
   ([#818](https://github.com/open-telemetry/opentelemetry-demo/pull/818))
 * Update OTel Collector
   ([#822](https://github.com/open-telemetry/opentelemetry-demo/pull/822))
+* Update OTel Collector to use spanmetrics connector instead of spanmetrics processors
+  ([#829](https://github.com/open-telemetry/opentelemetry-demo/pull/829))
 
 ## v0.1.0
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -616,14 +616,14 @@ services:
 
   # OpenTelemetry Collector
   otelcol:
-    image: otel/opentelemetry-collector-contrib:0.74.0
+    image: otel/opentelemetry-collector-contrib:0.75.0
     container_name: otel-col
     deploy:
       resources:
         limits:
           memory: 125M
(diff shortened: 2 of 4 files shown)


julianocosta89:
Assuming this is outdated: https://github.com/open-telemetry/opentelemetry-collector/blob/main/connector/README.md#status
It looks good to me.

If the `connectors` are still in `alpha`, than, I'd rather wait to it to get to at least to `beta`. Otherwise, maintaining all the possible changes will be a pain.

@austinlparker, you have mentioned that the README was probably outdated, right?

julianocosta89:
@fatsheep9146 could you also add a changelog?

fatsheep9146:
> @fatsheep9146 could you also add a changelog?

Done

fatsheep9146:
> If the `connectors` are still in `alpha`, than, I'd rather wait to it to get to at least to `beta`. Otherwise, maintaining all the possible changes will be a pain.

I think this also makes sense. But I just want to experience with this useful feature, since the `connector` and `spanmetrics connector` are both marvelous features introduced into collector. And I think their design is really fully reviewed. 
If you think we should not use them until they enter into beta, I'm also ok with that. :)

@julianocosta89

julianocosta89:
I had some personal issues with beta releases b4 😅 .
But that's a me thing.

I'm good to go either way, just want to hear from everyone first.

austinlparker:
Given the cadence of collector releases and how quickly things move thru maturity phases, I'd argue that we should pick up connectors now in order to get useful feedback for collector devs?

austinlparker (approved):

PR #848, opened 2023-04-17
[chore] update kubernetes manifest

### LINKED ISSUE #840: Kubectl command is not creating namespace
# Bug Report

Which version of the demo you are using? (please provide either a specific
[commit
hash](https://github.com/open-telemetry/opentelemetry-demo/commits/main)
or a specific
[release](https://github.com/open-telemetry/opentelemetry-demo/releases)).

Current main 02837fe7b71dd3b85db0efb253a97a693b5d5e3a

## Symptom

Command from https://opentelemetry.io/docs/demo/kubernetes-deployment/#install-using-kubectl
`kubectl apply -f https://raw.githubusercontent.com/open-telemetry/opentelemetry-demo/main/kubernetes/opentelemetry-demo.yaml`
passed without any additional commands

Demo is not installed as `otel-demo` namespace is not created.

## Reproduce

Just execute `kubectl apply -f https://raw.githubusercontent.com/open-telemetry/opentelemetry-demo/main/kubernetes/opentelemetry-demo.yaml`

## Additional Context

Other commands on this page seems to be also outdated.

puckpuck:
With a Helm chart released for 1.4 of the demo, this is an output of that release. Also forces the namespace to be `otel-demo` and when combined with a docs change will fix #840

* [x] Appropriate documentation updates in the [docs](https://github.com/open-telemetry/opentelemetry.io/pull/2611)

### DIFF
--- a/.env
+++ b/.env
@@ -1,7 +1,7 @@
 
 
 # Images
-IMAGE_VERSION=1.4.0-beta
+IMAGE_VERSION=1.4.0
 IMAGE_NAME=ghcr.io/open-telemetry/demo
 
 # Demo Platform
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ release.
 
 ## Unreleased
 
+* [chore] use `otel-demo` namespace for generated kubernetes manifests
+  ([#848](https://github.com/open-telemetry/opentelemetry-demo/pull/848))
+
+## 1.4.0
+
 * [cart] use 60m TTL for cart entries in redis
   ([#779](https://github.com/open-telemetry/opentelemetry-demo/pull/779))
 * spanmetrics dashboard service&operation rates & latencies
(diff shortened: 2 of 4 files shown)


julianocosta89 (approved):

liuyuanping:
Why did you add "tcp-" before the service port name instead of "grpc-"? For example, the opentelemetry-demo-paymentservice changes from "service" to "tcp-service".

I see that the GoogleCloudPlatform/microservices-demo is written as "grpc". Will there be any problem if I change it to "grpc -"?

The reason why I want to do this is that I only detected TCP traffic and no GRPC traffic in istio.

liuyuanping:
Hello! I hope you're doing well. I noticed that in your configuration, "tcp-" is added before the service port name instead of "grpc-". For instance, the opentelemetry-demo-paymentservice changes from "service" to "tcp-service".

I observed that in the GoogleCloudPlatform/microservices-demo, it's written as "grpc". Could there be any potential issues if I change it to "grpc-" instead of "tcp-"?

The reason I'm considering this change is that currently, I'm only detecting TCP traffic and not seeing any GRPC traffic in Istio. Your insights on this would be greatly appreciated!

PR #852, opened 2023-04-25
Add Business Metric Service

### LINKED ISSUE #850: New Service: Business Metric Processor
We need a new service to demonstrate a couple of concepts:
- Injecting instrumentation using k8s operator
- Custom Java metrics/logs

My suggestion would be for a small Java app that consumes the orders topic then creates new metrics displaying sales/hour, etc. ("business" metrics). However, there would be no OpenTelemetry instrumentation added to the container, it would only be added through operator injection.

austinlparker:
# Changes

Closes #850 

This adds a Java service that deliberately does _not_ enable OpenTelemetry instrumentation at build time. It only takes a dependency on the OpenTelemetry API to create a metric. The intended use of this service is to demonstrate the OpenTelemetry Operator's Instrumentation CRDs (as well as API-only usage of OpenTelemetry).

## Merge Requirements

essential items:


julianocosta89:
Are we adding a new java service?
Wouldn't be better to allow users to enable/disable instrumentation on the services we already have?

cartersocha:
I'm hesitant to add to the existing resource requirements given the kubecon feedback for a smaller footprint

julianocosta89:
> I'm hesitant to add to the existing resource requirements given the kubecon feedback for a smaller footprint

same here

austinlparker:
I don't anticipate this being used in a local deployment -- my 0.02 is that we should address the resource issue by creating a "minimal" set of services to be run locally while allowing for a larger set to be run in "production" deployments

austinlparker:
The advantage of adding a net new service here is that it's a terminal/leaf service, ensuring that disabling the service or running it without instrumentation doesn't impact the rest of the deployment.

puckpuck:
Let's create a plan to have minimal and more expansive deployment scenarios before we add anything new.  Something we can discuss in detail at the next SIG meeting.

cartersocha:
Can merge post simple+ demo

austinlparker:
I'll be able to publicly share the demo flow for this in a few weeks

puckpuck:
After we get #872 finished and merged, we can look to move this as an additional service in an optional module/preset.

PR #857, opened 2023-04-29
[collector] update collector to 0.76.1 and remove connectors feature gates

### LINKED ISSUE #833: Remove `connectors` feature gate option when included by default
The collector uses the new `connectors` feature gate leveraged by the `spanmetrics` functionality in the demo.

When the `connectors` feature gate is deprecated and enabled by default, we need to update the `command` attribute for the Collector in docker-compose.yml

fatsheep9146:
# Changes

update collector to 0.76.1 and remove connectors feature gates

and in 0.76.1 the metric names produced by spanmetric connector are all changed, so the dashboard should also be updated.

fix #833


julianocosta89 (approved):
Approving this, but it seems that this and the previous update on the collector broke most of the dashboards in Grafana.
We can fix this in a different PR though, the charts are already broken, and this PR keeps the `spanmetrics-demo-dashboard` updated and working.

puckpuck (approved):

Issue #866, opened 2023-04-30
Start the demo without Kafka


julianocosta89:
# Feature Request

As a dev that uses the demo to build and test new releases, I'd love to have a way to start the demo without Kafka.
The Kafka service takes ages to start, and building, starting, stopping and repeating is just too slow because of Kafka.

If I could start the demo without it, that would save me a lot of time.

PR #872, opened 2023-05-02
splitting up services into modules

joshleecreates:
Fixes: #866 

- splits up services into multiple docker files
- experiment with changing service definition for checkout service when kafka is enabled

todo:

* [x] fix checkout service to work without kafka
* [x] fix services that error without feature flag service 
* [x] provide noop config for collector when observability backends are disabled

# Changes

## Merge Requirements

essential items:


joshleecreates:
Amazing, thanks @puckpuck !! This is working as expected for me now.

puckpuck:
This one is ready for review. 

It splits out all components except for what is absolutely required to run the demo with an exposed UI. A start and stop script is included to facilitate specifying the proper command line options to start the various modules.

The following modules are what are split out from the core:

- `load` - includes the load generator
- `observability` - includes jaeger, prometheus, and grafana, as well as necessary changes to the collector
- `featureflags` - includes the feature flag service with postrgres, as well as an environment variable flag for several services that have logic for feature flags.
- `kafka-extras` - includes kafka, accounting, and fraud detection services as well as an environment variable flag for the checkout service
- `tests` - tests included with demo

A full version of the demo can be run with a special `full` or `all` module when specified with the start script.


/cc @open-telemetry/demo-approvers

cartersocha:
Wow this is fantastic. Need to spend some more time running it but looks good so far. 

Very exciting update! This opens up some really interesting extensibility options and a bit of maintainer dread 🍷

cartersocha:
It seems to spam the starting and stopping logs for some services. 
Screenshot 2023-05-20 at 1 18 14 PM

cartersocha:
product catalog, ad, checkout, and recommendation are failing when running just the script with no arguments due to the feature flag address env variable not being set

joshleecreates:
@cartersocha 

> product catalog, ad, checkout, and recommendation are failing when running just the script with no arguments due to the feature flag address env variable not being set

This sounds like the images need to be rebuilt maybe? These services work for me in all modes.

`docker system prune` followed by `docker compose build` should then let you run the script. This will be resolved for all users when the version tag gets bumped and the images based on these changes are published.

cartersocha:
Thanks for the tip @joshleecreates. To be more realistic I had pruned docker and just ran the start up script with no arguments when I ran into this bug. 

Wouldn’t docker compose build build all of the images including the ones you don’t want to run?

joshleecreates:
@cartersocha that makes sense to run it that way. Without any `-f` arguments, `build` will only build the "core" components. Alternatively, you could change the tag in your `.env` file to something that doesn't exist publicly which will force docker to build any images it needs based on the `up` commands, or you could just change the image tag for those specific services in docker-compose if you want to avoid rebuilding all core components.

cartersocha:
that fixed it. Thanks @joshleecreates. 

I noticed the redis service and collector get built when running the default script but not default docker compose build. Does it make sense to just include those in the default build?

joshleecreates:
> I noticed the redis service and collector get built when running the default script but not default docker compose build. Does it make sense to just include those in the default build?

Hmm maybe what you saw was those services being pulled/created? Any service without a `build:` key in it's docker-compose spec would never get built locally. This includes the collector, kafka, redis, postgres, and the observability components which all use off-the-shelf public images.

cartersocha (approved):

cartersocha:
Thanks @joshleecreates. Lgtm. 

Very exciting capabilities

austinlparker:
Per discussion on call today and in thread -

- We should try to keep `docker compose up` behavior the same between versions

Open question, could we make `docker compose` take parameters? i.e., `docker compose up lite` or whatever.

joshleecreates:
I can't find the original comment but I believe @julianocosta89 mentioned that `docker compose up` behavior should stay the same as this feature is implemented. I'm curious to hear wider thoughts on what is the most desirable dev experience that supports this enhancement.

Challenges:
- `make` doesn't have a built in mechanism for arguments so it cannot be as dynamically extensible as the bash script provided by @puckpuck 
- the `docker compose -f` syntax is verbose
- Either `bash` or `make` introduce an additional dev environment dependency (right now `make` is only needed for creating releases)

Ideas

# Option 1: Use the bash script and add a `docker-compose.override.yml` to restore default `docker compose up` functionality to be the full demo.
Full demo: `docker compose up`
Modules: `./start-demo.sh module`

# Option 2: Use make with an environment variable for modules and add a `docker-compose.override.yml`.
Full demo: `docker compose up`
Modules: `make run module`

# Option 3: Use make with hard-coded targets for modules and add a `docker-compose.override.yml`
Unlike with options 1 or 2, the script would need to be edited any time new modules are added, which would limit flexibility for vendor extensions on forks.

# Option 4: Restore `docker compose up` and require `docker compose -f` syntax for users who want to run the demo in lite mode. No additional dependencies, and the annoying syntax is applied to the less common use case. We could make a single `docker-compose.lite.yaml` that simplifies the amount of chaining needed to run the "lite" version with the observability backends and feature flag services.

julianocosta89:
Hey @joshleecreates I've mentioned that in here:
https://github.com/open-telemetry/opentelemetry-demo/discussions/927#discussioncomment-6103368

But that's only my opinion, I'm open to discuss this and go with the choice most of people like.

joshleecreates:
@austinlparker makes sense. 

I should have just made this last commit with my comment. It enables Options 1 + 4. 

I *do* find the bash scripts convenient and will probably keep them locally somewhere if the decision is to remove them from the official. If that's the case, then this probably just needs a documentation update showing how an example of the `lite` usage without the scripts.

puckpuck:
As discussed during the June 26th SIG meeting.

`docker compose up` will run the full demo and preserve functionality that was always there before.
The standard `docker-compose.yml` file is the same as it was before, with some updates on how the `depends_on` sections are done to be in line with the latest docker-compose file format.
A new file called `docker-compose.minimal.yml` can be used to run the demo without the Feature Flag service, or anything that depends on Kafka.  Load Generator, and all Observability tooling will still run in this mode, as well as the core components for the demo itself.

Technically someone could specify `docker compose -f docker-compose.minimal.yml up` to run the demo in minimal mode, where the `-f` argument appears in the command line matters, and humans make errors. So we have some new make targets to provide easy buttons to start and stop the demo in full or minimal mode.
- `make start` will start the full demo
- `make start-minimal` will start the minimal demo
- `make stop` will stop all things related to the demo

joshleecreates (commented):
This looks great, I appreciate the simplification over my original breakdown. 🚀 🎉

austinlparker (approved):

PR #878, opened 2023-05-04
enabling batch span processor metrics for quoteservice

brettmc:
# Changes

PHP's batch span processor can emit metrics. There is not yet an async implementation, so use react's event loop to periodically flush traces and metrics. Add a grafana chart for some of the batch span processor metrics: dropped, pending, processed, queued

## Merge Requirements

essential items:


julianocosta89 (approved):
LGTM.
Added the dependencies update in this PR.
Hope you don't mind!

mviitane (approved):
Looks good. For some reason the Recommendation service related graphs don't seem to work, but I think it's not related to this PR.
image

julianocosta89:
> Looks good. For some reason the Recommendation service related graphs don't seem to work, but I think it's not related to this PR. image

Yeah, this dashboard looks weird TBH.
If I select `quoteservice` I wouldn't expect to see any `recommendationservice` metrics.

But not related to this PR, so I'll move on with it.

cartersocha:
I had the same issue @julianocosta89 but apparently it was because we only had a recommendation service failure scenario previously so someone wanted to spotlight the data. but I think it’s odd too with the drop-down

brettmc:
I didn't consider being able to choose other services when I set up that new chart.

There are no semantic conventions that I'm aware of which cover how a batch processor should emit metrics (although likely similarities between SIGs - we based our implementation on Java's). If there were then that chart could show multiple services, and the metrics would be more useful, IMO.

PR #656, opened 2022-12-22
Ensure random ads code branch is exercised by the load generator

jlawrienyt:
# Changes

* Add a new product in a new category, "Books"
* Update load generator script to fetch the new product, as well as request ads in the new book category, and request ads without any category, in order to test the random ad delivery code path
* This is in preparation for [adding custom business metrics to the ad service](https://github.com/open-telemetry/opentelemetry-demo/issues/73).

## Merge Requirements

For new features contributions please make sure you have completed the following
essential items:

* [ ] `CHANGELOG.md` updated to document new feature additions
* [ ] Appropriate documentation updates in the [docs](../docs/) folder

Maintainers will not merge until the above have been completed. If you're unsure
which docs need to be changed ping the
[@open-telemetry/demo-approvers](https://github.com/orgs/open-telemetry/teams/demo-approvers).

PR #905, opened 2023-05-26
Update frontend and integration tests

danielbdias:
# Changes

Today the integration and frontend tests are broken because they validate only 9 products sent by the ProductCatalog, causing our tests to fail when running `make run-tests`.

Also, the email test returns a false positive because we have a validation `t.truthy(true)` on this test. I updated the test data and changed this validation to `t.truthy(res.ok)` to guarantee that this endpoint is returning an `HTTP 200` answer.

This PR fixes the assertions to consider the product added on  https://github.com/open-telemetry/opentelemetry-demo/pull/656 too.

To run these tests locally, first, you need to build both images with the following command:
```sh
docker compose build frontendTests integrationTests
```

And later just run:
```sh
make run-tests
```

## Merge Requirements

essential items:

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -207,7 +207,7 @@ test("email: confirmation", async (t) => {
     headers: { "Contenty-Type": "application/json" },
   });
 
-  t.truthy(true);
+  t.truthy(res.ok);
 });
 
 // --------------- Payment Service ---------------
--- a/test/test.js
+++ b/test/test.js
@@ -254,7 +254,7 @@ test("payment: expired credit card", (t) => {
 
 test("product: list", async (t) => {
   const res = await productList({});
-  t.is(res.products.length, 9);
+  t.is(res.products.length, 10);
 });
 
 test("product: get", async (t) => {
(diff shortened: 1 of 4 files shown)


test/test.js
```diff
@@ -283,8 +283,11 @@ test("recommendation: list products", async (t) => {
   const req = deepCopy(data.recommend);
 
   const res = await recommend(req);
-  t.is(res.productIds.length, 4);
-  t.is(arrayIntersection(res.productIds, req.productIds).length, 0);
+  t.is(res.productIds.length, 5);
+  // TODO: this validation is always returning 2 instead of 0, probably because of 
+  //       the cache leakage scenario, it makes sense to create two validations 
+  //       one for the leakage and another to the "normal" condition?
+  //t.is(arrayIntersection(res.productIds, req.productIds).length, 0);
```
danielbdias, line 290: Folks, I commented on this assertion because I'm not sure what we can do with this assertion. Do you have any suggestions?

austinlparker: The cache leak scenario should never be exercised when tests are running so I'd make it work for the happy path and not the sad one

austinlparker:
Any update on comments here?

danielbdias:
hi @austinlparker ! I was on vacation last week, I'm planning to tackle the issue with the ProductCatalog test in the next few days. I should have this PR ready to review again on Monday.

danielbdias:
Hi folks! Giving a heads up here, I'm still trying to fix the test mentioned in [this comment](https://github.com/open-telemetry/opentelemetry-demo/pull/905#discussion_r1207101535). 
Looking at the trace this test generates, it does not trigger the cache case, although it continues to fail, showing 2 or 3 products in common with the output. 
I'm doing more tests this week, and I believe that I should have a solution until the next SIG meeting on Monday.

PR #945, opened 2023-06-20
patch renovate config per conversation in WG

GaryPWhite:
# Changes

extra changes for #931


austinlparker:
Migrated these changes to #931

Issue #952, opened 2023-06-26
Add GitHub workflow for Go services


pellared:
At least a CI that would check that would check if the Go services can be build so that we can use dependabot for bumping deps.

PR #955, opened 2023-06-28
[chore] build images for PRs

pellared:
Fixes https://github.com/open-telemetry/opentelemetry-demo/issues/952

This should give more confidence when merging Renovate PRs (reference https://github.com/open-telemetry/opentelemetry-demo/pull/931).


.github/workflows/build-images.yml
```diff
@@ -0,0 +1,131 @@
...
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        with:
+          config-inline: |
+            [worker.oci]
+              max-parallelism = 2
+
+      - name: Matrix Build and push demo images
```
Kielek, line 120: I think that push should be done only for release.yml not for checks.

pellared:
```
      push:
        description: Should the images be pushed
        default: false
```

on the top

+

```
push: ${{ inputs.push }}
``` a few lines below

pellared:
https://github.com/open-telemetry/opentelemetry-demo/actions/runs/5398517652/jobs/9804417710?pr=955 is failing

However, I guess the same problem would happen during release and it is rather not related with this PR.

I do not think it should be a blocker for this PR.

GaryPWhite:
> I do not think it should be a blocker for this PR.

I'm not an approver; but I will say that introducing a step that fails should absolutely block the PR to avoid other folks having to address the change. Who else would be suited to resolve the problem?

GaryPWhite:
unless we can tell something got busted in the dependency pipeline it seems like the pipeline got borked somehow with this change, no?

pellared:
> nless we can tell something got busted in the dependency pipeline it seems like the pipeline got borked somehow with this change, no?

Normally I would agree. However, these jobs are used for building the container images during "release" workflow. So right now it gives fast feedback that the release would not work 😉

pellared:
> > I do not think it should be a blocker for this PR.
> 
> I'm not an approver; but I will say that introducing a step that fails should absolutely block the PR to avoid other folks having to address the change. Who else would be suited to resolve the problem?

Normally I would agree. However, these jobs are used for building the container images during "release" workflow. So right now it gives fast feedback that the release would not work 😉

pellared:
> does this only build what's changed in a PR, or everything

Everything 😉

pellared:
> we do need to fix whatever's broken in the build currently...

I created https://github.com/open-telemetry/opentelemetry-demo/issues/956 and also commented the line which needs to be uncommented https://github.com/open-telemetry/opentelemetry-demo/pull/955/commits/018b9ea120dc8d1cb6b3547afa0487a8695561d1

pellared:
Transient error: https://github.com/open-telemetry/opentelemetry-demo/actions/runs/5540546109/jobs/10112777382

Can somebody rerun? 😉

julianocosta89:
The build really takes some time 😢

austinlparker (approved):

austinlparker:
I think we're just gonna have to deal with the build times, but I'm more concerned about bringing stuff in to main that doesn't build properly. PR lgtm, lets goooo

PR #1021, opened 2023-07-28
added Middleware Python APM in demo setup

bhogayatakb:
Update python Dockerfile and lib versions to work well with middleware apm


julianocosta89:
Hello @bhogayatakb, thanks for your PR, but we can't accept it.
We try to keep the demo as vendor agnostic as possible, using just some OSS tools in some cases.

In order to modify the service to work with an specific vendor we recommend forking the repository, working on your changes and sending us a PR just to mention your fork in the main README, as other vendors have done:
https://github.com/open-telemetry/opentelemetry-demo#demos-featuring-the-astronomy-shop

PR #1028, opened 2023-07-31
Pin eclipse for renovate

austinlparker:
Per comments on #1007


trask (approved):
I discovered you can use `.json5` and include comments, e.g.

https://github.com/open-telemetry/opentelemetry-java-examples/blob/04b8372adad13f4030e842b3b8687eca71ea4a9d/.github/renovate.json5#L17-L22

julianocosta89 (approved):

PR #1048, opened 2023-08-08
Revert collector back to contrib distribution

julianocosta89:
# Changes

As discussed in our last SIG meeting, the OCB is great for producation environments, but for demoing purposes it is easier to all vendors forking the demo to simply use the Contrib distribution.
This PR reverts the OCB deployment back to `opentelemetry-collector-contrib`.


fatsheep9146 (approved):

mviitane (approved):
LGTM. Seems to be running as before!

PR #1049, opened 2023-08-08
Line break should be LF

julianocosta89:
# Changes

CartService had its line break changed to CRLF, which causes the build in Ubuntu to fail.
`LF` should be the default.


mviitane (approved):
Didn't see this problem in Mac, but it works fine also with this change.

PR #1129, opened 2023-09-18
Required werkzeug update

mviitane:
# Changes

#1128 also requires updating the Werkzeug package to get it compiling.


julianocosta89:
@mviitane can we add the flask update here as well?
We can then close the #1128 and merge just this one

PR #1132, opened 2023-09-18
Update java agent versions

mviitane:
# Changes

Update java agent versions.

With the old versions, there're problems with NaN metrics from Kafka/Fraud detection service. I noticed they get fixed with this version update.
```
otel-col  | Metric #85
otel-col  | Descriptor:
otel-col  |      -> Name: kafka.consumer.partition_assigned_latency_max
otel-col  |      -> Description: The max time taken for a partition-assigned rebalance listener callback
otel-col  |      -> Unit: 
otel-col  |      -> DataType: Gauge
otel-col  | NumberDataPoints #0
otel-col  | Data point attributes:
otel-col  |      -> client-id: Str(consumer-frauddetectionservice-1)
otel-col  | StartTimestamp: 2023-09-18 12:53:47.837352255 +0000 UTC
otel-col  | Timestamp: 2023-09-18 12:54:47.726008797 +0000 UTC
otel-col  | Value: NaN
```


julianocosta89 (approved):

trask (approved):

mviitane:
It appears the new java agent didn't, after all, fix the NaN metrics. That's why a new issue was created for it:
https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/9498

PR #1134, opened 2023-09-20
(otelcol): metrics filter and transform

puckpuck:
# Changes

Adds a filter to drop the `rpc.server.duration` metric due to high cardinality on this metric in the Go SDK, which causes downstream memory problems in the collector and Prometheus. We should remove this filter definition after implementing a Metrics View for the checkout and productcatalog services, or this [PR](https://github.com/open-telemetry/opentelemetry-go-contrib/pull/3730) on the Go SDK is merged, released, and incorporated into the demo.

Adds a transform to clear the description for the `http.client.duration` metric. This is required to eliminate errors with the Prometheus exporter in the collector. There is a [PR](https://github.com/open-telemetry/opentelemetry-python-contrib/pull/1959) at the Python SDK level to fix this, but that fix will take some time before it's released and incorporated into the demo.

### DIFF
--- a/src/otelcollector/otelcol-config.yml
+++ b/src/otelcollector/otelcol-config.yml
@@ -30,10 +30,6 @@ processors:
           statements:
             # FIXME: remove when this issue is resolved: https://github.com/open-telemetry/opentelemetry-java/issues/4834
             - set(description, "") where name == "queueSize"
-            # FIXME: remove when these 2 issues are resolved:
-            # Java: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/9478
-            # Go: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4301
-            - set(description, "") where name == "rpc.server.duration"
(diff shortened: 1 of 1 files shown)


src/otelcollector/otelcol-config.yml
```diff
@@ -17,16 +17,25 @@ exporters:
...
+    transform:
+      metric_statements:
+        - context: metric
+          statements:
+            # FIXME: remove when this issue is resolved: https://github.com/open-telemetry/opentelemetry-java/issues/4834
+            - set(description, "") where name == "queueSize"
+            # FIXME: remove when these 2 issues are resolved:
+            # Java: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/9478
+            # Go: https://github.com/open-telemetry/opentelemetry-go-contrib/issues/4301
+            - set(description, "") where name == "rpc.server.duration"
```
mviitane, line 36: This metric is already dropped with the filter processor. It should be removed here?

mviitane (approved):

PR #1142, opened 2023-09-24
[otelcol] Prevent overwriting of pipelines

djaglowski:
# Changes

Currently, some elements of the pipelines defined in `otelcol-config.yml` are clobbered by conflicting definitions in `otelcol-observability.yml`. Specifically, the exporters list in the `traces` and `metrics` pipelines. 
```yaml
# in otelcol-config.yml
pipelines:
  traces:
    ...
    exporters: [logging, spanmetrics] # blown away by list in otelcol-observability.yml
  metrics: 
    ...
    exporters: [logging]              # blown away by list in otelcol-observability.yml
```

```yaml
# in otelcol-observability.yml
pipelines:
  traces:
    exporters: [otlp, logging, spanmetrics] # blows away list in otelcol-config.yml
  metrics:
    exporters: [prometheus, logging]        # blows away list in otelcol-config.yml
```

Since we expect users to customize the configuration in many cases, I see the following issues:

1. If a user adds an exporter directly to one of the lists in `otelcol-config.yml`, it has no effect. There's no error or obvious way to detect this - it's just silently ignored.
2. The two exporter lists in `otelcol-config.yml` are effectively ignored so they _could_ be deleted. However, this would mean that it is not a valid config file on its own. i.e. A user who excludes `otelcol-observability.yml` may be surprised to find that this breaks the service.

The solution I'm proposing uses the `forward` connector to _mostly_ decouple the two files.
- Each pipeline in `otelcol-config.yml` is split into two, connected by the `forward` connector. One pipeline receives and processes data. The other exports. The functionality is the same, but the pipelines are no long clobbered _and_ its possible to exclude `otelcol-observability.yml`.
- The pipelines in `otelcol-observability.yml` are separated from those in `otelcol-config.yml`. Instead, they receive data from `forward` connector. This allows any additional components to be added to pipelines in either file without conflicts.

The downside of this approach is that it increases the apparent complexity of the configuration. This may be a barrier to entry for some users, even if it is less error prone.

## Merge Requirements

essential items:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ release.
   ([#1121](https://github.com/open-telemetry/opentelemetry-demo/pull/1121))
 * [kafka frauddetection adservice] update java agent versions
   ([#1132](https://github.com/open-telemetry/opentelemetry-demo/pull/1132))
+* [otelcol] use forward connector to decouple pipelines
+  ([#1142](https://github.com/open-telemetry/opentelemetry-demo/pull/1142))
 
 ## 1.5.0
 
--- a/src/otelcollector/otelcol-config.yml
+++ b/src/otelcollector/otelcol-config.yml
@@ -29,6 +29,7 @@ processors:
           - set(description, "") where name == "rpc.server.duration"
 
 connectors:
+  forward:
   spanmetrics:
 
 service:
@@ -36,12 +37,23 @@ service:
     traces:
       receivers: [otlp]
       processors: [batch]
-      exporters: [logging, spanmetrics]
+      exporters: [spanmetrics, forward]
+    traces/logging:
+      receivers: [forward]
+      exporters: [logging]
+
     metrics:
       receivers: [otlp, spanmetrics]
       processors: [transform, batch]
+      exporters: [forward]
+    metrics/logging:
+      receivers: [forward]
       exporters: [logging]
+
     logs:
       receivers: [otlp]
       processors: [batch]
+      exporters: [forward]
+    logs/logging:
+      receivers: [forward]
       exporters: [logging]
--- a/src/otelcollector/otelcol-observability.yml
+++ b/src/otelcollector/otelcol-observability.yml
@@ -15,7 +15,9 @@ exporters:
 
 service:
   pipelines:
-    traces:
-      exporters: [otlp, logging, spanmetrics]
-    metrics:
-      exporters: [prometheus, logging]
+    traces/observability:
+      receivers: [forward]
+      exporters: [otlp]
+    metrics/observability:
+      receivers: [forward]
+      exporters: [prometheus]


julianocosta89:
Hello @djaglowski, thanks for that!
I like the suggestion and I'd be up for adding it!

This is actually a problem that is there for some time already: https://github.com/open-telemetry/opentelemetry.io/issues/3065

Your approach would solve it in a better way!

djaglowski:
Thanks for the feedback @julianocosta89. I've added a changelog and proposed https://github.com/open-telemetry/opentelemetry.io/pull/3311 to update the docs. 

It's not clear to me if the same changes need to be applied to the kubernetes/helm config & docs because it seems users may be more inclined to modify the primary config directly. What do you think?

julianocosta89 (approved):
LGTM

julianocosta89:
Regarding the Helm chart I do agree with you.
Just pinging @puckpuck to bring him to the discussion

puckpuck:
I'm not quite clear on the use case we are trying to solve for here. We have `otelcol-config-extras.yml` which is where people and vendors should be doing all changes.

The demo should strive to try and adhere to idiomatic conventions where possible on how people should be instrumenting and managing the associated telemetry for their applications. The collector configuration in this PR significantly complicates the configuration away from those conventions.

I think we should clarify how to extend the telemetry pipeline in docs using `otelcol-config-extras.yml` as mentioned in this [issue](https://github.com/open-telemetry/opentelemetry.io/issues/3065)

djaglowski:
> I'm not quite clear on the use case we are trying to solve for here. 

My use case is as follows: I'd like to export all telemetry data via one additional exporter.

> We have `otelcol-config-extras.yml` which is where people and vendors should be doing all changes.
>
> The demo should strive to try and adhere to idiomatic conventions where possible on how people should be instrumenting and managing the associated telemetry for their applications. 

The problem is that we're asking users to write configuration in the `extras` file which clobbers configuration in the other files.

This is not simple or idiomatic in my opinion. Anecdotally, I'm a maintainer of the collector but I still spent hours struggling to accomplish my use case because I was of a mind that I could make additive changes in one file without needing to be concerned for the others.

To oversimplify the situation in order to highlight the problem, we basically have the following:

```yaml
# otelcol-config.yaml
exporters: [a, b]

# otelcol-observability.yaml
exporters: [a, b, c]

# otelcol-extras.yaml
# Add whatever you want to "exporters"!
exporters: [z] # User "added" z
```

The end result here is that `[a, b, c]` clobbers `[a, b]`, and then `[z]` clobbers `[a, b, c]`.
1. There's no point in defining `[a, b]` in the first place, if we expect everyone to let it be clobbered.
2. More to the point, we shouldn't encourage novice users to clobber anything. The worst part is that there's no error message that says this is happening because this is "expected" behavior when merging config elements that are lists. The user may notice that they aren't seeing data at `a`, `b`, or `c`, but the reason why is not going to be apparent to most users.

Effectively, we're presenting the `extras` file as an independent sandbox, but it seems to me that the design and guidance will steer most users to unintentionally invalidate elements of the other files.

> The collector configuration in this PR significantly complicates the configuration away from those conventions.

I think whatever conventions we were trying follow were misguided and will lead to frustration more often than results.

I agree that my proposed changes are more complicated, _if the goal is for novice users to grasp the totality of the configuration_. However, if the goal is _to make it easy for new users to add their own backend_, then I think what I've proposed is a much more reliable solution.

> I think we should clarify how to extend the telemetry pipeline in docs using otelcol-config-extras.yml as mentioned in this https://github.com/open-telemetry/opentelemetry.io/issues/3065

There is an accompanying PR [here](https://github.com/open-telemetry/opentelemetry.io/pull/3311) which I believe clarifies the expectations for those running with docker compose. The issue you mentioned appears to specifically relate to the kubernetes equivalent. I'd rather defer to others as to whether the same changes make sense there, but my intuition was that it is less of an issue.

julianocosta89:
I have to disagree @puckpuck.
Besides the cases already shown by @djaglowski in his initial description: https://github.com/open-telemetry/opentelemetry-demo/pull/1142#issue-1910234339

The point 1 is that we have some users struggling to use the `bring-your-own-backend` instructions since we introduced `spanmetrics`.
Second, atm, our docs do not create a 2nd pipeline, which causes the overwrite of the `traces`:

```
service:
  pipelines:
    traces:
      receivers: [otlp]
      processors: [batch]
      exporters: [otlphttp/example]
```

And breaks the `metrics` pipeline.

By simply creating a new pipeline we would avoid breaking the default ones:
```
service:
  pipelines:
    traces/byob:
      receivers: [otlp]
      processors: [batch]
      exporters: [otlphttp/example]
```

But in this approach we would have to duplicate everything.

In the suggested solution, we demonstrate a new connector and the code below wouldn't break anything:
  ```
  service:
    pipelines:
      traces/example:
        receivers: [forward]
        processors: [] # optional
        exporters: [otlphttp/example]
  ```

Another great point of the `forward` connector is that when used in the `metrics` pipeline we also get the `spanmetrics` into it (because it was defined before), so we do not need to keep re-adding `receivers` and `exporters` .

mviitane:
I have also struggled quite a bit with the current Collector config. I like this new proposal because when bringing a new backend, I don't need to care what's in the `otelcol-observability.yml`. As I understood, this fully removes the dependency between `otelcol-config-extras.yml` and `otelcol-observability.yml`.

To help users with this new setup, a picture would be nice, showing the forward connector and 3 different config files, with an example pipeline to the user’s own backend.

mviitane (approved):

puckpuck:
My issue is this will complicate a collector configuration that people use as a reference implementation to make it slightly easier to bring your backend. And even that slightly easier, I will argue, since people will need to know to use `forward` as the receiver. That will need to be documented. Similar is true if we document for people to include `spanmetrics` as an exporter for traces if they override that pipeline. In both cases, we need to enrich our docs. 

This PR is just trading one problem (documenting spanmetrics in exporter) with another (documenting forward in receiver), while also making the reference configuration far more complicated.

djaglowski:
> This PR is just trading one problem (documenting spanmetrics in exporter) with another (documenting forward in receiver), while also making the reference configuration far more complicated.

The problem I'm trying to solve is neither of these.

In any case, I think that if the goal is to provide a simple reference implementation, there should be only one config file so that no merging is necessary.

djaglowski:
> We only do the merging to make it easier to bring your own backend. We should probably add comments to the other files instructing people not to modify them and instead add what they want to the otel-config-extras.yml file instead.

The whole point is that modifying `otel-config-extras.yml` is problematic because it effectively destroys content in the other files. It doesn't make anything easier and you can't solve this with comments.

Issue #864, opened 2023-04-30
Jaeger Monitor not compatible with current spanmetrics setup


elbakerino:
# Bug Report

Which version of the demo you are using? 

[471204eec5031dffcb241c143822d7e4905abc75](https://github.com/open-telemetry/opentelemetry-demo/tree/471204eec5031dffcb241c143822d7e4905abc75)

## Symptom

Jaeger Monitor does not show metrics: "No data yet! Please see these[ instructions ](https://www.jaegertracing.io/docs/latest/spm/)on how to set up your span metrics."

## Reproduce

Clone the mentioned version, start all up, open Jaeger Monitor and you won't see metrics.

## Additional Context

It seems that `Jaeger` requires the setup of the `spanmetrics` PROCESSOR, but the e.g. Grafana Dashboard here requires the format of the `spanmetrics` CONNECTOR.

See this [SO post for other context](https://stackoverflow.com/questions/74991374/prometheus-is-not-reading-otel-data).

With adding the deprecated PROCESSOR, the Monitor tab works again.

What would be the correct fix or what other options would be the "reference setup" here?

With this config, an additional PROCESSOR is added - and the Monitor tab works again, but now data is of course written twice to Prometheus.

```yaml
receivers:
    otlp:
        protocols:
            grpc:

exporters:
    otlp:
        endpoint: "jaeger:4317"
        tls:
            insecure: true
    logging:
    prometheus:
        # this prometheus is used by `spanmetrics` CONNECTOR + PROCESSOR
        endpoint: "otelcol:9464"
        # these configs apply to the `spanmetrics` CONNECTOR
        resource_to_telemetry_conversion:
            enabled: true
        enable_open_metrics: true # even with `false`, the data wasn't compatible with `jaeger`

processors:
    batch:
    # deprecated spanmetrics PROCESSOR, used for compatibility with Jaeger Metrics
    spanmetrics:
        metrics_exporter: prometheus

connectors:
    # new spanmetrics CONNECTOR
    spanmetrics:

service:
    pipelines:
        traces:
            receivers: [ otlp ]
            processors: [ batch, spanmetrics ] # configuring spanmetrics PROCESSOR (deprecated)
            exporters: [ otlp, spanmetrics ] # configuring spanmetrics CONNECTOR (not compatible w/ jaeger)
        metrics:
[...]

PR #1174, opened 2023-10-13
(jaeger): Fix jaeger span metrics

puckpuck:
# Changes

Fixes: #864 

Adds proper command args to Jaeger to understand the proper metric names generated from the `spanmetrics` connectors.

## Merge Requirements


julianocosta89 (approved):
LGTM

julianocosta89:
@puckpuck just one question, what is the file: `src/prometheus/prometheus-config.yaml` that is empty but shouldn't be removed?

puckpuck:
> @puckpuck just one question, what is the file: `src/prometheus/prometheus-config.yaml` that is empty but shouldn't be removed?

Prometheus will error out if a config file is not passed into the startup, since the default image does not have one. Typically Prometheus is used to pull metrics, so a config will always be present, but since we are using it only in a push method, we don't have a scraping config to pass in.

```
prometheus  | ts=2023-10-20T03:47:16.236Z caller=main.go:487 level=error msg="Error loading config (--config.file=prometheus.yml)" file=/prometheus/prometheus.yml err="open prometheus.yml: no such file or directory"
```

austinlparker:
Turns out this file was _kinda_ important as it was scraping collector metrics. Removing it broke the collector dashboard in Grafana.

PR #1180, opened 2023-10-14
Fix service graph by enabling client spans in envoy proxy

cedricziel:
# Changes

The service graph generator uses a pair of server -> client spans to determine relations for many
scenarios. The current configuration of envoy will only produce kind=server spans. This change enables
envoy to create relevant client spans.

Bildschirmfoto 2023-10-14 um 19 08 25

## Merge Requirements

essential items:


julianocosta89:
@cedricziel could you add a change log?

cedricziel:
@julianocosta89 done :)

src/frontendproxy/Dockerfile
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
-FROM envoyproxy/envoy:v1.27-latest
+FROM envoyproxy/envoy-dev:6ef1bb476da8d050347b462b2d8efa1cddeb420f
```
puckpuck, line 5: This needs to be set to a released version. Using dev builds will lead to problems.

puckpuck: Does 1.28 have a target release date yet? My cursory checking says envoy releases quarterly, and it's been almost 3 months since 1.27 was released.

julianocosta89: 1.27.1 was release last week 🙃

julianocosta89: @cedricziel is the new config only available in `1.28`?

cedricziel: the `spawn_upstream_span` is only available in 1.28 and in 1.27, the `start_child_span` option starts another server span, which is semantically incorrect. Envoy 1.28-stable is expected any day now according to their schedule. My proposal is to merge this, as we don't have plans to cut a release soon-ish (please correct me 😬 ) and I'll add a followup when 1.28 is there.

julianocosta89: Fine for me, @puckpuck ?

puckpuck: [1.28.0](https://github.com/envoyproxy/envoy/releases/tag/v1.28.0) was released yesterday. Can we update this PR to use that instead?

puckpuck (approved):

Issue #1164, opened 2023-10-06
Add browser incidents in opentelemetry-demo


jordibisbal8:
# Feature Request

The company Skyscanner is currently using the opentelemetry-demo to gain knowledge on debugging/understanding production systems. However, we have observed that all the incidents are injected in a backend service, and we would like to introduce the functionality to also inject them in the frontend browser.

**Describe the solution you'd like:**

Since the load generator is using Locust, we first thought about using a Locust plugin for Webdriver ([example](https://github.com/SvenskaSpel/locust-plugins/blob/master/examples/webdriver_ex.py).

**Describe alternatives you've considered.**

An alternative would be to directly use the Webdriver in a Python script without any use of locust.

## Additional Context

Instead of directly implementing it in our company, we would like to first discuss it here and make it a standard and reusable solution for anyone.

danielgblanco:
Thanks @jordibisbal8. Summing a conversation with @austinlparker and @puckpuck on the `#otel-community-demo` channel:

* Probably a good idea to break this into two separate changes: one to create incident/regression and one to add a Locust test case to hit that client-side JS
* Probably a better idea to use the [Locust Playwright plugin](https://github.com/SvenskaSpel/locust-plugins/blob/master/examples/playwright_ex.py). The Webdriver plugin would require to standup a Selenium server (which would have to be on the same pod). This is not the case with Playwright.
* Looking at [Playwright installation](https://playwright.dev/python/docs/intro#installing-playwright) steps it seems we could install it in the same Locust image we currently use via `pip`, without external dependencies.

PR #1266, opened 2023-11-17
Added web browser traffic in loadgenerator

jordibisbal8:
# Changes

In this PR, we aim to add browser traffic by using playwright (as discussed [here](https://github.com/open-telemetry/opentelemetry-demo/issues/1164)). The main changes are:

- Installed playwright in the `DockerFile`
- Increase memory constraints for the load generator container to be able to run playwright for several concurrent users
- Added playwright user and two tasks in the locustfile. One goes to the `/cart` page and the other clicks the `Go shopping `button.
- Added environment variable to enable browser traffic (false by default)

## Merge Requirements

essential items:

* [x] Appropriate Helm chart updates in the [helm-charts][]https://github.com/open-telemetry/opentelemetry-helm-charts/pull/959


src/loadgenerator/Dockerfile
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
-FROM python:3.11-slim-buster as base
+FROM python:3.11-slim-bullseye as base
```
jordibisbal8, line 5: Used a newer python image version that supports playwright :)

src/loadgenerator/locustfile.py
```diff
@@ -129,3 +131,33 @@ def on_start(self):
...
+        @pw
+        async def open_home_browser_page(self, page: PageWithRetry):
+            try:
+                async with event(self, "Load up Astronomy Shop home page"):
+                    await page.goto("/")
+                async with event(self, "Click on a button"):
+                    async with page.expect_navigation(wait_until="domcontentloaded"):
+                        await page.click('button:has-text("Go Shopping")')
+            except:
+                pass
```
danielgblanco, line 163: Nitpick, new line missing here

src/loadgenerator/locustfile.py
```diff
@@ -129,3 +131,33 @@ def on_start(self):
...
+
+        @task
+        @pw
+        async def open_home_browser_page(self, page: PageWithRetry):
+            try:
+                async with event(self, "Load up Astronomy Shop home page"):
+                    await page.goto("/")
+                async with event(self, "Click on a button"):
```
danielgblanco, line 159: Should this be `Click on Go Shopping` button?

src/loadgenerator/locustfile.py
```diff
@@ -129,3 +131,33 @@ def on_start(self):
...
         self.index()
+
+
+browser_traffic_enabled = os.environ.get('LOCUST_BROWSER_TRAFFIC_ENABLED', False)
+
+if browser_traffic_enabled:
+    class WebsiteBrowserUser(PlaywrightUser):
+        headless = True  # to use a headless browser, without a GUI
+        multiplier = 1  # run concurrent playwright sessions/browsers for each Locust user
+        wait_time = between(1, 10)
```
danielgblanco, line 142: I don't think this used?

danielgblanco (commented):
I've left a couple of comments (not a maintainer to approve). I think this is a great addition and will help build up client-side instrumentation use cases.

src/loadgenerator/locustfile.py
jordibisbal8: This is just an event being logged, which does not produce any error. However, I do agree that Click on Go shopping button would be a better naming :)

jordibisbal8: https://github.com/open-telemetry/opentelemetry-demo/pull/1266/commits/45e0a4ef80d0ee9149d4812420e8a41837396097

src/loadgenerator/locustfile.py
jordibisbal8: https://github.com/open-telemetry/opentelemetry-demo/pull/1266/commits/45e0a4ef80d0ee9149d4812420e8a41837396097

julianocosta89:
This is does not build.
`greenlet` needs to be downgraded to v `3.0.0`

jordibisbal8:
> This is does not build. `greenlet` needs to be downgraded to v `3.0.0`

Thanks for spotting this @julianocosta89, I have downgraded `greenlet`

jordibisbal8:
@julianocosta89 can you have another look now?

jordibisbal8:
> Hello @jordibisbal8 and @danielgblanco 👋🏽
> 
> I've been playing around with the demo since this PR got merged and I couldn't figure it out what that added to the demo. I was expecting that after enabling `LOCUST_BROWSER_TRAFFIC_ENABLED` I would get traces where the root span would be the `frontend-web`.
> 
> It happens that this is not happening, and I only get the `frontend-web` spans, when manually navigating through the browser.
> 
> If that's not what this PR added, could you help me out understanding here what Playwright actually brings to the demo?
> 
> This was a big change on memory allocation, I just want to make sure we are getting something out of it.
> 
> FYI: @austinlparker

Yes, you should be able to see traces with the `loadgenerator` calling the browser. Let me test this again and I will let you know :)

julianocosta89:
> Yes, you should be able to see traces with the `loadgenerator` calling the browser. Let me test this again and I will let you know :) How are you setting the environment up? docker-compose?

Yes:
```
docker compose build
docker compose up
```

jordibisbal8:
Screenshot 2023-12-12 at 10 09 26
Do you see the browser events/tasks in the load generator? Like the screenshot I have attached :)

puckpuck:
I have also noticed a similar experience as @julianocosta89 and did a lot more digging to figure it out.

The current scripts for the playwright user scenarios, hitting the cart endpoint or going home then clicking a button, don't generate a second request. The scenario of going home and hitting a button is navigating to a link anchor built into the home page and not doing another request. 

I modified the script to click on a product and then the add to cart button so that it could generate multiple requests. This worked and did produce traces, **but they were all missing root spans**. This leads me to believe the playwright Javascript engine doesn't properly work with the OpenTelemetry SDK, and that's why we do not see any traces that have a service of `frontend-web`.

It makes sense to open a new issue for this and track it. I don't like the idea of having a 10x memory requirement for the load generator if we don't produce traces that originate from frontend-web. I would like to get this resolved before we release.

jordibisbal8:
@puckpuck thanks for reporting this, I will take a closer look :)

danielgblanco:
I believe the issue is related to having a hardcoded `PUBLIC_OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` pointing to `localhost` in

https://github.com/open-telemetry/opentelemetry-demo/blob/889d455951ac091d9926d151f540d33e4d844b63/.env#L16

and also in the chart values

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/764920aac4d1fe1147ba187dd86631e021c2e297/charts/opentelemetry-demo/values.yaml#L343

In both cases, it expects the collector to be available in `localhost`. In the K8s example, expecting `port-forward` to be run beforehand to achieve this. I can see two main options:

1. Making the same host routable from within the `loadgenerator` container and from outside a cluster (hitting the browser manually). This may add extra complexity.
2. Using a different `PUBLIC_OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` depending on the type of request. As there is a https://github.com/open-telemetry/opentelemetry-demo/blob/889d455951ac091d9926d151f540d33e4d844b63/src/loadgenerator/locustfile.py#L129 this could be doable and we have a different endpoint for those after checking baggage.

danielgblanco:
The issue @puckpuck related to missing parents could be a timing issue though, possibly related to https://github.com/open-telemetry/opentelemetry-js/issues/2205. Would be good to execute some requests, then navigate away to another page (so `unload` is triggered) and wait for `load` or for network idle before finishing the test case. This should ensure that spans are exported.

So, another PR on this would need:

1. For the `frontend` to know when a request is coming from `loadgen` (via `synthetic_request` baggage) and use a different `PUBLIC_OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` (should be configurable via a new env property).
2. Some tests that generate some XHR requests that can be instrumented (not just the initial page load).
3. Trigger an `unload` and wait for the browser to push spans (i.e. wait for next `load` or network idle).

danielgblanco:
Actually, when the batch processor is configured here

https://github.com/open-telemetry/opentelemetry-demo/blob/889d455951ac091d9926d151f540d33e4d844b63/src/frontend/utils/telemetry/FrontendTracer.ts#L34C13-L34C13

We could set a different `exportTimeoutMillis` (or via env var) that allows us to control how long to wait on the Locust side for spans to be exported.

PR #1296, opened 2023-12-13
update cpp and pin to 3.18

austinlparker:
This PR fixes the currencyservice build failure; We also need to update some flags to handle the updated protobuf libraries.


puckpuck (approved):

mviitane (approved):
Fixed the build issue

PR #1302, opened 2023-12-15
[chore] dev tooling: add restart-service 

puckpuck:
# Changes

For developer tooling. I have a script I keep in my local environment, to build, then stop, remove, and create and start that service within an existing local docker deployment of the demo. When done this way you can be guaranteed that the latest code is running. Very handy to quickly see how your changes take place without having to restart the entire demo, or run a bunch of `docker compose ...` commands.

To use this with the checkout service you would run: 

`./restart-service.sh checkoutservice`

or

`make restart service=checkoutservice`


julianocosta89 (approved):
Nice one!

PR #1305, opened 2023-12-19
[chore] - update kubernetes manifests to 1.7 release

puckpuck:
Updates k8s manifest to 1.7 release

`make generate-kubernetes-manifests`


julianocosta89 (approved):
LGTM!
Should we also create the namespace `otel-demo` in this file?

puckpuck:
That's a good idea. I will figure out the right way to do this with the generator script.

mviitane:
A bit related comment: 
Our helm and kubectl install commands in the docs are not aligned with what comes to the namespace. Helm install doesn't include namespace at all, even if it should?, and leads to installing to the default namespace.

puckpuck:
I added a create namespace manifest the hard way to the generated output. Unfortunately, the `helm template ...` command will not emit this for us.

@julianocosta89 can you take another look?

fatsheep9146 (approved):

julianocosta89 (approved):

Issue #748, opened 2023-02-18
Add CI for running .NET tests


sungam3r:
Surprised that CI checks were not running the tests 👎 .
@sungam3r -To unblock you progress here, please comment out the tests temporarily, create an issue to fix them (and assign to me). I hope this is okay with the maintainers of the repo to unblock temporarily.

_Originally posted by @cijothomas in https://github.com/open-telemetry/opentelemetry-demo/pull/746#discussion_r1107931240_

puckpuck:
This issue is to help finish a now merged PR, and references a Closed PR. Closing this issue as such.

If we lack .NET test coverage, please open a new issue with specifics on what coverage we lack.

PR #746, opened 2023-02-12
Avoid calling obsolete methods

sungam3r:
Rel: https://github.com/open-telemetry/opentelemetry.io/pull/2330

PR #1328, opened 2024-01-14
Add Startup.cs to tests

sungam3r:
First part of changes to fix https://github.com/open-telemetry/opentelemetry-demo/issues/748

See https://github.com/open-telemetry/opentelemetry-demo/pull/746#discussion_r1105341222

I'll add CI in separate PR.


Kielek (commented):
You are executing tests, but the code you are executing is not used in `Program.cs`. I think that you should include more changes, and tests _production_ code.

sungam3r:
> You are executing tests, but the code you are executing is not used in Program.cs

Do you mean I shoul revert changes from #489?

Kielek:
> > You are executing tests, but the code you are executing is not used in Program.cs
> 
> Do you mean I should revert changes from #489?

I am not expert in ASP.NET Core, but more or less, yes You should keep changes made directly in Program.cs.

Now, you have created class which is used only by tests, not in the normal program execution.

@cijothomas, FYI

PR #1335, opened 2024-01-16
[currency] - fix metricreader export and temporality

### LINKED ISSUE #1330: Recommendations on updating export interval for currency service
# Question

Hi, I'm trying to run the demo on GCP and hitting a consistent error which I think could be fixed by changing the [export interval of the currency service](https://github.com/open-telemetry/opentelemetry-demo/blob/bfc18c7340d9c5b7c62b15945914feaf0cc6f8d2/src/currencyservice/src/meter_common.h#L34) from 1s to 5s+. I only need this due to a GCP-specific limitation on the monitoring backend, so I was hoping there was a way to do this with configuration instead of recompiling the app.

More specifically, the collector returns an error on export from the GCP metrics service indicating that points are being sent too frequently. I see this occasionally from other demo components too, but it is pretty consistent from the currency service.

First I tried fixing this with collector processors (like the metrics transform processor) to aggregate points in the same batch. I also tried setting `OTEL_METRIC_EXPORT_INTERVAL` env var on the currency service pod, but it doesn't look like that had any effect (I don't know if C++ instrumentation handles that environment variable?)

Wondering if you have any recommendation for tweaking these settings or handling this kind of rate limiting/aggregation in the collector

puckpuck:
# Changes

Fixes #1330

Sets the OTLP metric export temporality to be cumulative (default), similar to other services. Sets the MetricReader to export every 60 seconds (default), similar to other services. 

I left the lines of code to change the options in as comments set to their default values, so other users can see what the options objects are used for.

## Merge Requirements

essential items:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ release.
 
 * [grafana] update grafana to 10.2.3
   ([#1332](https://github.com/open-telemetry/opentelemetry-demo/pull/1332))
+* [currency] fix metric exporter options
+  ([#1335](https://github.com/open-telemetry/opentelemetry-demo/pull/1335))
 
 ## 1.7.2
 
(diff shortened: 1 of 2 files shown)


cedricziel (approved):

julianocosta89 (approved):
LGTM

PR #1347, opened 2024-01-25
Remove container name to enable scaling in compose

cedricziel:
# Changes

Services with fixed container names can not be scaled in docker compose.

This change removes the container name so the container can be scaled like this:

```sh
docker-compose up -d --scale loadgenerator=3
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -362,7 +362,6 @@ services:
   # Load Generator
   loadgenerator:
     image: ${IMAGE_NAME}:${IMAGE_VERSION}-loadgenerator
-    container_name: load-generator
     build:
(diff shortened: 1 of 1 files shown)


julianocosta89:
what about the other container names?
What I'm wondering is why do we have container names at first 🤔

julianocosta89:
It was added in this PR: https://github.com/open-telemetry/opentelemetry-demo/pull/200
And the issue with scaling it was even mentioned in the PR 😅

puckpuck:
Should we do this to other services as well... or really all of them?

Scaling the load generator is certain to lead to a need to scale up other services like frontend and checkout as well.

PR #1359, opened 2024-02-02
[loadgenerator] fix browser traffic enabled flag

puckpuck:
# Changes

The previous way the load generator checked if browser traffic should be enabled needed to be corrected. If the environment variable was present, regardless of its value, the condition was evaluated to be true. Setting `LOCUST_BROWSER_TRAFFIC_ENABLED=false` did not have the intended effect.

This change will only evaluate to true if the environment variable is set to `true`, `yes`, or `on`.

As part of all this, I also lowered Jaeger traces limit to 5000. The traces are getting larger, and this helps with keeping the container useable after running for a while.

## Merge Requirements

essential items:


julianocosta89:
`LOCUST_BROWSER_TRAFFIC_ENABLED=false` works for me

puckpuck:
When I set `LOCUST_BROWSER_TRAFFIC_ENABLED=false` in the docker-compose.yml file, I see the LoadGenerator still doing browser traffic.

The issue is that environment variables are string, not boolean. In Python, `if "false"` will be evaluated as true because it's a non-empty string.

julianocosta89:
Weird, I could swear I've tested it 😅 

You are right.
QQ though, why moving the env var to the docker-compose file instead of keeping it on the `.env`?

puckpuck:
> QQ though, why moving the env var to the docker-compose file instead of keeping it on the `.env`?

The reasoning is the "minimal" deploy of the demo should not do browser traffic as it will keep the requirements low. So since this variable is different between the 2 deployment scenarios we offer I moved it out of the .env and define it in the compose files instead.

julianocosta89 (approved):
Awesome 🤩 
Thanks for all answers @puckpuck

PR #1363, opened 2024-02-02
[productcatalog] - allow products to be extended

puckpuck:
# Changes

Now that we have the `mountedConfigMaps` feature in Helm charts, allowing people to add text-based configuration or data files to any components, we can setup the productcatalog service to have the product list extended.

1. It removes the image path prefix from the catalog itself. This is a frontend implementation detail, so that detail should live within the frontend.
2. Moves the products.json file into a folder, and changes code to load all .json files from that folder.

With this change, using the Helm chart you can add new products like this:

```
components:
  productCatalogService:
    mountedConfigMaps:
      - name: extraProducts
        mountPath: /usr/src/app/products/extra.json
        subPath: extra.json
        data:
          extra.json: |
            {"products":[ ... ]}
```

This could also be done with a volume mount in docker compose.

## Merge Requirements

essential items:


src/productcatalogservice/Dockerfile
```diff
@@ -17,7 +17,7 @@ FROM alpine AS release
 
 WORKDIR /usr/src/app/
 
-COPY ./src/productcatalogservice/products.json ./
+COPY ./src/productcatalogservice/products/ ./products/
```
julianocosta89, line 20: I've played around with this and I believe it would be easier to simply add the products as a volume. WDYT?

julianocosta89: With this we can modify the products without the need to re-build the container image.

puckpuck: This approach (mounting as a volume) would make the Helm deployment kinda messy as we would also need to mount it as a volume + config map in Helm.

julianocosta89: hmmmm, but how are we handling this?
https://github.com/open-telemetry/opentelemetry-demo/blob/main/docker-compose.yml#L562-L563

I couldn't find it on the Helm charts

puckpuck: > hmmmm, but how are we handling this? https://github.com/open-telemetry/opentelemetry-demo/blob/main/docker-compose.yml#L562-L563
> 
> I couldn't find it on the Helm charts

That file is empty and meant to be used as an example. The Helm chart doesn't have a default for it, but if you wanted to update the feature flags, you would use the same `mountedConfigMaps` option from Helm as we would do to add additional products.

Making this a default would cause some messiness in Helm (arrays don't merge), but leaving it as an optional thing makes it easier to manage.

julianocosta89: Got it!
Thanks @puckpuck! 🚀

julianocosta89 (approved):
🚀

PR #1383, opened 2024-02-15
[paymentservice] Add paymentServiceFailure & paymentServiceUnreachable featureflags

EislM0203:
# Changes

Add the featureflags `paymentServiceFailure` & `paymentServiceUnreachable`. 
- paymentServiceFailure: fails calls to the charge method of the paymentservice
- paymentServiceUnreachable: fails calls to paymentservice itself

Added `grpc-ff-client.js` for communication with the featureflagservice.

## Merge Requirements

essential items:

* [x] Appropriate documentation updates in the [docs](https://github.com/open-telemetry/opentelemetry.io/pull/4005)
* [x] Appropriate Helm chart updates in the [helm-charts](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1037)

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,16 +56,16 @@ keeping it up to date for you.
 
 |                                         |                             |                                                                |
 |-----------------------------------------|-----------------------------|----------------------------------------------------------------|
-| [AlibabaCloud LogService][AlibabaCloud] | [Elastic][Elastic]          | [OpenSearch][OpenSearch]                                       |
-| [AppDynamics][AppDynamics]              | [Google Cloud][GoogleCloud] | [Sentry][Sentry]                                               |
-| [Aspecto][Aspecto]                      | [Grafana Labs][GrafanaLabs] | [ServiceNow Cloud Observability][ServiceNowCloudObservability] |
-| [Axiom][Axiom]                          | [Guance][Guance]            | [Splunk][Splunk]                                               |
-| [Axoflow][Axoflow]                      | [Helios][Helios]            | [Sumo Logic][SumoLogic]                                        |
-| [Azure Data Explorer][Azure]            | [Honeycomb.io][Honeycombio] | [TelemetryHub][TelemetryHub]                                   |
-| [Coralogix][Coralogix]                  | [Instana][Instana]          | [Teletrace][Teletrace]                                         |
-| [Dash0][Dash0]                          | [Kloudfuse][Kloudfuse]      | [Tracetest][Tracetest]                                         |
-| [Datadog][Datadog]                      | [Logz.io][Logzio]           | [Uptrace][Uptrace]                                             |
-| [Dynatrace][Dynatrace]                  | [New Relic][NewRelic]       |                                                                |
+| [AlibabaCloud LogService][AlibabaCloud] | [Elastic][Elastic]          | [New Relic][NewRelic]                                          |
+| [AppDynamics][AppDynamics]              | [Google Cloud][GoogleCloud] | [OpenSearch][OpenSearch]                                       |
+| [Aspecto][Aspecto]                      | [Grafana Labs][GrafanaLabs] | [Sentry][Sentry]                                               |
+| [Axiom][Axiom]                          | [Guance][Guance]            | [ServiceNow Cloud Observability][ServiceNowCloudObservability] |
+| [Axoflow][Axoflow]                      | [Helios][Helios]            | [Splunk][Splunk]                                               |
+| [Azure Data Explorer][Azure]            | [Honeycomb.io][Honeycombio] | [Sumo Logic][SumoLogic]                                        |
+| [Coralogix][Coralogix]                  | [Instana][Instana]          | [TelemetryHub][TelemetryHub]                                   |
+| [Dash0][Dash0]                          | [Kloudfuse][Kloudfuse]      | [Teletrace][Teletrace]                                         |
+| [Datadog][Datadog]                      | [Liatrio][Liatrio]          | [Tracetest][Tracetest]                                         |
+| [Dynatrace][Dynatrace]                  | [Logz.io][Logzio]           | [Uptrace][Uptrace]                                             |
 
 ## Contributing
 
@@ -123,6 +123,7 @@ Emeritus:
 [Honeycombio]: https://github.com/honeycombio/opentelemetry-demo
 [Instana]: https://github.com/instana/opentelemetry-demo
 [Kloudfuse]: https://github.com/kloudfuse/opentelemetry-demo
+[Liatrio]: https://github.com/liatrio/opentelemetry-demo
 [Logzio]: https://logz.io/learn/how-to-run-opentelemetry-demo-with-logz-io/
 [NewRelic]: https://github.com/newrelic/opentelemetry-demo
 [OpenSearch]: https://github.com/opensearch-project/opentelemetry-demo


julianocosta89:
Hey @EislM0203, nice to see you here 👋🏽 

Before we check your PR, could you sign the EasyCLA?

EislM0203:
> Hey @EislM0203, nice to see you here 👋🏽
> 
> Before we check your PR, could you sign the EasyCLA?

> Hey there @julianocosta89, its also nice to see you 👋 ! Regarding the CLA, I am working on it

EislM0203:
> Hey @EislM0203, nice to see you here 👋🏽
> 
> Before we check your PR, could you sign the EasyCLA?
Done 🥳

src/paymentservice/charge.js
```diff
@@ -56,4 +62,4 @@ module.exports.charge = request => {
   logger.info({transactionId, cardType, lastFourDigits, amount: { units, nanos, currencyCode }}, "Transaction complete.");
   transactionsCounter.add(1, {"app.payment.currency": currencyCode})
   return { transactionId }
-}
+}
```
julianocosta89, line 65: The file must end with a blank line

src/paymentservice/grpc-ff-client.js
```diff
@@ -0,0 +1,21 @@
+let grpc = require('@grpc/grpc-js');
```
julianocosta89, line 1: Missing copyright:

```
// Copyright The OpenTelemetry Authors
// SPDX-License-Identifier: Apache-2.0
```

src/paymentservice/grpc-ff-client.js
```diff
@@ -0,0 +1,21 @@
...
+      client.getFlag({ name: featureFlag }, (error, response) => {
+        error ? reject(error) : resolve(response);
+      });
+    });
+
+    return response.flag?.enabled || false;
+  } catch (error) {
+    logger.error(`Error getting FeatureFlag: ${error}`);
+  }
+}
```
julianocosta89, line 21: Line break

src/paymentservice/index.js
```diff
@@ -56,4 +62,4 @@ server.bindAsync(`0.0.0.0:${process.env['PAYMENT_SERVICE_PORT']}`, grpc.ServerCr
 )
 
 process.once('SIGINT', closeGracefully)
-process.once('SIGTERM', closeGracefully)
+process.once('SIGTERM', closeGracefully)
```
julianocosta89, line 65: Line break

docker-compose.yml
```diff
@@ -410,6 +410,7 @@ services:
     ports:
       - "${PAYMENT_SERVICE_PORT}"
     environment:
+      - FEATURE_FLAG_GRPC_SERVICE_ADDR
```
julianocosta89, line 413: We also need to add this env var to [docker-compose.minimal.yml](https://github.com/open-telemetry/opentelemetry-demo/blob/main/docker-compose.minimal.yml).

julianocosta89 (commented):
Looks good, quick question though.

Why are we adding 2 environment variables if the error is the same?
I mean, the msg is different, but all the rest is the same.

docker-compose.yml
EislM0203: Good catch, thanks!

EislM0203:
@julianocosta89 @austinlparker 
You raise valid points. To clarify, I see two interesting observability scenarios for the payment service: 

**There is an exception with the payment provider**
What if the response from the payment provider is an error? What if the request fails? The `paymentServiceFailure` feature flag covers this scenario by throwing an exception **in**  the charge method after the corresponding span has been started.

**For some reason the service is unavailable**
As far as I know we currently do not have any feature flag which make an entire service unavailable, which imo is an interesting and valid scenario. I chose a straightforward approach for several reasons. Firstly, I refrained from modifying the underlying Kubernetes resources to make the service unavailable. Secondly, alternatives like dynamically changing the gRPC server port during runtime proved impractical due the way the featureflagservice is implemented in this demo (where 0.5 = 50% activation rate). These methods result in the service being either available or not. The currently implemented approach fails the paymentservice right at the start, if the featureflagservice returns true. This approach not only works quite well with the modus operandi of the featureflagservice but also mimics the behavior of the paymentservice being unavailable. 

Also, the main difference in the implementation of the two featureflags is the location in which the error is thrown. For the `paymentServiceFailure` flag this happens in the charge method _after_ the corresponding span has been started whereas for the `paymentServiceUnreachable` flag the error is thrown right off the bat when the service is called.

I hope this clears up your questions :)

kubernetes/opentelemetry-demo.yaml
```diff
@@ -10093,6 +10093,8 @@ spec:
           - containerPort: 8080
             name: service
           env:
+          - name: FEATURE_FLAG_GRPC_SERVICE_ADDR
```
puckpuck, line 10096: This file is generated after from the Helm chart, after we do a release in that repo. Please revert the changes in this file.

puckpuck:
I like the idea of a service being unreachable, thus `paymentServiceUnreachable` makes sense here. 

I struggle to see the value from the `paymentServiceFailure` flag. We have several feature flags across multiple services that do simple failures similar to this already. If we are adding more feature flags, I would like to see them lead towards different types of failures so we can showcase how to use observability to find them.

puckpuck:
When I use this fork and enable the `paymentServiceUnreachable` flag, I still see the Payment service in my traces. If the service is unreachable, I would expect it to fail completely and not respond to any portion of the request, having the error manifest upstream.

Screenshot 2024-02-21 at 10 52 25 PM

EislM0203:
@puckpuck 
Hey there, sorry for being a bit unresponsive here. To address your comments:

Regarding the `paymentServiceFailure`, while I see your point that there are already somewhat similar flags, they perform slightly different functions. The `adServiceFailure` flag throws an error with a status of resource exhaustion. The `cartServiceFailure` creates a bad cartstore object (with the wrong redis connection URL) and throws an exception as a result. The most similar feature flag to the one I propose is the `productCatalogFailure` feature flag. However, even this one affects only a single product and not an entire (integral) feature of a service. Additionally, given that the payment service uses JS, the feature flag would support showcasing how OpenTelemetry handles errors in JS which is not yet available in the OtelDemo. In short, while I understand your perspective, I still believe that an error in the charge method of the payment service is a probable real-world scenario/problem that people would find interesting to address using observability to find the root cause.

Regarding the `paymentServiceUnreachable` flag, what you are saying makes a lot of sense. The exception occurring at the very beginning when the payment service is called is a result of a compromise, as I refrained from modifying the underlying kubernetes resources to truly render the service unavailable. If you don't like the current implementation, there is one other viable approach: Move the feature flag from the `paymentService` itself to the `checkoutService` and, when enabled, update the `PAYMENT_SERVICE_ADDR` so that it no longer points to the `paymentService`. Please let me know if this would be a valid approach for you. If yes, I will make the changes.

I hope this answers your questions/comments.

julianocosta89:
@EislM0203  we have just merged a PR that removed the current FeatureFlagService in favour of OpenFeature and FlagD.
Unfortunately your PR needs to be revisited.
Would you be able to update it?

EislM0203:
@julianocosta89 Sure, I will look into it

PR #1417, opened 2024-02-27
[frontend] update sharp to latest to avoid docker build issue on M1

basti1302:
Building the frontend container image on an Apple M1 would result in the following error during docker build:

    15.56 > Build error occurred
    15.56 [Error: ENOENT: no such file or directory, copyfile '/app/node_modules/sharp/build/Release/sharp-darwin-arm64v8.node' -> '/app/.next/standalone/node_modules/sharp/build/Release/sharp-darwin-arm64v8.node'] {
    15.56   errno: -2,
    15.56   code: 'ENOENT',
    15.56   syscall: 'copyfile',
    15.56   path: '/app/node_modules/sharp/build/Release/sharp-darwin-arm64v8.node',
    15.56   dest: '/app/.next/standalone/node_modules/sharp/build/Release/sharp-darwin-arm64v8.node'
    15.56 }

Updating sharp to v0.33.x avoids this problem, as that version includes pre-built sharp binaries for various platforms, see https://sharp.pixelplumbing.com/changelog#v0330---29th-november-2023

----

# Changes

* Update version of sharp dependency in frontend service

### DIFF
--- a/src/frontend/package-lock.json
+++ b/src/frontend/package-lock.json
@@ -36,7 +36,7 @@
         "next": "12.3.4",
         "react": "18.2.0",
         "react-dom": "18.2.0",
-        "sharp": "0.32.6",
+        "sharp": "0.33.2",
         "styled-components": "6.1.1",
         "ts-proto": "1.164.0",
         "uuid": "9.0.1"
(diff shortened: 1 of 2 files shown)


mviitane (approved):
I wasn't able to reproduce the build failure on M1 on main/latest, but this seems like a good update anyway.

basti1302 (commented):
The failed integration tests do not look like they are related to this change. Maybe flaky? Might need to be retriggered.

Edit: Weird, now the integration test check is showing up as "Skipped", it showed up as failed before. 🤷

PR #1418, opened 2024-02-27
[frontend] fix typo in ShippingGateway

basti1302:
# Changes

[frontend] fix typo in ShippingGateway

### DIFF
--- a/src/frontend/gateways/rpc/Shipping.gateway.ts
+++ b/src/frontend/gateways/rpc/Shipping.gateway.ts
@@ -8,7 +8,7 @@ const { SHIPPING_SERVICE_ADDR = '' } = process.env;
 
 const client = new ShippingServiceClient(SHIPPING_SERVICE_ADDR, ChannelCredentials.createInsecure());
 
-const AdGateway = () => ({
+const ShippingGateway = () => ({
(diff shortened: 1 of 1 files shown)


austinlparker (approved):

austinlparker:
hey @basti1302 we generally prefer that PRs aren't made from protected branches so that we can merge upstream into them, in the future could you either PR from a personal fork or un-protect branches on your downstream repo?

basti1302:
> hey @basti1302 we generally prefer that PRs aren't made from protected branches so that we can merge upstream into them, in the future could you either PR from a personal fork or un-protect branches on your downstream repo?

Noted. I assume you are referring to this, right? 

image

Just to be clear: I might be wrong, but I do not think this option is available at all if you create the PR from an organization's fork, it seems that in this case it is [not possible](https://github.com/orgs/community/discussions/5634) to allow upstream maintainers to add commits to the PR. Apparently this can only be allowed from personal forks as you have mentioned.

That said, I don't think branch protection has anything to do with it, also the branch I created the PR from is not protected, so I am a bit confused by your mention of me un-protecting the branch in my orgs repo.

Anyway, I'll keep this in mind for the future.

puckpuck (approved):

PR #1433, opened 2024-03-05
[shippingservice] fix context propagation

### LINKED ISSUE #1318: [ShippingService] Parent span not set properly
# Bug Report

[v1.7.0](https://github.com/open-telemetry/opentelemetry-demo/releases/tag/1.7.0)

## Symptom

When `shippingservice` calls `quoteservice` the parent_id sent is not from the caller, but from the caller's parent.
The trace is currenly looking like this:

image

But it should be something like this:

```
checkoutservice oteldemo.ShippingService/GetQuote
    |--> shippingservice oteldemo.ShippingService/GetQuote
        |--> shippingservice POST
            |--> quoteservice POST /getquote
                |--> quoteservice {closure}
                    |--> quoteservice calculate-quote
```

## Reproduce

Simply run the demo with `docker compose up` and check any trace containing `shippingservice` spans.

julianocosta89:
# Changes

Fix #1318.

image

As I was touching this service I've added OTLP logs exporter and refactored the code to make it more readable.
Unfortunately the `opentelemetry-appender-log` doesn't add `spanid` and `traceid` to the log entry, but we can already see the logs in Grafana.

image

## Merge Requirements

essential items:

### DIFF
--- a/src/shippingservice/Cargo.toml
+++ b/src/shippingservice/Cargo.toml
@@ -19,9 +19,9 @@ reqwest = { version = "0.11", default-features = false, features = ["json"] }
 prost = "0.12"
 prost-types = "0.12"
 tokio = { version = "1.36", features = ["macros", "rt-multi-thread"] }
-opentelemetry = { version = "0.21", features = ["trace", "logs"] }
+opentelemetry = { version = "0.21", features = ["trace", "metrics", "logs"] }
 opentelemetry-appender-log = "0.2.0"
-opentelemetry_sdk = { version = "0.21", features = ["rt-tokio", "logs_level_enabled"]}
+opentelemetry_sdk = { version = "0.21", features = ["rt-tokio", "metrics", "logs_level_enabled"]}
 opentelemetry-semantic-conventions = "0.13.0"
 opentelemetry-http = "0.10.0"
 opentelemetry-otlp = { version = "0.14.0", features = ["tonic", "metrics", "logs"] }
--- a/src/shippingservice/Cargo.toml
+++ b/src/shippingservice/Cargo.toml
@@ -19,13 +19,13 @@ reqwest = { version = "0.11", default-features = false, features = ["json"] }
 prost = "0.12"
 prost-types = "0.12"
 tokio = { version = "1.36", features = ["macros", "rt-multi-thread"] }
-opentelemetry = { version = "0.21", features = ["trace"] }
-opentelemetry_sdk = {version = "0.21", features = ["rt-tokio"]}
+opentelemetry = { version = "0.21", features = ["trace", "logs"] }
+opentelemetry-appender-log = "0.2.0"
+opentelemetry_sdk = { version = "0.21", features = ["rt-tokio", "logs_level_enabled"]}
 opentelemetry-semantic-conventions = "0.13.0"
 opentelemetry-http = "0.10.0"
-opentelemetry-otlp = "0.14.0"
+opentelemetry-otlp = { version = "0.14.0", features = ["tonic", "metrics", "logs"] }
 log = "0.4.20"
-simplelog = "0.12"
 reqwest-middleware = "0.2.4"
 reqwest-tracing = { version = "0.4.7", features = ["opentelemetry_0_21"] }
 tracing = { version = "0.1", features = ["max_level_debug", "release_max_level_info"] }
--- a/src/shippingservice/Cargo.toml
+++ b/src/shippingservice/Cargo.toml
@@ -2,7 +2,7 @@
 
 [package]
 name = "shippingservice"
-version = "1.4.0"
+version = "1.8.0"
 edition = "2021"
 
 [[bin]]
@@ -13,12 +13,12 @@ path = "src/main.rs"
 dockerproto = [] # compile in docker or not
 
 [dependencies]
-tonic = "0.10"
-tonic-health = "0.10.2"
+tonic = "0.11"
+tonic-health = "0.11.0"
 reqwest = { version = "0.11", default-features = false, features = ["json"] }
 prost = "0.12"
 prost-types = "0.12"
-tokio = { version = "1.34", features = ["macros", "rt-multi-thread"] }
+tokio = { version = "1.36", features = ["macros", "rt-multi-thread"] }
 opentelemetry = { version = "0.21", features = ["trace"] }
 opentelemetry_sdk = {version = "0.21", features = ["rt-tokio"]}
 opentelemetry-semantic-conventions = "0.13.0"
@@ -27,13 +27,13 @@ opentelemetry-otlp = "0.14.0"
 log = "0.4.20"
 simplelog = "0.12"
 reqwest-middleware = "0.2.4"
-reqwest-tracing = { version = "0.4.6", features = ["opentelemetry_0_20"] }
+reqwest-tracing = { version = "0.4.7", features = ["opentelemetry_0_21"] }
 tracing = { version = "0.1", features = ["max_level_debug", "release_max_level_info"] }
 tracing-opentelemetry = "0.22.0"
 tracing-subscriber = "0.3"
 
 [dependencies.uuid]
-version = "1.5.0"
+version = "1.7.0"
 features = [
     "v4",                # Lets you generate random UUIDs
     "fast-rng",          # Use a faster (but still sufficiently random) RNG
--- a/src/shippingservice/Dockerfile
+++ b/src/shippingservice/Dockerfile
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
-FROM --platform=${BUILDPLATFORM} rust:1.73 as builder
+FROM --platform=${BUILDPLATFORM} rust:1.76 as builder
 
 ARG TARGETARCH TARGETPLATFORM BUILDPLATFORM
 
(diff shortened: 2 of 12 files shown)


julianocosta89:
@TommyCpp would you be able to take a look at this one?

julianocosta89:
We need to implement some quote mock in order to run the tests, and as I'm not a Rust expert I've opted to remove the non-working tests as they were not working since we added `quoteservice` which is a lot!
`quoteservice` was added before we release `v1` and it seems nobody was running the tests anyways.

src/shippingservice/Cargo.toml
```diff
@@ -13,27 +13,27 @@ path = "src/main.rs"
...
+tonic-health = "0.11.0"
 reqwest = { version = "0.11", default-features = false, features = ["json"] }
 prost = "0.12"
 prost-types = "0.12"
-tokio = { version = "1.34", features = ["macros", "rt-multi-thread"] }
-opentelemetry = { version = "0.21", features = ["trace"] }
-opentelemetry_sdk = {version = "0.21", features = ["rt-tokio"]}
+tokio = { version = "1.36", features = ["macros", "rt-multi-thread"] }
+opentelemetry = { version = "0.21", features = ["trace", "metrics", "logs"] }
+opentelemetry-appender-log = "0.2.0"
```
cijothomas, line 23: I'd recommend to use `tracing` and `opentelemetry-appender-tracing`. 
`log` crate does not support structured logging, and we want to showcase best practices in the demo, right?

cijothomas:
Will have to decide if we should rewrite this service using OTel Tracing API. It looks like its currently using an external instrumentation library which converts spans emitted using tokio/tracing crate into OTel spans. It is [undecided ](https://github.com/open-telemetry/opentelemetry-rust/issues/1571) how such usage is going to be supported going forward.

This might also explain why logs are not correlated to spans, as Spans are not produced using OTel Tracing APIs.

TommyCpp:
Sorry missed this earlier

>  It looks like its currently using an external instrumentation library which converts spans emitted using tokio/tracing crate into OTel spans

It's a third party instrument on grpc clients IIRC. Probably better if we rewrite to handle the propagation ourselfs

src/shippingservice/Dockerfile
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
-FROM --platform=${BUILDPLATFORM} rust:1.73 as builder
```
TommyCpp, line 5: Is there a issue with 1.73? We should still support this, if not we need to check our MSRV setup

julianocosta89: No issue, I'm just updating to latest rust version.

austinlparker:
What's the status here, should this be merged?

cijothomas (changes requested):
We should using `tracing` and not `log` crate, if we plan to add logging in this PR.
https://github.com/open-telemetry/opentelemetry-demo/pull/1433/files#r1516487971

or scope down to just fixing context propagation issue alone.

cijothomas:
> What's the status here, should this be merged?

No. I left a blocking comment just now.

Outside of that I'd love to get some opinions from demo maintainers about the Rust app - it is *NOT* using OTel Tracing API, but uses `tokio-tracing` along with a "bridge" (maintained outside OTel repos) to map them to OTel Tracing. OTel Rust SIG is undecided on how to navigate this problem : https://github.com/open-telemetry/opentelemetry-rust/issues/1571

Options:
1. Modify shipping service to *only* OTel Tracing API. If OTel Rust SIG ultimately decides to sacrifice OTel Tracing API, then this requires rework.
2. Keep the existing - but we need to add owners of the respective packages as code reviewers here. (I don't know if they have interest/time). Again, rework is require if OTel Rust SIG decides to go forward with OTel Tracing API and ignore `tokio-tracing`.
3. Something else? Please share.

Either requires rework (after all, no signal is stable in OTel Rust). I'd suggest to do 1. 

@open-telemetry/rust-approvers Please share your thoughts too.

julianocosta89:
I'll take a look at it again soon.

austinlparker:
I see. I'd probably opt for option 1 then.

julianocosta89:
> Either requires rework (after all, no signal is stable in OTel Rust). I'd suggest to do 1.

Do we have any instrumentation library in OTel Rust ATM?
If not, I'd keep the `tokio-tracing` in order to demonstrate how it instruments automatically Rust code.

cijothomas:
> > Either requires rework (after all, no signal is stable in OTel Rust). I'd suggest to do 1.
> 
> Do we have any instrumentation library in OTel Rust ATM? If not, I'd keep the `tokio-tracing` in order to demonstrate how it instruments automatically Rust code.

Nothing maintained by opentelemetry.
Using tokio-tracing also requires another library (tracing-opentelemetry), not maintained by opentelemetry.

julianocosta89:
I'd say we can do that in a follow-up PR.
I've added the `opentelemetry-appender-tracing` and got some logs with Span and Trace ID
image

This should be good to go, and we can continue the discussion in an issue. WDYT?

julianocosta89:
Just my 2 cents here:

I'm in favour of getting rid of `tokio-tracing`.
It is really cumbersome using it and making sure all the dependencies are aligned, as they almost never are on the latest OTel rust release.

julianocosta89:
@cijothomas I'd appreciate if you could double check the code whenever you have some minutes.

src/shippingservice/src/telemetry/resources_conf.rs
```diff
@@ -0,0 +1,25 @@
...
+        EnvResourceDetector, OsResourceDetector, ProcessResourceDetector, ResourceDetector,
+        SdkProvidedResourceDetector, TelemetryResourceDetector,
+    },
+    Resource,
+};
+use std::time::Duration;
+
+pub fn get_resource_attr() -> Resource {
+    let os_resource = OsResourceDetector.detect(Duration::from_secs(0));
+    let process_resource = ProcessResourceDetector.detect(Duration::from_secs(0));
```
lalitb, line 15: Not for now, just fyi - this would be a breaking change in future as OsResourceDetector and ProcessResourceDetector are going to move out of otel-sdk to separate crate - https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/49

src/shippingservice/src/telemetry/logs_conf.rs
```diff
@@ -0,0 +1,29 @@
...
+        .install_batch(runtime::Tokio)
+}
+
+pub fn configure_global_logger(tracer: opentelemetry_sdk::trace::Tracer) {
+    let telemetry = tracing_opentelemetry::layer().with_tracer(tracer);
+    let logger_provider = opentelemetry::global::logger_provider();
+    let layer = OpenTelemetryTracingBridge::new(&logger_provider);
+    let _ = tracing_subscriber::registry()
+        .with(layer)
+        .with(telemetry)
```
lalitb, line 27: what are we using `tracing_opentelemetry` crate for? From what I can see, we are using the otel tracing API for traces, and the tokio-tracing api (through OpenTelemetryTracingBridge) for logs.

src/shippingservice/src/telemetry/resources_conf.rs
julianocosta89: Great to know beforehand!
Thank you!!!

src/shippingservice/src/telemetry/logs_conf.rs
julianocosta89: @lalitb I'm not a Rust expert, I did a bunch of try and fail in order to land in this.
What I can say is that without the `telemetry` layer I wasn't getting the `POST` request span, which comes from the `reqwest` instrumentation, and the trace was broken.
Once I've added it, everything was back to properly working.

cijothomas: Not fully familiar with how tracing-opentelemetry works, but my **guess** of how this is attempted to be setup in this PR is:
1. Uses tokio-tracing as the *logging* API in the shipping service.
2. Uses OpenTelemetryTracingBridge, to send the logs from above to otel->otlp.
3. Also enables `reqwest-tracing` crate, which (I am **guessing** here..), is using tokio-tracing for *tracing* needs (i.e to create Spans, and possibly SpanEvents (in OTel terminology) or just Events in tokio-tracing terminology)
4. The `tracing_opentelemetry` layer is to send the traces from above to otel->otlp.

Good chance there will be duplication of telemetry if `SpanEvents` are used, as (2) will convert them to OTLP Logs, and (4) will convert them to SpanEvents inside OTLP Traces.

Unfortunately, I don't have a good suggestion to fix this, at least until https://github.com/open-telemetry/opentelemetry-rust/issues/1571 is resolved.
I understand it is frustrating, and the same frustration is faced by end users as well, which is why OTel Rust community is trying to address this and not declare 1.0/GA until this is solved.

lalitb: It seems to me that your guess is correct, @cijothomas. And good point on duplicate events (as SpanEvents and Logs) Unfortunately, this can't be avoided if `reqwest-tracing` needs `tracing-opentelemetry`.

julianocosta89: Hmmmm... Okay?

Should I just forget the logs, and only send the PR fixing the Context Prop?

cijothomas: Give us ~1 more week. Will discuss with OTel Rust community and get back to you on how best to resolve this!

cijothomas: From OTel Rust SIG/community meeting:
1. Please scope down this PR, to only address the context propagation issue.
2. Remove tracing and tracing-appender, as that will cause duplicate telemetry via SpanEvent and LogRecord.
3. Come back and add logging via tracing+tracing-appender, once OTel Rust fixes the underlying issue.

julianocosta89: @cijothomas thanks for keeping me in the loop!

I've removed the tracing-appender for logging, but I can't remove tracing as it is used in the `reqwest_tracing`.
I've updated the PR and in theory it should be good to go.

Would you be able to double check?

hdost (commented):
Looks good from my perspective. I'll try running it later today.

austinlparker (approved):

PR #1388, opened 2024-02-18
Remove Feature Flag Service in favor of OpenFeature + flagd

austinlparker:
# Changes

This PR removes the Erlang featureflag service and its dependencies in favor of [OpenFeature](https://openfeature.dev) ecosystem projects.

## Rationale

Feature Flag Service is a bit of a heavy dependency for the project to maintain. We don't have any regular Erlang contributors in maintainer/approver ranks, which makes it difficult to evaluate changes. There's been a fairly long-standing issue with cross-platform builds in QEMU and various Erlang components which has kept the service on outdated versions of OTP. We've had several feature requests for more flexibility around the feature flags that require setting via API, and the lack of a dedicated client for the feature flag service itself makes it somewhat challenging for contributors to add new things (as they have to write handlers for each application).

OpenFeature is a CNCF incubating project that's under active development and provides several things:
- Client libraries for most languages
- Lightweight flag daemon (flagd)
- Kubernetes integration
- Native OpenTelemetry support

## Merge Requirements

For new features contributions please make sure you have completed the following
essential items:

* [x] `CHANGELOG.md` updated to document new feature additions
* [x] Appropriate documentation updates in the [docs][]
* [ ] Appropriate Helm chart updates in the [helm-charts][]

Maintainers will not merge until the above have been completed. If you're unsure
which docs need to be changed ping the
[@open-telemetry/demo-approvers](https://github.com/orgs/open-telemetry/teams/demo-approvers).

[docs]: https://opentelemetry.io/docs/demo/
[helm-charts]: https://github.com/open-telemetry/opentelemetry-helm-charts

PR #1484, opened 2024-03-25
Adservice returns error status code when feature flag is enabled

jack-berg:
When the [adServiceFailture](https://opentelemetry.io/docs/demo/feature-flags/) feature flag is enabled, AdService returns a `RESOURCE_EXHAUSTED = 8` grpc status code. This does not result in the span status being set to error according to the grpc semantic conventions: https://github.com/open-telemetry/semantic-conventions/blob/main/docs/rpc/grpc.md

Updating to return `UNAVAILABLE = 14` so that span status is reflected as error as the feature flag implies.

Side note: The docs site says that when this feature flag is enabled, an error is generated 1/10th of the time. I don't see any code that would limit this to only 10% of the time. Maybe I'm missing something.

### DIFF
--- a/src/adservice/src/main/java/oteldemo/AdService.java
+++ b/src/adservice/src/main/java/oteldemo/AdService.java
@@ -182,7 +182,7 @@ public final class AdService {
                 adRequestTypeKey, adRequestType.name(), adResponseTypeKey, adResponseType.name()));
 
         if (getFeatureFlagEnabled(ADSERVICE_FAILURE)) {
-          throw new StatusRuntimeException(Status.RESOURCE_EXHAUSTED);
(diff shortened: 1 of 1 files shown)


jack-berg:
> Side note: The docs site says that when this feature flag is enabled, an error is generated 1/10th of the time. I don't see any code that would limit this to only 10% of the time. Maybe I'm missing something.

I see. The 1/10th explicit randomness was removed in #1388 but restored via some flagd magic here: https://github.com/open-telemetry/opentelemetry-demo/pull/1388/files#diff-a30814cab829c01672e133c4738a4bff8fd5452a39238f0bbfb32529d9fde441

puckpuck (approved):

PR #1502, opened 2024-04-06
[frontend + adservice] Pass Session ID through baggage

austinlparker:
# Changes

Adds the session ID to the baggage when calling adservice in order to consume it in feature flag context.

## Merge Requirements

essential items:

* [X] `CHANGELOG.md` updated to document new feature additions


src/adservice/src/main/java/oteldemo/AdService.java
```diff
@@ -151,6 +155,15 @@ public void getAds(AdRequest req, StreamObserver<AdResponse> responseObserver) {
         AdRequestType adRequestType;
         AdResponseType adResponseType;
 
+        Baggage baggage = Baggage.fromContextOrNull(Context.current());
+        if (baggage != null) {
+          final String sessionId = baggage.getEntryValue("app.session.id");
+          span.setAttribute("app.session.id", sessionId);
```
puckpuck, line 161: This should be using "session.id" as that is now the [semantic standard](https://opentelemetry.io/docs/specs/semconv/general/session/#attributes). We will need to update it in the frontend service as well.

src/frontend/gateways/Api.gateway.ts
```diff
@@ -82,11 +83,18 @@ const ApiGateway = () => ({
...
   listAds(contextKeys: string[]) {
-    return request<Ad[]>({
-      url: `${basePath}/data`,
-      queryParams: {
-        contextKeys,
-      },
+    // TODO: Figure out a better way to do this so session ID gets propagated to
+    // all endpoints
+    const baggage = propagation.getActiveBaggage() || propagation.createBaggage();
+    const newBaggage = baggage.setEntry('app.session.id', { value: userId });
```
puckpuck, line 89: We should use the AttributeNames enum for the baggage key.

```
import {AttributeNames} from "../utils/enums/AttributeNames";
```

then use `AttributeNames.SESSION_ID`.

puckpuck (commented):
We should also update `utils/enums/AttributeNames.ts` to use the [semantic convention standard](https://opentelemetry.io/docs/specs/semconv/general/session/#attributes) of `session.id` instead of with the app prefix.

puckpuck (approved):

Issue #1437, opened 2024-03-07
loadgenerator failed to start when ipv6 is disabled


MoZhonghua:
# Bug Report

Which version of the demo you are using? (please provide either a specific
1.8.0

## Symptom
loadgenerator failed to start when ipv6 is disabled, with this error:
```text
[2024-03-07 07:16:25,064] 66b18b2f58e5/INFO/locust.main: Starting web interface at http://0.0.0.0:8089 (accepting connections from all network interfaces)
[2024-03-07 07:16:25,074] 66b18b2f58e5/INFO/locust.main: Starting Locust 2.18.2
[2024-03-07 07:16:25,074] 66b18b2f58e5/INFO/locust.main: No run time limit set, use CTRL+C to interrupt
Traceback (most recent call last):
  File "src/gevent/greenlet.py", line 908, in gevent._gevent_cgreenlet.Greenlet.run
  File "/usr/local/lib/python3.12/site-packages/locust/web.py", line 512, in start_server
    self.server.serve_forever()
  File "/usr/local/lib/python3.12/site-packages/gevent/baseserver.py", line 398, in serve_forever
    self.start()
  File "/usr/local/lib/python3.12/site-packages/gevent/baseserver.py", line 336, in start
    self.init_socket()
  File "/usr/local/lib/python3.12/site-packages/gevent/pywsgi.py", line 1668, in init_socket
    StreamServer.init_socket(self)
  File "/usr/local/lib/python3.12/site-packages/gevent/server.py", line 173, in init_socket
    self.socket = self.get_listener(self.address, self.backlog, self.family)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/gevent/server.py", line 185, in get_listener
    return _tcp_listener(address, backlog=backlog, reuse_addr=cls.reuse_addr, family=family)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/gevent/server.py", line 260, in _tcp_listener
    sock = GeventSocket(family=family)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1509, opened 2024-04-08
[loadgenerator] start locust with host "0.0.0.0" (#1437)

MoZhonghua:
starts `locust` with env `LOCUST_WEB_HOST=0.0.0.0`

fix https://github.com/open-telemetry/opentelemetry-demo/issues/1437


puckpuck:
Why is this change needed?

MoZhonghua:
> Why is this change needed?

locust fails to start if ipv6 is disabled. With this env, we force locust to bind ipv4 address.

julianocosta89 (approved):

puckpuck (changes requested):
This implementation is fine, but we should be following the pattern for the other `LOCUST_*` environment variables, which is defining them all in the .env file, and referencing them from the 2 compose files.

SIMILAR to how we LOCUST_WEB_PORT and LOCUST_HOST defined.

MoZhonghua:
> LOCUST_WEB_PORT



> This implementation is fine, but we should be following the pattern for the other `LOCUST_*` environment variables, which is defining them all in the .env file, and referencing them from the 2 compose files.
> 
> SIMILAR to how we LOCUST_WEB_PORT and LOCUST_HOST defined.

I Can't figure out how to change 'LOCUST_WEB_HOST' in .env file, it's already defined in `.env` as `LOCUST_WEB_HOST=loadgenerator`.

I guess `LOCUST_WEB_HOST` in `.env` is used by other service to get node name (or dns name? i don't known) of `loadgenerator`, not used to tell `locust` to bind to this address.

So there are tow different things (in my guess):
- how other service find address of loadgenerator
- which ip address shoud loadgenerator bind

These two things just use same env name by coincidence.

puckpuck:
hmmm... I see now. We did some unfortunate things with the naming of environment variables used for the frontend-proxy.

Let's get this one approved and merged. I'll create a separate issue to track fixing the frontend-proxy.

puckpuck (approved):

PR #1528, opened 2024-04-15
[kafka][checkoutservice][frauddetectionservice] add kafkaQueueProblems featureflag

EislM0203:
# Changes

This PR adds a new feature flag `kafkaQueueProblems` to the opentelemetry demo. Upon activating the feature flag, the producer (`checkoutservice`) overloads Kafka by sending 100 extra messages to the queue per actual order. Simultaneously, the consumer (`frauddetectionservice`) delays the claiming of the messages by 1 second per message. This leads to a sudden spike in consumer lag. This is an interesting, real world observability scenario because it simulates queue problems in kafka which afaik no feature flag does yet. Metrics that monitor consumer lag can be viewed in Grafana (e.g. `kafka_consumer_lag_avg`). 

Also increased the resource limitations of the frauddetection service since it kept dying due to resource exhaustion. This also happened without my modifications to the service.

Looking forward to your feedback!

## Merge Requirements

essential items:

* [x] Appropriate documentation updates in the [docs](https://github.com/open-telemetry/opentelemetry.io/pull/4293)
* [x] Appropriate Helm chart updates in the [helm-charts](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1156)

### DIFF
--- a/src/checkoutservice/main.go
+++ b/src/checkoutservice/main.go
@@ -316,6 +316,7 @@ func (cs *checkoutService) PlaceOrder(ctx context.Context, req *pb.PlaceOrderReq
 
 	// send to kafka only if kafka broker address is set
 	if cs.kafkaBrokerSvcAddr != "" {
+		log.Infof("sending to postProcessor")
 		cs.sendToPostProcessor(ctx, orderResult)
 	}
 
@@ -439,7 +440,7 @@ func (cs *checkoutService) convertCurrency(ctx context.Context, from *pb.Money, 
 
 func (cs *checkoutService) chargeCard(ctx context.Context, amount *pb.Money, paymentInfo *pb.CreditCardInfo) (string, error) {
     paymentService := cs.paymentSvcClient
-	if cs.checkPaymentFailure(ctx) {
+	if cs.isFeatureFlagEnabled(ctx, "paymentServiceUnreachable") {
         badAddress := "badAddress:50051"
         c := mustCreateClient(context.Background(), badAddress)
 		paymentService = pb.NewPaymentServiceClient(c)
@@ -505,6 +506,18 @@ func (cs *checkoutService) sendToPostProcessor(ctx context.Context, result *pb.O
 	cs.KafkaProducerClient.Input() <- &msg
 	successMsg := <-cs.KafkaProducerClient.Successes()
 	log.Infof("Successful to write message. offset: %v", successMsg.Offset)
+
+	if cs.isFeatureFlagEnabled(ctx, "kafakQueueProblems") {
+		log.Infof("Warning: FeatureFlag 'kafakQueueProblems' is activated, overloading queue now.")
+		messageCount := 100
+		for i := 0; i < messageCount; i++ {
+    		go func(i int) {
+    		    cs.KafkaProducerClient.Input() <- &msg
+				_ = <-cs.KafkaProducerClient.Successes()
+            }(i)
+		}
+		log.Infof("Done with #%d messages for overload simulation.", messageCount)
+	}
 }
 
 func createProducerSpan(ctx context.Context, msg *sarama.ProducerMessage) trace.Span {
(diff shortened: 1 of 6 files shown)


src/frauddetectionservice/src/main/kotlin/frauddetectionservice/main.kt
```diff
@@ -46,8 +58,27 @@ fun main() {
...
+* Retrieves the status of a feature flag from the Feature Flag service.
+*
+* @param ff The name of the feature flag to retrieve.
+* @return `true` if the feature flag is enabled, `false` otherwise or in case of errors.
+*/
+fun getFeatureFlagValue(ff: String): Int {
+    val client = OpenFeatureAPI.getInstance().client
+    // TODO: Plumb the actual session ID from the frontend via baggage?
+    val uuid = UUID.randomUUID()
+    client.evaluationContext = MutableContext().add("session", uuid.toString())
```
beeme1mr, line 81: I recommend using `ImmutableContext` and setting the session id in the constructor. That will be mapped to the `targetKey` property in the evaluation context, which is used in the [pseudo-random bucketing algorithm](https://flagd.dev/reference/custom-operations/fractional-operation/).

src/checkoutservice/main.go
```diff
@@ -533,11 +546,32 @@ func createProducerSpan(ctx context.Context, msg *sarama.ProducerMessage) trace.
...
 
-func (cs *checkoutService) checkPaymentFailure(ctx context.Context) bool {
-	openfeature.AddHooks(otelhooks.NewTracesHook())
-	client := openfeature.NewClient("checkout")
-	failureEnabled, _ := client.BooleanValue(
-		ctx, "paymentServiceUnreachable", false, openfeature.EvaluationContext{},
-	)
-	return failureEnabled
+func (cs *checkoutService) isFeatureFlagEnabled(ctx context.Context, featureFlagName string) bool {
+    openfeature.AddHooks(otelhooks.NewTracesHook())
```
beeme1mr, line 550: Please make sure this hook isn't already registered elsewhere in the app.

EislM0203: done

src/frauddetectionservice/src/main/kotlin/frauddetectionservice/main.kt
EislM0203: done

src/frauddetectionservice/src/main/kotlin/frauddetectionservice/main.kt
```diff
@@ -46,8 +59,30 @@ fun main() {
                     val newCount = accumulator + 1
                     val orders = OrderResult.parseFrom(record.value())
                     logger.info("Consumed record with orderId: ${orders.orderId}, and updated total count to: $newCount")
+                    if (getFeatureFlagValue("kafkaQueueProblems") > 0) {
```
puckpuck, line 62: My Kotlin isn't that great, but this seems to be checking the feature flag after the record is consumed from Kafka. I think we want it before so internal Kafka and message metrics indicate a lag in consuming the message.

EislM0203: > My Kotlin isn't that great, but this seems to be checking the feature flag after the record is consumed from Kafka. I think we want it before so internal Kafka and message metrics indicate a lag in consuming the message.

You are correct that it is sleeping after the record is consumed (i can change this if you would like to have the sleep before the claim). However, in practice we are still producing a lag since the consumer is busy with the sleep after having consumed the message. This results in the consumer only being able to claim a record approx. every second.

puckpuck: I would still do this before. There are span attributes we automatically attach for show consumer lag time which I would hope will show an increase when this flag is enabled.

EislM0203: Alright, will move the sleep up!

puckpuck (approved):

austinlparker (approved):

src/frauddetectionservice/src/main/kotlin/frauddetectionservice/main.kt
```diff
@@ -44,10 +57,32 @@ fun main() {
...
+
+/**
+* Retrieves the status of a feature flag from the Feature Flag service.
+*
+* @param ff The name of the feature flag to retrieve.
+* @return `true` if the feature flag is enabled, `false` otherwise or in case of errors.
+*/
+fun getFeatureFlagValue(ff: String): Int {
+    val client = OpenFeatureAPI.getInstance().client
+    // TODO: Plumb the actual session ID from the frontend via baggage?
```
beeme1mr, line 80: I recently opened a [PR](https://github.com/open-telemetry/opentelemetry-demo/pull/1545) to improve how session IDs are handled in the app. I focused on the AdService, but the same logic could be used here, too.

EislM0203: Interesting approach, thanks for the input! Is that something you would like to have addressed in this PR or should I open a second PR regarding this issue?

EislM0203: @beeme1mr Would it be okay for you if i merged this PR and open up another for the changes you are suggesting? This PR has been open for a while and it is already approved.

PR #1538, opened 2024-04-19
Change AccountService from Go to DotNet (auto)

### LINKED ISSUE #589: Add dotnet auto instrumentation
Dotnet should have non instrumentation library based auto instrumentation examples. 

https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation

RassK:
# Changes

Fixes #589

Changes AccountingService from Go to .NET service.
Uses [OpenTelemetry .NET Automatic Instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation).

NOTE: It's a first preview of demo app that is instrumented with OpenTelemetry .NET Automatic instrumentation. Some parts are still TBD. Please review and add comments about improvements and missing parts / docs.

## Merge Requirements

essential items:


src/accountingservice/Dockerfile
```diff
@@ -1,32 +1,36 @@
...
+COPY --from=publish /app/publish .
+
+USER root
+RUN mkdir -p "/var/log/opentelemetry/dotnet"
+RUN chown app "/var/log/opentelemetry/dotnet"
+USER app
+
+ENV OTEL_TRACES_EXPORTER=otlp
+ENV OTEL_METRICS_EXPORTER=otlp
+ENV OTEL_LOGS_EXPORTER=otlp
```
Kielek, line 33: It should be default values.

src/accountingservice/Dockerfile
```diff
@@ -1,32 +1,36 @@
...
+ENV OTEL_LOG_LEVEL=debug
```
Kielek, line 34: Do we really want to enable debug log for _production_ demo env?

RassK:
Example Jaeger traces:
image
image

CC: @lachmatt @Kielek Should parentless Kafka consume requests be filtered? Seems without additional context these just trash the output.

RassK:
After removing EOF

image

puckpuck:
@RassK Can you add a Chanlog entry for this? Also, we will need to update docs and the Helm chart for this. If you need help to prep a PR in those areas, please let us know.

RassK:
@puckpuck thanks for helping to guide the last steps missing. I have updated the changelog. Would be glad to accept some help with other missing parts as well. It's my first PR to this repo.

Kielek:
LGTM, I think this can be merged as the PR to opentlemetry.io is ready.
As I understand, helm charts should not be affected. We are changing just one type of service to another.

puckpuck:
We will need to bump up the memory for the service in the Helm chart, but nothing beyond that. We can merge this before having to do anything in Helm, since Helm is bound to a release and not nightlies.

I want to keep the label on this PR so we can track all items that need to be updated when doing our next release.

puckpuck:
When I run this branch, I notice clock skews with the accounting service, which causes issues with the trace waterfall. I'm not sure what this service does differently than our other .NET service (cart), but we don't see the same clock skew there. @RassK can you take a look at this?

RassK:
@lachmatt could you recheck that kafka instrumentation, I think that is the source of this issue.

RassK:
@puckpuck So consulted with @lachmatt and we reached conclusion that it is not the issue with clock skew. 

In the .NET the Kafka span's instrumentation start time is the start of the polling. Previously we had issues with the amount of spans, so every poll (even an empty one) created a span. Now as we by default filter such spans, I could lower the polling max timeout to default (100ms), so accounting service's Kafka consume spans could be now <100ms, which can be still huge compared to some other services reporting processing time around 1ms. The default setting now renders the Jaeger UI in more understandable way... but we cannot make the UI look more beautiful just because of the messaging behaviour, it still has to reflect the actual logic. 

I see it more like Jaeger could improve theirs's UI by using more dynamic time axle and collapse big portions like this (10s wait time).

Here is the improved look >
image

julianocosta89:
@RassK wouldn't that be better handled in a linked span, instead of a child span?

We did that for `fraudDetectionService`.

RassK:
@julianocosta89 that's unfortunately not configurable in auto instrumentation.

puckpuck:
Something is wrong with the .NET SDK here. No other SDKs start the span when you start polling.

lachmatt:
@puckpuck 
I don't think this is related to .NET SDK, but rather Kafka client instrumentation library which is part of OTel .NET Autoinstrumentation.
Semantic conventions for messaging recommend to create `receive` spans for [pull-based scenarios](https://github.com/open-telemetry/semantic-conventions/blob/507daf4bb9b9abe615bd641bbad248d810791177/docs/messaging/messaging-spans.md#consumer-spans).
Most of the existing instrumentations for Kafka client are creating `process` spans, but these were problematic, as explained in [OTEP](https://github.com/open-telemetry/oteps/blob/dc619dfc70f174ef31caf90f14e8b00600da4049/text/trace/0220-messaging-semantic-conventions-span-structure.md#consumer), and in [PR](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3055#discussion_r1395787430).

puckpuck:
We should investigate doing this so the trace doesn't look like accounting service is starting before everything else, which looks like this should be done with a process consumer span. In its current state, this makes the consumption of the PlaceOrder trace very hard to do and will only raise questions from users about why the accounting service started before the request itself.

lachmatt:
> @RassK wouldn't that be better handled in a linked span, instead of a child span?

Creation context is used as a parent, in addition to being added as a link. This scenario is described in [OTEP](https://github.com/open-telemetry/oteps/blob/dc619dfc70f174ef31caf90f14e8b00600da4049/text/trace/0220-messaging-semantic-conventions-span-structure.md?plain=1#L200) and was discussed when working on the [instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3055#discussion_r1399074286).

> We should investigate doing this so the trace doesn't look like accounting service is starting before everything else, which looks like this should be done with a process consumer span. In its current state, this makes the consumption of the PlaceOrder trace very hard to do and will only raise questions from users about why the accounting service started before the request itself.

@pyohannes What would be the recommendation from Messaging SIG side in such scenario?

pyohannes:
> Creation context is used as a parent, in addition to being added as a link. This scenario is described in [OTEP](https://github.com/open-telemetry/oteps/blob/dc619dfc70f174ef31caf90f14e8b00600da4049/text/trace/0220-messaging-semantic-conventions-span-structure.md?plain=1#L200) and was discussed when working on the https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/3055#discussion_r1399074286.

This part is correct: you can use the creation context both as parent and link in "Receive" or "Process" operations.

The problem lies more in the duration of the "Receive" operation, which also includes the time spent in polling (which technically isn't part of this trace). I see you achieve this by starting a span only after a message was finally received, but by explicitly setting the start time of the span to the start of the polling operation.

I'll put this on the agenda for discussion in this week's messaging SIG meeting. While what is happening here doesn't really violate the semantic conventions, I think the "Receive" duration should only start _after_ a message was received from the broker. This would result in more meaningful traces.

I submitted https://github.com/open-telemetry/semantic-conventions/issues/1050 to track related discussions.

pyohannes:
This was discussed in the messaging workgroup:
* The duration of "Receive" spans should include the time spent for polling.
* "Receive" spans by default should link to the creation context, but shouldn't use it as parent.

See a more detailed summary [here](https://github.com/open-telemetry/semantic-conventions/issues/1050#issuecomment-2133128833).

RassK:
I'll update the solution when a newer version of otel auto is available.

julianocosta89:
@RassK should we keep this PR as a Draft then?

RassK:
> @RassK should we keep this PR as a Draft then?

Sure, I have no preferences there.

RassK:
Updated solution (OTel Auto 1.7.0) >
image

CHANGELOG.md
```diff
@@ -55,6 +55,9 @@ the release.
   ([#1594](https://github.com/open-telemetry/opentelemetry-demo/pull/1594))
 * [checkout] increase Kafka resiliency and observability
   ([#1590](https://github.com/open-telemetry/opentelemetry-demo/pull/1590))
+* [accountingservice] convert from Go service to .NET service, uses
```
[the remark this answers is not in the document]
RassK: https://github.com/open-telemetry/opentelemetry-demo/pull/1538/commits/d02a6c1869c4d998990443557e0f818403b16a50

julianocosta89 (commented):
Hey @RassK it looks good, before approving would you mind explaining those log entries?

```
Accounting service started
[CORECLR_ENABLE_PROFILING, 1]
[CORECLR_PROFILER, {918728DD-259F-4A6A-AC2B-B85E1B658318}]
[CORECLR_PROFILER_PATH, /app/OpenTelemetry.AutoInstrumentation.Native.so]
[DOTNET_RUNNING_IN_CONTAINER, true]
[DOTNET_STARTUP_HOOKS, /app/OpenTelemetry.AutoInstrumentation.StartupHook.dll]
```

RassK:
> Hey @RassK it looks good, before approving would you mind explaining those log entries?
> 
> ```
> Accounting service started
> [CORECLR_ENABLE_PROFILING, 1]
> [CORECLR_PROFILER, {918728DD-259F-4A6A-AC2B-B85E1B658318}]
> [CORECLR_PROFILER_PATH, /app/OpenTelemetry.AutoInstrumentation.Native.so]
> [DOTNET_RUNNING_IN_CONTAINER, true]
> [DOTNET_STARTUP_HOOKS, /app/OpenTelemetry.AutoInstrumentation.StartupHook.dll]
> ```

That's a quick startup diagnostics (by `AccountingService.Helpers` that filters relevant environment variables). Due automatic instrumentation is configured mostly by envs, then this could be helpful for a quick diagnosis.

Ideally a built in diagnostics (OTEL_LOG_LEVEL=debug) should do the similar but instead logs to the file.

Kielek (approved):

julianocosta89 (approved):

julianocosta89:
I've updated the tracetest to reflect the change, hope you don't mind. 😅 

This should be good to go, I'll just wait to see if any other maintainer/approver has something to say.

Thanks for taking care of it 🤩

RassK:
Just need to increase mem limit in helm chart left? Can anyone help / explain where, I have had 0 connection to helm charts so far 😬

julianocosta89:
the Helm chart will not be updated till next version release.

If you want to update it, you can do the memory limit here:
https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-demo/values.yaml#L165

Then you need to bump a minor version here: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-demo/Chart.yaml#L4

And last you need to run the: `make generate-examples CHARTS=opentelemetry-demo`.

You could also sync with @rogercoll and add it to his opened PR:
https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1232

Whenever you add it, you will still need to run the `make generate` command above.

puckpuck:
I tested again, and all looks fine using a linked span here instead. I will make another PR so the frauddetectionservice goes back to creating a child span instead of a linked span. This will allow us to showcase both examples with a message queue.

puckpuck (approved):

PR #1551, opened 2024-04-30
[chore] specify default value for tracetest image version

### LINKED ISSUE #1550: tracetest workflow is failing
The build and push images action we use doesn't specify build args in the matrix. If no build arg is specified the image build will download and use the latest CLI, which was recently released as 1.0.0. This CLI is not compatible with the 0.16.0 tracetest-server version.

puckpuck:
Fixes #1550 

Since the tracetest server and CLI version need to match, we must ensure both images are aligned. The solution is to hardcode the tracetest version in the GitHub workflow or the Dockerfile. I opted to specify this in the Dockerfile, adding a comment on why we need to do it there.

### DIFF
--- a/.env
+++ b/.env
@@ -14,8 +14,7 @@ OPENSEARCH_IMAGE=opensearchproject/opensearch:2.12.0
 POSTGRES_IMAGE=postgres:16.2
 PROMETHEUS_IMAGE=quay.io/prometheus/prometheus:v2.51.1
 REDIS_IMAGE=redis:7.2-alpine
-TRACETEST_IMAGE_VERSION=v0.16.0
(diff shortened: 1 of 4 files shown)


julianocosta89:
Any reason why you removed the Argos?

With this, whenever we update the TraceTest we will need to update it in 2 places.

austinlparker (approved):

puckpuck:
> Any reason why you removed the Args?
> 
> With this, whenever we update the TraceTest we will need to update it in 2 places.

For build-args to work with our published images the GitHub action workflow would require us to hardcode the version value there. So no matter what we do we need to update the version in 2 places. I opted to use the route where we update the Dockerfile instead of a GitHub action workflow yml.

julianocosta89 (approved):

PR #1569, opened 2024-05-10
Use `solarwinds-apm` for JS

raphael-theriault-swi:
The browser tracer was left as-is for now. I also couldn't use the Next instrumentation hook in the end cause the demo uses an outdated version that doesn't have it.


raphael-theriault-swi:
Wrong repo my bad

PR #1588, opened 2024-06-01
[chore] remove link failing CI (Coralogix)

puckpuck:
# Changes

This removes the link for Coralogix but keeps the text as a vendor supporting the demo. This is to resolve the CI error where the markdown-link-checker running in our GitHub Actions CI fails when checking that link. This is a temporary solution so Coralogix can update the link or fix their web hosting configuration to ensure the CI failure no longer happens.

/cc @oded-dd who did the original PR to add Coralogix


puckpuck:
It seems that Coralogix's hosting provider no longer has an issue with the markdown link checker. I am closing this as it is no longer needed.

PR #1590, opened 2024-06-01
[checkout] fix kafka restart

puckpuck:
# Changes

If Kafka is restarted, the checkout service will start to report very long response/acknowledgment times from Kafka. This may be due to the demo's unique configuration, which only has a single broker. To get around the issue, the Kafka/Sarama SDK is set to respond without full acknowledgments. This may lead to missed errors in certain situations but will allow the demo to recover from Kafka restarts and continue working without issues.

Part of this PR is to do proper handling of the Kafka message in a success or error scenario and also add attributes to the Kafka producer span related to details about the interaction with Kafka. A docs update will be required.

## Merge Requirements

essential items:

~* [ ] Appropriate Helm chart updates in the [helm-charts][]~


src/checkoutservice/main.go
```diff
@@ -504,18 +505,52 @@ func (cs *checkoutService) sendToPostProcessor(ctx context.Context, result *pb.O
...
-	log.Infof("Successful to write message. offset: %v", successMsg.Offset)
+	// Send message and handle response
+	startTime := time.Now()
+	select {
+	case cs.KafkaProducerClient.Input() <- &msg:
+		log.Infof("Message sent to Kafka: %v", msg)
+		select {
+		case successMsg := <-cs.KafkaProducerClient.Successes():
+			span.SetAttributes(
+				attribute.Bool("kafka.producer.success", true),
```
julianocosta89, line 516: 2 questions on all those span attributes:
- Any reason why we are adding all of them?
- Any chance we have those attributes in the semantic conventions?

julianocosta89: I'm not opposed to keeping the attributes, but I'd love if we could follow the semantic conventions on them

src/checkoutservice/main.go
```diff
@@ -504,18 +505,52 @@ func (cs *checkoutService) sendToPostProcessor(ctx context.Context, result *pb.O
...
+				attribute.Int64("kafka.offset", successMsg.Offset),
```
julianocosta89, line 517: Found this one:

```suggestion
				attribute.KeyValue(semconv.MessagingKafkaMessageOffset(int(successMsg.Offset))),
```

puckpuck:
@julianocosta89 I moved to use semantic conventions or attribute names that align closer to what already exists. Let me know if we still need to update them.

julianocosta89 (approved):
LGTM, thanks @puckpuck!

PR #1600, opened 2024-06-11
chore: parametrize services Dockerfile path

rogercoll:
# Changes

Define services paths and Dockerfiles using and environment variable so it can be easily overridden. With the added variables, a vendor can easily point to another Dockerfile using the `.env.override` file:

```
PAYMENT_SERVICE_DOCKERFILE=./src/paymentservice/Dockerfile.custom
```

That would help to make changes additive, thus minimizing merge conflicts. 
In addition, the demo can also benefit from it and add different Dockerfile for the services and reflect different Otel deployment strategies. 

The PR also aligns the `emailservice` Dockerfile with the other's services strategy by expecting to be built from the root's repository path.

## Merge Requirements

essential items:

~~* [ ] `CHANGELOG.md` updated to document new feature additions~~
~~* [ ] Appropriate documentation updates in the [docs][]~~
~~* [ ] Appropriate Helm chart updates in the [helm-charts][]~~


julianocosta89:
Hey @rogercoll 👋🏽 

This is interesting, and I like where this is going.
One note thought, would you be able to change the env vars to contain the whole path?

That would avoid the concatenation in the compose file and it would help when things do not go as expected.
WDYT?

rogercoll:
>  One note thought, would you be able to change the env vars to contain the whole path?

Definitely, it will be even more helpful for vendors that want to completely move a service to a different directory. Something like https://github.com/open-telemetry/opentelemetry-demo/pull/1600/commits/8a40cd048891c88d14bd999ff0a984ea782aa524 ?

PS I will bring this PR/idea to today's SIG meeting

src/emailservice/Dockerfile
```diff
@@ -8,15 +8,15 @@ FROM base as builder
 
 WORKDIR /tmp
 
-COPY Gemfile Gemfile.lock .
+COPY ./src/emailservice/Gemfile ./src/emailservice/Gemfile.lock ./
```
rogercoll, line 11: Use `Dockerfile` from root's demo directory instead (like the other services approach)

julianocosta89:
@rogercoll, with 2 env vars to do the solution you mention we would need to update 2 places:

```
PAYMENT_SERVICE_DOCKERFILE=./src/paymentservice/Dockerfile.custom
```

My initial thought was having 1 env var containing the full path, do you see other use cases where having 2 would make more sense?

puckpuck:
I would like to use a single variable for each service containing the path and name of the Dockerfile. If someone needs to update this, they only need to update a single spot to reflect either a new folder or a new Dockerfile name.

puckpuck (approved):
Thanks for this.

PR #1613, opened 2024-06-18
Use native runners

austinlparker:
This PR should target native runners (amd64, arm64) for builds.


puckpuck (approved):

puckpuck:
Let's get this merged and manually kick off a `Build and Publish` workflow so we can validate the built images are correct.

Unfortunately there is no real other way to test GitHub actions.

Issue #1627, opened 2024-06-23
Grafana doesn't work without an Internet connection


flands:
# Bug Report

Which version of the demo you are using? 1.10.0

## Symptom

If you attempt to start the demo application without an Internet connection (containers already pulled down), then Grafana is not access and restarts every minute

**What is the expected behavior?**

If you have the demo environment containers locally, then the demo environment should work without an Internet connection.

**What is the actual behavior?**

The demo environment requires an Internet connection for Grafana to work even if container are already pulled down.

## Reproduce

Pull version 1.10.0 and run `make start-minimal` with an Internet connection. Once running, run `make stop`. Disable your Internet. Run `make start-minmal`. Once running, attempt to access Grafana.

## Additional Context

If you look at the logs for Grafana without an Internet connection you will see: `Error: ✗ Get "https://grafana.com/api/plugins/grafana-opensearch-datasource/versions": dial tcp: lookup grafana.com on 127.0.0.11:53: no such host`

I see several potential issues. The root cause appears to be because `src/grafana/provisioning/datasources/opensearch.yaml` has `access: proxy` set. If you remove this, the issue is resolved. I can make this change if everyone is aligned.

Beyond this, it appears `src/grafana/grafana.ini` has `check_for_updates = true` and `check_for_plugin_updates = true`. I'm not sure whether this will result in Grafana being unavailable, but also seems unnecessary for the demo environments. I can make this change if everyone is aligned.

julianocosta89:
I'm good with the suggestions.
I actually faced the same during my flight this week 😅
I wanted to run the demo to test the feature flags issue you reported and I saw that Grafana wasn't starting.

flands:
Turns out this is a bit more complicated - see https://github.com/open-telemetry/opentelemetry-demo/pull/1633#issuecomment-2198552666.

SightStudio:
is this solved? 

I still got the same message.

```
Error: ✗ Get "https://grafana.com/api/plugins/grafana-opensearch-datasource/versions
```

using under version

````
helm search repo telemetry/opentelemetry-demo --versions

NAME                                    CHART VERSION   APP VERSION     DESCRIPTION 
open-telemetry/opentelemetry-demo       0.33.4          1.12.0          opentelemetry demo helm chart
```

puckpuck:
[...]

PR #1633, opened 2024-06-26
Allow Grafana to start without an Internet connection

flands:
# Changes

Addresses #1627

## Merge Requirements

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ the release.
   ([#1610](https://github.com/open-telemetry/opentelemetry-demo/pull/1610))
 * [Valkey] Replace Redis with Valkey
   ([#1619](https://github.com/open-telemetry/opentelemetry-demo/pull/1619))
+* [grafana] Allow Grafana to run without an Internet connection
+  ([#1633](https://github.com/open-telemetry/opentelemetry-demo/pull/1633))
 
 ## 1.10.0
 
--- a/src/grafana/grafana.ini
+++ b/src/grafana/grafana.ini
@@ -208,14 +208,14 @@ serve_from_sub_path = true
 # in some UI views to notify that a grafana update exists.
 # This option does not cause any auto updates, nor send any information
 # only a GET request to https://raw.githubusercontent.com/grafana/grafana/main/latest.json to get the latest version.
-;check_for_updates = true
+check_for_updates = false
 
 # Set to false to disable all checks to https://grafana.com
 # for new versions of plugins. The check is used
 # in some UI views to notify that a plugin update exists.
 # This option does not cause any auto updates, nor send any information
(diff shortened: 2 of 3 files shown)


puckpuck:
With these changes, and disconnected to the Internet, I still get the `no healthy upstream` error when going to http://localhost/grafana/. The grafana logs show this message: 

```
Error: ✗ Get "https://grafana.com/api/plugins/grafana-opensearch-datasource/versions": dial tcp: lookup grafana.com on 127.0.0.11:53: no such host
```

puckpuck:
I also ask if this is functionality that we should target to support since several of the images needed to run the demo, including Grafana itself, are not even built locally.

cedricziel (approved):

julianocosta89 (commented):
I'm not sure why, but I still get an error when offline:

```
Error: ✗ Get "https://grafana.com/api/plugins/grafana-opensearch-datasource/versions": dial tcp: lookup grafana.com on 127.0.0.11:53: no such host
```

puckpuck:
I believe the issue here is that the OpenSearch plugin (and perhaps others) is not part of Grafana's core distribution and image. When Grafana starts up, it will download the latest version of the plugin and enable it.

flands:
It appears the proposed changes will not work. If you start with an Internet connection, you will see the following Grafana logs:

```
✔ Downloaded and extracted grafana-opensearch-datasource v2.16.1 zip successfully to /var/lib/grafana/plugins/grafana-opensearch-datasource
logger=plugins.registration t=2024-06-30T12:45:07.602529512Z level=info msg="Plugin registered" pluginId=grafana-opensearch-datasource
logger=provisioning.datasources t=2024-06-30T12:45:07.681012346Z level=info msg="inserting datasource from configuration" name=OpenSearch uid=P9744FCCEAAFBD98F
```

It seems like one approach to solve this would be to bundle the plugin with the demo releases and mount it on start. Alternatively, state would need to be persisted, at least for Grafana, across restarts. Closing this.

GeorgeGkinis:
I changed the opensearch configuration to use the bundled elasticsearch datasource in `src/grafana/provisioning/datasources/opensearch.yaml`:

```
# Copyright The OpenTelemetry Authors

apiVersion: 1

datasources:
  - name: OpenSearch
    type: elasticsearch
    url: http://opensearch:9200/
    access: proxy
    editable: true
    isDefault: false
    jsonData:
      database: otel
      flavor: opensearch
      logLevelField: severity
      logMessageField: body
      pplEnabled: true
      timeField: observedTimestamp
      version: 2.17.1

```
I also had to comment this out in `docker-compose.yaml`:
```
    #environment:
      #- "GF_INSTALL_PLUGINS=grafana-opensearch-datasource"
```

grafana starts up and seems to work so far.
I see data in the dashboards.

PR #1654, opened 2024-07-02
frauddetection uses child span for messaging

puckpuck:
# Changes

With #1538, the accountingservice moved from a child span to a linked span from the Kafka queue. We no longer had any child spans from Kafka. This change will set the fraud detection service to generate child spans from Kafka, allowing the demo to showcase both use cases from messaging queues.

## Merge Requirements

essential items:

~* [ ] Appropriate documentation updates in the [docs][]~
~* [ ] Appropriate Helm chart updates in the [helm-charts][]~


julianocosta89:
Not sure if we should add this.
AFAIK the recommended way to do it is via linked spans, as we are doing now.

https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#topic-with-multiple-consumers

puckpuck:
I don't see the example listed there, but I believe a child span should be created for single producer/consumer scenarios. I understand we don't have that scenario here at all, but being able to demonstrate trace propagation using a child span is still something I think we should showcase.

julianocosta89:
All examples in the page use `link`.
The recommendation by the Messaging SIG is using link: https://github.com/open-telemetry/opentelemetry-demo/pull/1538#issuecomment-2133132629

PR #1655, opened 2024-07-02
add honey web tracing

maxaguirre5:
# Changes

Adding Honeycomb tracing file, modifying the frontend service a tad to use hny web, other minor changes to match the old ca-otel-demo


maxaguirre5:
oof.png silly opening pr against the wrong thing

Issue #1653, opened 2024-07-02
Add Infra metrics and dashboards


puckpuck:
# Feature Request

With Docker stats added to the demo in #1650, we need to ensure that we add the rest of the functionality to get infra metrics working and consumable for the user.

- [ ] Grafana dashboard with docker stats by service
- [ ] Equivalent metrics integration for Kubernetes / Helm deployment
- [ ] Grafana dashboard for kubelet stats by service
- [ ] Optimization of Prometheus deployment with increased metrics load

ref: #1624

PR #1675, opened 2024-07-17
[otel-col] Add host metrics receiver

rogercoll:
# Changes

This PR adds the hostmetrics receiver to the demo.

Related issue: https://github.com/open-telemetry/opentelemetry-demo/issues/1653

## Merge Requirements

essential items:

~~* [ ] Appropriate documentation updates in the [docs][]~~
~~* [ ] Appropriate Helm chart updates in the [helm-charts][]~~

### DIFF
--- a/src/otelcollector/otelcol-config.yml
+++ b/src/otelcollector/otelcol-config.yml
@@ -69,6 +69,9 @@ receivers:
             - tracefs
           match_type: strict
       memory:
+        metrics:
+          system.memory.utilization:
+            enabled: true
       network:
       paging:
       processes:
--- a/src/otelcollector/otelcol-config.yml
+++ b/src/otelcollector/otelcol-config.yml
@@ -26,6 +26,9 @@ receivers:
     root_path: /hostfs
     scrapers:
       cpu:
+        metrics:
+          system.cpu.utilization:
+            enabled: true
       disk:
       load:
       filesystem:
--- a/src/otelcollector/otelcol-config.yml
+++ b/src/otelcollector/otelcol-config.yml
@@ -21,6 +21,58 @@ receivers:
     endpoint: "valkey-cart:6379"
     username: "valkey"
     collection_interval: 10s
+  # Host metrics
+  hostmetrics:
+    root_path: /hostfs
+    scrapers:
+      cpu:
+      disk:
+      load:
+      filesystem:
+        exclude_mount_points:
+          mount_points:
+            - /dev/*
+            - /proc/*
+            - /sys/*
+            - /run/k3s/containerd/*
+            - /var/lib/docker/*
+            - /var/lib/kubelet/*
+            - /snap/*
+          match_type: regexp
+        exclude_fs_types:
+          fs_types:
+            - autofs
+            - binfmt_misc
+            - bpf
+            - cgroup2
+            - configfs
+            - debugfs
+            - devpts
+            - devtmpfs
+            - fusectl
+            - hugetlbfs
+            - iso9660
+            - mqueue
+            - nsfs
+            - overlay
+            - proc
+            - procfs
+            - pstore
+            - rpc_pipefs
+            - securityfs
+            - selinuxfs
+            - squashfs
+            - sysfs
+            - tracefs
+          match_type: strict
+      memory:
+      network:
+      paging:
+      processes:
+      process:
+        mute_process_exe_error: true
+        mute_process_io_error: true
+        mute_process_user_error: true
   # Collector metrics
   prometheus:
     config:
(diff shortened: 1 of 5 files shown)


src/otelcollector/otelcol-config.yml
```diff
@@ -21,6 +21,58 @@ receivers:
...
     username: "valkey"
     collection_interval: 10s
+  # Host metrics
+  hostmetrics:
+    root_path: /hostfs
+    scrapers:
+      cpu:
+      disk:
+      load:
+      filesystem:
```
rogercoll, line 31: Same filesystem configuration as upstream helm chart: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/_config.tpl#L99

src/otelcollector/otelcol-config.yml
```diff
@@ -21,6 +21,58 @@ receivers:
...
+            - tracefs
+          match_type: strict
+      memory:
+      network:
+      paging:
+      processes:
+      process:
+        mute_process_exe_error: true
+        mute_process_io_error: true
+        mute_process_user_error: true
```
rogercoll, line 75: Container is not mounted as privileged, thus it cannot access some root's owned files. These configuration options mute those "errors". 


see: https://github.com/bsponge/opentelemetry-collector-contrib/blob/6ef98e9c2a6bfe267c6e821361da891bd28ea4ae/receiver/hostmetricsreceiver/README.md#process

rogercoll:
Draft PR to enable them in Helm: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1261

src/otelcollector/otelcol-config.yml
```diff
@@ -21,6 +21,58 @@ receivers:
     endpoint: "valkey-cart:6379"
     username: "valkey"
     collection_interval: 10s
+  # Host metrics
+  hostmetrics:
+    root_path: /hostfs
+    scrapers:
```
julianocosta89, line 27: Should we add some optional metrics?

src/otelcollector/otelcol-config.yml
```diff
@@ -21,6 +21,58 @@ receivers:
...
+      cpu:
```
julianocosta89, line 28:
```suggestion
      cpu:
        metrics:
          system.cpu.utilization:
            enabled: true
```

src/otelcollector/otelcol-config.yml
julianocosta89, line 31:
```suggestion
      filesystem:
        metrics:
          system.filesystem.utilization:
            enabled: true
```

src/otelcollector/otelcol-config.yml
```diff
@@ -21,6 +21,58 @@ receivers:
...
+            - rpc_pipefs
+            - securityfs
+            - selinuxfs
+            - squashfs
+            - sysfs
+            - tracefs
+          match_type: strict
+      memory:
+      network:
+      paging:
```
julianocosta89, line 70:
```suggestion
      paging:
        metrics:
          system.paging.utilization:
            enabled: true
```

src/otelcollector/otelcol-config.yml
```diff
@@ -21,6 +21,58 @@ receivers:
...
+            - procfs
+            - pstore
+            - rpc_pipefs
+            - securityfs
+            - selinuxfs
+            - squashfs
+            - sysfs
+            - tracefs
+          match_type: strict
+      memory:
```
julianocosta89, line 68:
```suggestion
      memory:
        metrics:
          system.memory.limit:
            enabled: true
          system.memory.utilization:
            enabled: true
```

src/otelcollector/otelcol-config.yml
```diff
@@ -21,6 +21,58 @@ receivers:
...
+      network:
+      paging:
+      processes:
+      process:
```
julianocosta89, line 72:
```suggestion
      process:
        metrics:
          process.cpu.utilization:
            enabled: true
          process.memory.utilization:
            enabled: true
```

julianocosta89 (approved):

julianocosta89:
@rogercoll LGTM, I've added some suggestions, but definitely not a blocker

src/otelcollector/otelcol-config.yml
rogercoll: That's a good point. I am not sure about this change, I firstly enabled all optional metrics, but then I thought it would make sense to have the same configuration as in K8s. The upstream Helm chart does not enable any of the optional metrics, thus we would need to provide a custom hosmetrics receiver configuration in the opentelemetry-demo chart: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/templates/_config.tpl#L103

Given this context, would you still prefer to enable those optional metrics? (I don't have strong opinion)

julianocosta89: Hmmm, I think the main question here is if those metrics are important or not.
I'd say that at least the following 2 are:
- system.cpu.utilization
- system.memory.utilization

But I also don't have a strong opinion, this was more like a suggestion than a blocker on the PR.

rogercoll: Sounds good, let's start with `system.cpu.utilization` and `system.memory.utilization` for now.

PR #1690, opened 2024-07-30
[shippingservice] set rpc.method and rpc.service

mottibec:
# Changes
Added `rpc.method` and `rpc.service` to shippingservice grpc spans


puckpuck (approved):
LGTM

Issue #1692, opened 2024-07-30
accountingservice container image fails as CreateContainerError when not running as root


iblancasa:
When the `accountingservice` container is run as a non-root user, an error is triggered by Kubernetes.

`kubectl describe pod <pod_name>` shows this:
```sh
Error: container create failed: time="2024-07-30T18:39:52Z" level=error msg="runc create failed: unable to start container process: exec: \"./instrument.sh\": permission denied"
```

I checked in the container and the script has root permissions.

PR #1696, opened 2024-07-31
Fix accountingservice container for non-root users

iblancasa:
# Changes

Currently, when you are using an user different than root to run the container, it fails because it has not permissions to run `instrumentation.sh`.
Fixes: #1692

## Merge Requirements

essential items:

### DIFF
--- a/src/accountingservice/Dockerfile
+++ b/src/accountingservice/Dockerfile
@@ -26,6 +26,7 @@ COPY --from=publish /app/publish .
 USER root
 RUN mkdir -p "/var/log/opentelemetry/dotnet"
 RUN chown app "/var/log/opentelemetry/dotnet"
+RUN chmod 755 "/app/instrument.sh"
 USER app
 
 ENTRYPOINT ["./instrument.sh", "dotnet", "AccountingService.dll"]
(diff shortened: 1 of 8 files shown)


src/accountingservice/Dockerfile
```diff
@@ -26,6 +26,7 @@ COPY --from=publish /app/publish .
 USER root
```
julianocosta89, line 26: would it make sense to change the user here, as now no root is required?

iblancasa: It is needed because is the user with permission to change permissions over the other files.

src/accountingservice/Dockerfile
```diff
@@ -26,6 +26,7 @@ COPY --from=publish /app/publish .
...
 RUN mkdir -p "/var/log/opentelemetry/dotnet"
 RUN chown app "/var/log/opentelemetry/dotnet"
+RUN chmod 755 "/app/instrument.sh"
```
rogercoll, line 29:
```suggestion
RUN chown app "/app/instrument.sh"
```

The `app` user seems to be the one running the script, would it work just setting it as the owner? (no execution permissions to others)

iblancasa: Let me check if this works and come back :)

iblancasa: It works. Commited

rogercoll: Thanks for the double check!

rogercoll (approved):
lgtm

rogercoll:
CI failure is unrelated to this PR (fix https://github.com/open-telemetry/opentelemetry-demo/pull/1702)

PR #1711, opened 2024-08-28
Upgrade flagd to v0.11.2

### LINKED ISSUE #1625: Changing a feature flag `defaultVariant` doesn't dynamically update services while demo is running
# Bug Report

Which version of the demo you are using? 1.10.0

## Symptom

Changing a feature flag `defaultVariant` while the demo application is running doesn't dynamically update services

Changing a feature flag `defaultVariant` to `on` while demo is running dynamically updates services

Changing a feature flag `defaultVariant` to `on` while demo is running is a noop.

## Reproduce

Per https://opentelemetry.io/docs/demo/feature-flags/, To enable a flag, change the defaultVariant value in the config file for a given flag to “on”, while the demo application is running edit `src/flagd/demo.flagd.json` and change a `defaultVariant` to `on`

## Additional Context

Perhaps this is expected behavior. If it is, then the documentation should be updated to state that if the demo application is running and a feature flag `defaultVariant` is changed then you need to restart the flagd service (`make restart service=flagd`) or restart the demo application for it to take effect.

I can make the documentation change if this is the recommended approach.

Starefossen:
# Changes

Upgrade flagd to v0.11.2 to make it reload feature toggles properly.

Close #1625

### DIFF
--- a/kubernetes/opentelemetry-demo.yaml
+++ b/kubernetes/opentelemetry-demo.yaml
@@ -9910,7 +9910,7 @@ spec:
       serviceAccountName: opentelemetry-demo
       containers:
         - name: flagd
-          image: 'ghcr.io/open-feature/flagd:v0.11.1'
+          image: 'ghcr.io/open-feature/flagd:v0.11.2'
           imagePullPolicy: IfNotPresent
           command:
           - /flagd-build
--- a/.env
+++ b/.env
@@ -7,7 +7,7 @@ DEMO_VERSION=latest
 
 # Dependent images
 COLLECTOR_CONTRIB_IMAGE=otel/opentelemetry-collector-contrib:0.105.0
-FLAGD_IMAGE=ghcr.io/open-feature/flagd:v0.11.1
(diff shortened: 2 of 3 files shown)


beeme1mr (approved):

rogercoll (approved):

kubernetes/opentelemetry-demo.yaml
```diff
@@ -9910,7 +9910,7 @@ spec:
       serviceAccountName: opentelemetry-demo
       containers:
         - name: flagd
-          image: 'ghcr.io/open-feature/flagd:v0.11.1'
+          image: 'ghcr.io/open-feature/flagd:v0.11.2'
```
julianocosta89, line 9913: Would you mind reverting this change here?

This file should be updated only with:

```
make generate-kubernetes-manifests
```

We usually run it when there is a new release and helm update.
https://github.com/open-telemetry/opentelemetry-demo/blob/main/Makefile#L114-L126

julianocosta89: @Starefossen would you be able to revert this one so we can get this merged?

PR #1714, opened 2024-09-02
Currency service

randax:
instrumentation assignment


julianocosta89:
Hey @randax, it's nice to see that you have built a workshop with the demo ⭐ !
I quickly had a look over it, just out of curiosity and noticed that you are using the old version of docker compose.

In the instructions you are mentioning `docker-compose`, the new version doesn't use the `-`.
To make sure all the environment variables are properly populated I'd suggest using the latest version of docker compose.
The sintax is basically the same, just remove the `-` and you are good to go:

```
docker compose [COMMAND]
```

PR #1715, opened 2024-09-06
[chore] Move dependencies control to .env file

julianocosta89:
# Changes

A couple of dependencies had to be updated in multiple places, this PR moves all of them to `.env` file.

## Merge Requirements

essential items:

### DIFF
--- a/.env
+++ b/.env
@@ -10,13 +10,16 @@ COLLECTOR_CONTRIB_IMAGE=otel/opentelemetry-collector-contrib:0.108.0
 FLAGD_IMAGE=ghcr.io/open-feature/flagd:v0.11.2
 GRAFANA_IMAGE=grafana/grafana:11.2.0
 JAEGERTRACING_IMAGE=jaegertracing/all-in-one:1.60
-# must also update version field in ./src/grafana/provisioning/datasources/opensearch.yaml
-OPENSEARCH_IMAGE=opensearchproject/opensearch:2.16.0
+OPENSEARCH_IMAGE_VERSION=2.16.0
+OPENSEARCH_IMAGE=opensearchproject/opensearch:${OPENSEARCH_IMAGE_VERSION}
 POSTGRES_IMAGE=postgres:16.4
 PROMETHEUS_IMAGE=quay.io/prometheus/prometheus:v2.54.1
-VALKEY_IMAGE=valkey/valkey:8.0-alpine
-# must also update the version arg in ./test/tracetesting/Dockerfile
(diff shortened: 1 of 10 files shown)


julianocosta89:
`build_images` is not using docker compose 😞

I'm open to suggestions, because I strongly believe we should have a centralised place to manage the versions.

As of today we have to update 3 files every time the java agent updates.
2 places every time there is an update to opensearch or tracetest.

.env
```diff
@@ -10,13 +10,16 @@ COLLECTOR_CONTRIB_IMAGE=otel/opentelemetry-collector-contrib:0.108.0
...
+OPENSEARCH_IMAGE_VERSION=2.16.0
+OPENSEARCH_IMAGE=opensearchproject/opensearch:${OPENSEARCH_IMAGE_VERSION}
 POSTGRES_IMAGE=postgres:16.4
 PROMETHEUS_IMAGE=quay.io/prometheus/prometheus:v2.54.1
-VALKEY_IMAGE=valkey/valkey:8.0-alpine
-# must also update the version arg in ./test/tracetesting/Dockerfile
-TRACETEST_IMAGE=kubeshop/tracetest:v1.5.2
+TRACETEST_IMAGE_VERSION=v1.5.4
+TRACETEST_IMAGE=kubeshop/tracetest:${TRACETEST_IMAGE_VERSION}
+VALKEY_IMAGE=valkey/valkey:7.2-alpine
```
rogercoll, line 19: Is there any reason behind this version change?

```suggestion
VALKEY_IMAGE=valkey/valkey:8.0-alpine
```

rogercoll (approved):
.env FTW! I like the idea of controlling as much as possible the versions from a single file

rogercoll:
> build_images is not using docker compose 😞

Could this be changed to use `docker compose build`? :thinking: I agree that relying on a single dependencies file would ease the maintenance

.env
julianocosta89: @rogercoll I've reverted the update because it seems `8.0` is `RC`.
The latest stable is `7.2`.

julianocosta89:
Unfortunately I'm not a Github actions expert.
I've googled a bit around and I'm still not sure.

We have this in the GH marketplace: https://github.com/marketplace?query=docker+compose
But only 5 of those are `verified` and none of those 5 do what we need here.

@puckpuck any ideas/suggestions?

julianocosta89:
I'm gonna close this one as I can't fix the GH actions right now.

Issue #1735, opened 2024-10-07
[featureflag] `productCatalogFailure` flag raising 128 events


julianocosta89:
# Bug Report

Which version of the demo you are using?
Main (a1cfe470c6c1e412865b351e473060b66674b2fb)

## Symptom

When enabling the feature flag `productCatalogFailure`, the spans from productCatalog service are emitting 128 events.
![image](https://github.com/user-attachments/assets/74696d4a-563b-4a05-82a9-59a8b9069f11)

**What is the expected behavior?**

The feature flag should be checked only once.

**What is the actual behavior?**

The feature flag is checked 127 times.

## Reproduce

Could you provide the minimum required steps to resolve the issue you're seeing?

- pull `main`
- run `make start`
- enable `productCatalogFailure`feature flag
- navigate to jaeger and check any span with error from the `productCatalogService`.

julianocosta89:
@beeme1mr would you be able to take a look at this?

I've quickly checked the productCatalogService and the implementation seems about right:
https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/productcatalogservice/main.go#L253-L257

I couldn't find a place where we would set 127 span events in the productCatalogService, maybe I've missed something.

LMK if you want me to move this issue to the flagd repo.

beeme1mr:
It looks like a new OTel hook is registered globally for every request, which is very likely the root cause. Either it should be done once during initialization, or the hook should be set at the client level.

https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/productcatalogservice/main.go#L303

julianocosta89:
Thank you @beeme1mr! 🤩

PR #1736, opened 2024-10-07
fix: only register otel during initialization

beeme1mr:
# Changes

Move the feature flag hook registration to the main method. This prevents the hook from being registered on every request.

This change brings the `product-catalog-service` in line with the `checkout-service`.

https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/checkoutservice/main.go#L163-L164

## Merge Requirements

essential items:

Resolves: #1735

### DIFF
--- a/src/productcatalogservice/main.go
+++ b/src/productcatalogservice/main.go
@@ -128,10 +128,13 @@ func main() {
 		}
 		log.Println("Shutdown meter provider")
 	}()
-	openfeature.SetProvider(flagd.NewProvider())
 	openfeature.AddHooks(otelhooks.NewTracesHook())
+	err := openfeature.SetProvider(flagd.NewProvider())
+	if err != nil {
+		log.Fatal(err)
+	}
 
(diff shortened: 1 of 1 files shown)


src/productcatalogservice/main.go
```diff
@@ -128,12 +128,10 @@ func main() {
 		}
 		log.Println("Shutdown meter provider")
 	}()
-	err := openfeature.SetProvider(flagd.NewProvider())
-	if err != nil {
-		log.Fatal(err)
-	}
+	openfeature.SetProvider(flagd.NewProvider())
```
rogercoll, line 131: Is there a reason behind removing the error handling/propagation? I would suggest propagating the error whenever possible

beeme1mr: I'll revert this change. That does mean this service is hard-dependent on the feature flag service. Perhaps that's by design, but the other Go service behaves differently.

rogercoll (approved):

julianocosta89 (approved):

PR #1750, opened 2024-10-22
[chore] use otel-linux-arm64 runners for trace testing

puckpuck:
It uses the `otel-linux-arm64` runners for the integration tests workflow. This prevents the error caused by running out of disk space when using the default GitHub runners.

There are other risks here, as these runners have less memory and run on ARM64 instead of AMD64, but they do offer much more disk space.


rogercoll (approved):

rogercoll:
It seems we have another issue now: `dependency failed to start: container tracetest-server exited (1)`

Maybe because of moving into arm?

puckpuck:
Closing this. It was created to try and help with our disk space issues in the tracetests CI, but we fixed that already.

PR #1759, opened 2024-10-30
Makefile does not always use `DOCKER_COMPOSE_ENV` var

### LINKED ISSUE #1758: Makefile does not always use `DOCKER_COMPOSE_ENV` var
# Bug Report

Which version of the demo you are using? 167729a 

## Symptom

When running `make redeploy`, `make build` or `make restart`, it should take into account the values in the `.env.override` file.

## Reproduce

Could you provide the minimum required steps to resolve the issue you're seeing?

Let say that you add a `/src/adservice/Dockerfile.elastic` to the project to build a specific version of the image.
So you add the following line to the `.env.override` file:

```env
AD_SERVICE_DOCKERFILE=./src/adservice/Dockerfile.elastic
```

If you run:

```sh
make redeploy
```

The default `/src/adservice/Dockerfile` will be used as by default, docker compose will read the `.env` file only.

## Additional Context

We just have to change the way we are calling docker compose in the makefile. For example:

```makefile
restart:
# work with `service` or `SERVICE` as input
ifdef SERVICE
	service := $(SERVICE)
endif

ifdef service
	$(DOCKER_COMPOSE_CMD) stop $(service)
	$(DOCKER_COMPOSE_CMD) rm --force $(service)
	$(DOCKER_COMPOSE_CMD) create $(service)
	$(DOCKER_COMPOSE_CMD) start $(service)
else
	@echo "Please provide a service name using `service=[service name]` or `SERVICE=[service name]`"
endif
```

would become:

```makefile
restart:
# work with `service` or `SERVICE` as input
ifdef SERVICE
	service := $(SERVICE)
endif

ifdef service
	$(DOCKER_COMPOSE_CMD) $(DOCKER_COMPOSE_ENV) stop $(service)
	$(DOCKER_COMPOSE_CMD) $(DOCKER_COMPOSE_ENV) rm --force $(service)
	$(DOCKER_COMPOSE_CMD) $(DOCKER_COMPOSE_ENV) create $(service)
	$(DOCKER_COMPOSE_CMD) $(DOCKER_COMPOSE_ENV) start $(service)
else
	@echo "Please provide a service name using `service=[service name]` or `SERVICE=[service name]`"
endif
```

I'm going to send a PR.

dadoonet:
# Changes

Let say that you add a `/src/adservice/Dockerfile.elastic` to the project to build a specific version of the image. So you add the following line to the `.env.override` file:

```env
AD_SERVICE_DOCKERFILE=./src/adservice/Dockerfile.elastic
```

If you run:

```sh
make redeploy
```

The `/src/adservice/Dockerfile.elastic` should be used instead of the default one.

This commit adds `DOCKER_COMPOSE_ENV` variable anywhere it's needed.

Closes #1758.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -149,8 +149,8 @@ start-minimal:
 
 .PHONY: stop
 stop:
-	$(DOCKER_COMPOSE_CMD) down --remove-orphans --volumes
-	$(DOCKER_COMPOSE_CMD) -f docker-compose-tests.yml down --remove-orphans --volumes
+	$(DOCKER_COMPOSE_CMD) $(DOCKER_COMPOSE_ENV) down --remove-orphans --volumes
+	$(DOCKER_COMPOSE_CMD) $(DOCKER_COMPOSE_ENV) -f docker-compose-tests.yml down --remove-orphans --volumes
 	@echo ""
 	@echo "OpenTelemetry Demo is stopped."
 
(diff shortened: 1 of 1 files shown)


rogercoll (approved):
Looks great! Thanks for fixing this

trentm (approved):

dadoonet:
There's still an error with markdownlinkcheck but I did not change anything ;)

trentm:
> There's still an error with markdownlinkcheck but I did not change anything ;)

I think that was a fluke error: the link checker got a 502 for a particular GitHub repo URL during that check. Perhaps @rogercoll could re-run the test (I don't have access to re-run the workflow).

rogercoll:
> > There's still an error with markdownlinkcheck but I did not change anything ;)
> 
> I think that was a fluke error: the link checker got a 502 for a particular GitHub repo URL during that check. Perhaps @rogercoll could re-run the test (I don't have access to re-run the workflow).

Correct, rerunning it fixed the issue.

PR #1803, opened 2024-12-01
[chore] Remove force push true in nightly-build

julianocosta89:
# Changes

The Nightly Build Action should only push new images when there are file changes for that service.
Currently, the env var `push` is set to `true`, bypassing this condition.

This Pr removes the env var to let the decision be made on `check_changes` step:
https://github.com/open-telemetry/opentelemetry-demo/blob/main/.github/workflows/component-build-images.yml#L150-L165


rogercoll (commented):
The code that checks if there are file changes rely on the GHA executor being pull request:

```
FILES_CHANGED=$(git diff --name-only ${{ github.event.pull_request.base.sha }} ${{ github.sha }} -- $DOCKERFILE_DIR)
```

Do you know if that will work for nightly runs? Also, why not doing a build every night? Dependencies within Dockerfiles can change even if there are no file changes in them. For example, Dockerfile using `alpine` without pinning the patch version, the minor is overwritten whenever a new patch is available.

rogercoll:
> I'm fine in leaving it as is. But it seems like a lot of unnecessary images being pushed. All this data is basically useless and we are just increasing the footprint of the project. Currently we have 5844 images published: https://hub.docker.com/r/otel/demo/tags And we add 40 every night.

What about instead of creating a new tag every day that contains the full date, create a tag with the day of the week `nightly-{day}`? For example: `nightly-mon-frontend`, `nightly-tus-frontend`, etc. That would override the previous week's images, but I think keeping the last 7 days builds would be enough.

I agree on the issue of having this huge amount of images, some rotation mechanism should be applied.

julianocosta89:
> > I'm fine in leaving it as is. But it seems like a lot of unnecessary images being pushed. All this data is basically useless and we are just increasing the footprint of the project. Currently we have 5844 images published: https://hub.docker.com/r/otel/demo/tags And we add 40 every night.
> 
> What about instead of creating a new tag every day that contains the full date, create a tag with the day of the week `nightly-{day}`? For example: `nightly-mon-frontend`, `nightly-tus-frontend`, etc. That would override the previous week's images, but I think keeping the last 7 days builds would be enough.
> 
> I agree on the issue of having this huge amount of images, some rotation mechanism should be applied.

I like that!

I'll put some time later to have it in place.
in any case, this one can be closed.

Thx @rogercoll!

PR #1805, opened 2024-12-01
[accountingservice] fix-build-targetarch

### LINKED ISSUE #1798: accounting-service failing to start
# Bug Report

Which version of the demo you are using? ghcr.io/open-telemetry/demo:latest-accountingservice@sha256:a4c22f54468bf16fc32fc788e8f71d663b6886158847a5e555eb90381b6aae0e

## Symptom

The container is in a restart loop. The logs report:

```
Unhandled exception. System.IO.FileLoadException: Could not load file or assembly 'AccountingService, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'.
```

I would expect the container to reach a running state.

Container is in a restart loop.

## Reproduce

Run `make start`

We will close this issue if:

* The steps you provided are complex.
* If we can not reproduce the behavior you're reporting.

## Additional Context

I am attempting to run this on a Macbook Pro M2, so it would be pulling the arm64 versions.

julianocosta89:
# Changes

Fix #1798.
This PR adds `-r linux-musl-$TARGETARCH` to the `dotnet publish` step.


src/accountingservice/Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN dotnet build "./AccountingService.csproj" -r linux-musl-$TARGETARCH -c $BUIL
 
 FROM builder AS publish
```
MMartyn, line 17: You need to configure the ARG for each stage you want to use it in, otherwise it won't get set.

```suggestion
FROM builder AS publish
ARG TARGETARCH
```

MMartyn (approved):
LGTM. Thanks.

reyang (approved):

julianocosta89:
@MMartyn I ran the demo today with this change and it seems that the traces from `accountingservice` are now gone.
I'll revert this change to investigate the case further.

PR #1815, opened 2024-12-06
Variable paymentservice error rate and introduce loyalty level attributes

rcastley:
# Changes

Introduced an enhancement to the feature flag paymentServiceFailure:

- Selectable error rate percentage in the Flagd UI
- `app.loyalty.level` also added as a span attribute. Gold will show errors when feature flag set.
- Changed `pino` to include the `service.name` in log entries and also change from numeric level to text level e.g. info, warn etc.

This makes for a more interesting use case when the payment service fails. Only the `gold` loyalty level will show errors, all other levels will show success.

## Merge Requirements

essential items:


src/paymentservice/charge.js
```diff
@@ -1,24 +1,39 @@
...
+const { FlagdProvider } = require('@openfeature/flagd-provider');
 const flagProvider = new FlagdProvider();
 
 const logger = require('./logger');
 const tracer = trace.getTracer('paymentservice');
 const meter = metrics.getMeter('paymentservice');
-const transactionsCounter = meter.createCounter('app.payment.transactions')
+const transactionsCounter = meter.createCounter('app.payment.transactions');
+
+const TENANT_LEVEL = ['gold', 'silver', 'bronze'];
```
julianocosta89, line 16: By tenant level you mean something like user level?
Tenant is a bit misguiding to me.

src/paymentservice/charge.js
```diff
@@ -1,24 +1,39 @@
...
+
+/** Return random element from given array */
+function random(arr) {
+  const index = Math.floor(Math.random() * arr.length);
+  return arr[index];
+}
```
julianocosta89, line 22: Ideally we would configure this directly on the Feature Flag, using fractional evaluation: https://openfeature.dev/specification/glossary/#fractional-evaluation

src/paymentservice/charge.js
```diff
@@ -1,24 +1,39 @@
...
 module.exports.charge = async request => {
   const span = tracer.startSpan('charge');
 
   await OpenFeature.setProviderAndWait(flagProvider);
+
   if (await OpenFeature.getClient().getBooleanValue("paymentServiceFailure", false)) {
-    throw new Error("PaymentService Fail Feature Flag Enabled");
+    // 75% chance to fail with version 350.10 span tag
+    if (Math.random() < 0.75) {
+      span.setAttributes({ 'app.version': '350.10', 'app.tenant.level': random(TENANT_LEVEL) });
```
julianocosta89, line 32: To set `service.version` we are following the OTel semantic conventions.
What would `app.version` be? And is there any case where this value would be different from `service.version`?

Currently this is being set here: https://github.com/open-telemetry/opentelemetry-demo/blob/main/.env#L40

julianocosta89 (changes requested):
Hello @rcastley :wave::skin-tone-4: 

Thanks for suggesting this.
I'd change a bit the approach though. Instead of having 75% set on the code, ideally we would configure this directly on the Feature Flag, using fractional evaluation: https://openfeature.dev/specification/glossary/#fractional-evaluation
Imagine a slider, where we select the % of failure we want to have.

Hardcoding 75% will create some unexpected behaviour.

src/paymentservice/charge.js
rcastley: It's more of a customer status level, like a rewards programme level highlighting their customer loyalty. So, this could be change to `status_level` or `loyalty_level`.

src/paymentservice/charge.js
rcastley: Excellent idea, will look into this, thank you :-)

src/paymentservice/charge.js
julianocosta89: +1 for `loyalty_level`

src/paymentservice/charge.js
rcastley: `app.version` is specific to the `paymentservice` whereas (I believe) `service.version` applies to all services running the application. `app.version` has two versions where one will error and the other will not. Here is a visual that shows this off better (I hope) :-)

image

julianocosta89: I believe both are referring to the same `version`.
In the demo it happens to be the same, as we opted to set it globally, but each service could have its own version.

To set the app version following the OTel semantic conventions, we need to use `service.version`.

src/paymentservice/charge.js
rcastley: Or how about doing something like this:

image

Using the following `flagd` configuration:
```
    "paymentServiceFailure": {
      "description": "Fail payment service charge requests n%",
      "state": "ENABLED",
      "variants": {
        "100%": 1.00,
        "90%": 0.95,
        "75%": 0.75,
        "50%": 0.50,
        "25%": 0.25,
        "10%": 0.10,
        "off": 0.00
      },
      "defaultVariant": "off"
    },
```

src/paymentservice/charge.js
rcastley: Moving to use `app.variant` instead with the using values `A` & `B`. Removes confusion around versions. Service will still maintain the version set in the resource attributes.

julianocosta89: I believe that in this case is not matter of key, but actually meaning.
You will call it `app.variant` but it will still be a version identifier.

Let me bring some folks to the discussion, I'd like to hear more opinions.
@rogercoll, @puckpuck and @mviitane WDYT?

kubernetes/opentelemetry-demo.yaml
```diff
@@ -542,8 +542,13 @@ data:
...
           "variants": {
-            "on": true,
-            "off": false
+            "100%": 1.00,
+            "90%": 0.95,
+            "75%": 0.75,
+            "50%": 0.50,
+            "25%": 0.25,
+            "10%": 0.10,
+            "off": 0.00
```
puckpuck, line 551: This is a generated file. Please revert the changes to this file, which will get updated after a Helm release is complete.

puckpuck (changes requested):

src/paymentservice/charge.js
puckpuck: I don't think `app.variant` makes sense here. Are we trying to demonstrate a fictional state that a failure only happens on a certain app variant or deployed version? This demonstration mechanism should be handled within the vendor's demo or some other means ([DemoMonkey](https://www.demomonkey.net/) comes to mind). I wouldn't want to give a user the impression that we deployed 2 different variants/versions of the app.

This should be called `app.<featureflagname>` and have its value be what the feature flag returned.

src/paymentservice/charge.js
```diff
@@ -40,29 +57,34 @@ module.exports.charge = async request => {
...
   const baggage = propagation.getBaggage(context.active());
-  if (baggage && baggage.getEntry("synthetic_request") && baggage.getEntry("synthetic_request").value === "true") {
+  if (baggage && baggage.getEntry('synthetic_request') && baggage.getEntry('synthetic_request').value === 'true') {
     span.setAttribute('app.payment.charged', false);
   } else {
     span.setAttribute('app.payment.charged', true);
   }
 
-  span.end();
+  span.setAttributes({ 'app.variant': 'A', 'app.loyalty.level': random(LOYALTY_LEVEL) });
```
puckpuck, line 82: This `app.loyalty.level` attribute is added similarly, but the code is repeated multiple times across the various scenarios. Can we break this out to add just this attribute by itself at the top of the charge function and let the logic control the other span attribute only?

rcastley: I have removed the repetition of the `setAttributes` and now have the scenario that when the feature flag is set only `gold` level will error. Have introduced another level `platinum` so along with `silver` and `bronze` will show success.

image

src/paymentservice/charge.js
rcastley: `app.variant` removed from PR.

src/paymentservice/charge.js
```diff
@@ -1,24 +1,41 @@
...
+
+  const numberVariant =  await OpenFeature.getClient().getNumberValue("paymentServiceFailure", 0);
+
+  if (numberVariant > 0) {
+    // n% chance to fail with app.loyalty.level=gold
+    if (Math.random() < numberVariant) {
+      span.setAttributes({'app.loyalty.level': 'gold' });
+      span.end();
+
+      throw new Error('Payment request failed. Invalid token. app.logalty.level=gold');
```
julianocosta89, line 37:
```suggestion
      throw new Error('Payment request failed. Invalid token. app.loyalty.level=gold');
```

src/paymentservice/charge.js
```diff
@@ -1,24 +1,41 @@
...
+const { FlagdProvider } = require('@openfeature/flagd-provider');
 const flagProvider = new FlagdProvider();
 
 const logger = require('./logger');
 const tracer = trace.getTracer('paymentservice');
 const meter = metrics.getMeter('paymentservice');
-const transactionsCounter = meter.createCounter('app.payment.transactions')
+const transactionsCounter = meter.createCounter('app.payment.transactions');
+
+const LOYALTY_LEVEL = ['platinum', 'silver', 'bronze'];
```
puckpuck, line 16: Can we add `gold` back to this? So, Success also includes gold, but failures are gold only.

```suggestion
const LOYALTY_LEVEL = ['platinum', 'gold', 'silver', 'bronze'];
```

rcastley: Just taking this for a spin, to ensure all works as expected!

rcastley: All tested and working as expected.

puckpuck (approved):
Thank you @rcastley this is great!

Issue #1788, opened 2024-11-21
Drop "service" suffix, and standardize names


puckpuck:
When running the demo, it is composed of 25+ individual containers/components/microservices. The names of these components are carried over from the demo's original roots, and several additional services and external components were added to support the demo's goal of showcasing OpenTelemetry across various languages and frameworks.

Currently, most services in the demo have a `service` suffix as part of their image and OpenTelemetry service name. Containers will have a `-service` suffix. Some services, such as the product catalog service, have a space in their name. The container name will be hyphenated, but the service/pod and OpenTelemetry definitions will have this as a single word.

A well-understood, concise, and consistent naming for all services would significantly improve the usability of the demo for users in observability tooling as well as contributors that support this effort. I propose dropping the `service` suffix from all services and using a hyphen (-) for spaces in names. The names should be consistent across the docker compose service, container name, OpenTelemetry service name, and Kubernetes deployment name.

The following table describes the current and suggested name for each Demo component:
EDIT: Updated to incorporate feedback from comments

| Current Name           | Suggested Name  |
|------------------------|-----------------|
| accountingservice      | accounting      |
| adservice              | ad              |
| cartservice            | cart            |
| checkoutservice        | checkout        |
| currencyservice        | currency        |
| emailservice           | email           |
| flagd                  | flagd           |
| flagd-ui               | flagd-ui        |
| frauddetectionservice  | fraud-detection |
| frontend               | frontend        |
| frontendproxy          | frontend-proxy  |
| frontend-web           | frontend-web    |
| grafana                | grafana         |
| imageprovider          | image-provider  |
| jaeger                 | jaeger          |
| kafka                  | kafka           |
| loadgenerator          | load-generator  |
| opensearch             | opensearch      |
| otelcollector          | otel-collector  |
| paymentservice         | payment         |
| productcatalogservice  | product-catalog |
| prometheus             | prometheus      |
| quotesservice          | quote           |
| recommendationsservice | recommendation  |
| shippingservice        | shipping        |
| valkey-cart            | valkey-cart     |

I welcome feedback and suggestions on this proposal. 

UPDATE:
After discussing in a SIG meeting on December 11, we decided to move forward with this proposal. 

This is a tracking list for each demo component so we can ensure the effort is complete:
- [x] accounting
- [x] ad
- [x] cart
- [x] checkout
- [x] currency
- [x] email
- [x] flagd
- [x] flagd-ui
- [x] fraud-detection
[...]

PR #1848, opened 2024-12-17
Fix tag_suffix for accounting service

alanwest:
The Helm chart generates a kubernetes manifest that expects the suffix of the account service's image to be `accountingservice`.

https://github.com/open-telemetry/opentelemetry-demo/blob/fa7847efb691e323236955c6126d486cf2c1b558/kubernetes/opentelemetry-demo.yaml#L9533-L9535

Few options to fix this:

1. This PR
2. Fix the Helm chart to use the new name
3. Others?

I'd be happy to follow up on whatever solution people would prefer.

~~If people like the solution in this PR, I think it would also make sense to change the directory name and all contained files back to `accountingservice` for consistency - e.g., `/src/accounting` becomes `/src/accountingservice`. Thoughts?~~

I see now that there is an effort to remove the term `service` from the names of the services in #1788. I recommend leaving the tag_suffix for services untouched for now.

### DIFF
--- a/.github/workflows/component-build-images.yml
+++ b/.github/workflows/component-build-images.yml
@@ -52,7 +52,7 @@ jobs:
       matrix:
         file_tag:
           - file: ./src/accounting/Dockerfile
-            tag_suffix: accounting
+            tag_suffix: accountingservice
             context: ./
             setup-qemu: true
           - file: ./src/adservice/Dockerfile


puckpuck:
Closing this as the nightly releases are not part of anything we would deploy on Kubernetes or do with a fresh repo clone + `make a start.` The Kubernetes deployments that we support are fixed to the last good release, which is 1.12.0 right now. The nighlies will carry the new names for services as we change and we need to continue to do that for development reasons.

PR #1854, opened 2024-12-19
[paymentservice] Bump dependencies

julianocosta89:
# Changes

Bump dependencies of PaymentService and upgrade node to LTS version.


src/paymentservice/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "paymentservice",
-  "version": "1.4.0",
+  "version": "1.12.0",
```
puckpuck, line 3: Do we need this entry here? Seems like an extra place we need to maintain a version number that I would prefer not to.

puckpuck (approved):
I left 1 comment about needing to keep a version in package.json, but approved regardless.

PR #1860, opened 2024-12-23
[react-native-app] cleanup build instructions

puckpuck:
# Changes

This updates the Readme for the react-native-app. All commands to run have not changed.

1. Adds structure to the file with use of H3 and H4 blocks
2. Re-orders some steps that only need to be run the first time
3. Re-orders troubleshooting to have topics show up in same order as build instructions
4. Enforces line breaks at 80 characters on each line
5. Removes link to Android Developer site that causes maxRedirects error with markdown-link-check


julianocosta89 (approved):
LGTM

PR #1898, opened 2025-01-10
Revert "build(deps): bump the composer-production-dependencies group across 1 directory with 11 updates"

julianocosta89:
Reverts open-telemetry/opentelemetry-demo#1890


rogercoll (approved):
Do we know why is broken? Is it because of a specific dependency?

PR #1919, opened 2025-01-16
load product catalog periodically

### LINKED ISSUE #1892: reload Product catalog periodically
# Feature Request

Currently, the product catalog is initialized only on service startup. If a change is made to products after the service is started, the changes will not be recognized. We should change the product catalog to load on startup and on a configurable periodic interval (e.g., every 10 seconds).

puckpuck:
# Changes

Fixes #1892 

- adds a new environment variable called: `PRODUCT_CATALOG_RELOAD_INTERVAL`
- reloads the product catalog according to an interval based on the env variable (defaults to 10 seconds)
- mounts the product.json file at run time instead of including it at image build time

This allows the user to change the products without rebuilding the product-catalog service. The products can also be updated after the demo is running. The change also paves the way for reloading the product catalog on a period timer, allowing the product list to be backed by more dynamic sources such as a database. 

Finally, this change will require some updates to the Helm chart to load and mount the product list via a ConfigMap.

## Merge Requirements

essential items:


src/product-catalog/products/products.json
```diff
@@ -26,7 +26,7 @@
     },
     {
       "id": "1YMWWN1N4O",
-      "name": "Eclipsmart Travel Refractor Telescope",
+      "name": "THIS IS MY PRODUCT NAME",
```
rogercoll, line 29: Left over?

src/product-catalog/main.go
```diff
@@ -178,6 +177,43 @@ type productCatalog struct {
...
+		if interval <= 0 {
+			interval = DEFAULT_RELOAD_INTERVAL
+		}
+	}
+	log.Infof("Product Catalog reload interval: %d", interval)
+
+	ticker := time.NewTicker(time.Duration(interval) * time.Second)
+
+	go func() {
+		for {
```
rogercoll, line 203: It would be best to have the programs' context here so it can be properly shutdown, but it seems it would require a major refactor to the codebase (all other init functions as well). We can keep it for a future improvement.

docker-compose.minimal.yml
```diff
@@ -399,6 +399,7 @@ services:
     environment:
       - FLAGD_HOST
       - PRODUCT_CATALOG_PORT
+      - PRODUCT_CATALOG_RELOAD_INTERVAL
```
rogercoll, line 402: Should we mount the products to the container?

rogercoll (commented):
lgtm overall

docker-compose.minimal.yml
puckpuck: products.json is mounted in the minimal config now

src/product-catalog/main.go
puckpuck: This is something I think we can do in a future improvemenbt

rogercoll (approved):

julianocosta89 (approved):

Issue #1866, opened 2024-12-23
Collector complains about "Could not dynamically add mapping for field [http.connection.id]"


mviitane:
# Bug Report

Which version of the demo you are using?

Using main latest:
$ git log
commit e3548c621744514f48e71d7ada96632edd345545 (HEAD -> main, upstream/main)
Date:   Sun Dec 22 08:49:02 2024 -0500

## Symptom
Collector batchprocessor and opensearch exporter complains about 
`Could not dynamically add mapping for field [http.connection.id]`
`Existing mapping for [attributes.http] must be of type object but found [text]`

```
otel-collector  | 2024-12-23T12:40:13.035Z	warn	batchprocessor@v0.116.0/batch_processor.go:263	Sender failed	{"kind": "processor", "name": "batch", "pipeline": "logs", "error": "not retryable error: Permanent error: {\"type\":\"mapper_parsing_exception\",\"reason\":\"Could not dynamically add mapping for field [http.connection.id]. Existing mapping for [attributes.http] must be of type object but found [text].\",\"caused_by\":
...
otel-collector  | 2024-12-23T12:40:13.449Z	error	internal/base_exporter.go:139	Exporting failed. Rejecting data.	{"kind": "exporter", "data_type": "logs", "name": "opensearch", "error": "not retryable error: Permanent error: {\"type\":\"mapper_parsing_exception\",\"reason\":\"Could not dynamically add mapping for field [http.connection.id]. Existing mapping for [attributes.http] must be of type object but found [text].\",\"caused_by\":
...
```

**What is the expected behavior?**

No complains.

## Reproduce

Could you provide the minimum required steps to resolve the issue you're seeing?

docker compose build
docker compose run
docker compose logs otel-collector

## Workaround

Disable logs pipeline in the Collector.

## Additional context

Related to https://github.com/open-telemetry/opentelemetry-demo/issues/1833

However, this occurs with standard demo components without enabling any vendor backends.

puckpuck:
Can you do a `docker system prune -a`, doing a new fresh build and trying again

mviitane:
I did and got the same problem. I'm running Docker on x86_64, Ubuntu.

TrevorSquillario:
I'm seeing this same error and it's causing the logs not to show up in opensearch. 

```
Sender failed   
{"kind": "processor", "name": "batch", "pipeline": "logs", "error": "not retryable error: Permanent error: 
[...]

PR #1930, opened 2025-01-21
Fix envoy access logs

mviitane:
# Changes

Fix envoy access logs
- Remove http attributes that cause problems in the Collector pipelines
- Fixes https://github.com/open-telemetry/opentelemetry-demo/issues/1866

## Merge Requirements

essential items:


rogercoll (commented):
Some of these attributes are defined by the OpenTelemetry Semantic Conventions: https://github.com/open-telemetry/semantic-conventions/tree/main/model/http
 
 Would it be feasible to fix the opensearch mappings instead?

mviitane:
> Some of these attributes are defined by the OpenTelemetry Semantic Conventions: https://github.com/open-telemetry/semantic-conventions/tree/main/model/http
> 
> Would it be feasible to fix the opensearch mappings instead?

I know that some of these are in the semantic conventions, but I believe Envoy doesn't send them correctly.

I don't see how we could fix this with the OpenSearch mapping because the batch processor also shows the same problem.

rogercoll:
> I don't see how we could fix this with the OpenSearch mapping because the batch processor also shows the same problem.

The error is shown in the batch processor because is the error propagated by the OpenSearch exporter, see https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/batch_processor.go#L263

mviitane:
OK, I didn't realize this is only opensearch related. I wonder why the automatic mappings are not enough.
https://opensearch.org/docs/latest/field-types/

rogercoll:
> OK, I didn't realize this is only opensearch related. I wonder why the automatic mappings are not enough. https://opensearch.org/docs/latest/field-types/

Not sure, I am not very familiar with OpenSearch. Have you tried removing the corresponding index? (maybe the `http.connection.id` mapping was created with some other data)

mviitane:
> > OK, I didn't realize this is only opensearch related. I wonder why the automatic mappings are not enough. https://opensearch.org/docs/latest/field-types/
> 
> Not sure, I am not very familiar with OpenSearch. Have you tried removing the corresponding index? (maybe the `http.connection.id` mapping was created with some other data)

I ran `docker system prune -a` and started a clean environment and got the problem almost right away. Are you not seeing the same problem?

rogercoll:
> > > OK, I didn't realize this is only opensearch related. I wonder why the automatic mappings are not enough. https://opensearch.org/docs/latest/field-types/
> > 
> > 
> > Not sure, I am not very familiar with OpenSearch. Have you tried removing the corresponding index? (maybe the `http.connection.id` mapping was created with some other data)
> 
> I ran `docker system prune -a` and started a clean environment and got the problem almost right away. Are you not seeing the same problem?

Not on my environment, just tested with nightly `DEMO_VERSION=nightly-12898327927` and collector logs are clear.

puckpuck:
@mviitane I suspect the issue may be related to another service emitting this same attribute but with a different data type. When you run into it, can you use the OpenSearch API to query for this attribute and determine which service is emitting it?

mviitane:
> @mviitane I suspect the issue may be related to another service emitting this same attribute but with a different data type. When you run into it, can you use the OpenSearch API to query for this attribute and determine which service is emitting it?

I grepped the debug exporter (verbose) and saw that all signals with `http.connection.id` are logs with 
`event.name: Str(proxy.access)`, so I think they're all from Envoy.

otel-collector  |      -> event.name: Str(proxy.access)
otel-collector  |      -> http.connection.id: Str(85)

puckpuck:
If you update the PR to only drop the `http.connection.id` field do the other ones from Envoy still work fine? Maybe we do just that instead.

mviitane:
> If you update the PR to only drop the `http.connection.id` field do the other ones from Envoy still work fine? Maybe we do just that instead.

No, unfortunately, that wasn't enough. I tried a few of the first ones.

puckpuck:
We should merge this PR, but I want to create an issue so we can learn more about the actual root cause. This only affects specific environments and is related to attributes that start with `http.`.

@mviitane, can you resolve the conflicts so we can merge this?

puckpuck (approved):

Issue #2051, opened 2025-02-19
Load Generator (locust) doesn't make the application emit traces


aloussase:
Hello,

I am using the Load Generator included with the demo.

I would expect it to cause the application to emit traces, but it doesn't.

I am inspecting the traces on WireShark with the following filter:
```
tcp.port == 8080 && http && http.request.uri contains "traces" && http.request
```

This works when I manually use the application, but not when using the Load Generator.

My guess is just hitting the API endpoints does not cause the app to emit trace data.

When exactly is trace data emitted?

julianocosta89:
Hello @aloussase, the load generator does generate traffic and emit traces.
Are you changing anything on your end?
If you run the Demo as is, you can navigate to Jaeger and visualise traces without navigating to any page from the Demo.

aloussase:
Hello @julianocosta89 thanks for your reply

https://github.com/user-attachments/assets/f1892077-30d4-4c11-b985-bf5173597243

I have added a demo of the behaviour I am experiencing. As you can see, the load generator doesn't seem to produce traces.

Actually, before trying the load generator I wrote some scripts to simulate some scenarios, since I didn't know about the load generator. But these scripts did not cause the application to emit traces either.

This makes me think that traces are being emitted from somewhere else.

Do you see anything I might be doing wrong in this demo?

Thank you very much.

julianocosta89:
You are correct, I've just tested and it seems that playwright is not working properly.

Thanks for reporting it, we'll look into it

puckpuck:
I wonder if the default Otel SDK batch 5000ms timeout is the cause here?

aloussase:
I am now using this code with Playwright to simulate the application:

```python
from playwright.sync_api import sync_playwright, expect
import random
import logging

import config


logger = logging.getLogger(__name__)


def execute(n: int = 1):
    """This workflow adds some products to the cart and then checks out."""
    with sync_playwright() as p:
        browser = p.chromium.launch()

        context = browser.new_context()
        context.set_default_timeout(5000)

        iterations = 0

        while iterations < n:
            page = context.new_page()
            page.goto(config.get_app_url())

            product_list = page.locator('css=[data-cy="product-list"]')
            logging.debug("Got products list")

            product_card = product_list.locator("a").nth(random.randrange(0, 4))
            product_url = product_card.get_attribute("href")
            logging.debug("Selected a product card: %s", product_url)

[...]

PR #2114, opened 2025-03-10
feat(load-generator): Adapt OpenFeature OFREP instead of IN_PROCESS

aepfli:
# Changes

OpenFeature supports multiple different modes.
The newest implementation relies on threads and other python internal, which seem to be monkeypatched by locust.
Due to this, the IN_PROCESS mode is not working properly anymore and is causing issues.

Instead we suggest to switch to the OFREP (OpenFeature Remote Evaluation Protocol) for this usecase. This is a simple http call which will return all evaluate flags, and will regularly poll for changes.

relates:
- https://github.com/open-feature/python-sdk-contrib/issues/198
- https://github.com/open-telemetry/opentelemetry-demo/issues/2051

## Merge Requirements

essential items:


aepfli:
i created https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1563 - but my helm knowledge is limited, if there is anything i can or should improve please let me know

aepfli:
I checked the documentation, but i could not find any deeper explanation of the ports, nor the used modes for OpenFeature, hence i do not see any need for a documentation update. wdyt?

julianocosta89 (approved):

julianocosta89:
@aepfli this one should be good to go.
In regards the Helm chart, it requires a couple of other steps, but I'll get back to you on the other PR.

I've added to your PR the flagd ports on the `docker-compose.minimal.yml` file, hope you don't mind.

Is there anything here you want us to wait before merging it?

julianocosta89 (changes requested):
I've noticed that in the Helm update, you are adding `FLAGD_OFREP_PORT` to the load-generator, but not in here.
Is that missing?

aepfli:
> Is there anything here you want us to wait before merging it?

from my side we can go ahead with this pr ;) thank you

aepfli:
okay here comes my limited OpenTelemetry know-how ;) 

Yes there is, our `OpenTelemetryHook` implementation is really simple and we are "attaching" ourselves to the current span. the question is, who is responsible to create the current span, and should this maybe be done by the load-generator. see https://github.com/open-feature/python-sdk-contrib/blob/345e7934b9de3879d3aff45c8213ece1a98e3711/hooks/openfeature-hooks-opentelemetry/src/openfeature/contrib/hook/opentelemetry/__init__.py#L16-L42

As I said, my knowledge here is limited but am I on the right track, and should we generate a new span for each [WebsiteUser](https://github.com/open-telemetry/opentelemetry-demo/blob/ae753db6cc13455b621b6356e5a596c74d6b2f87/src/load-generator/locustfile.py#L174-L178) invocation?

julianocosta89:
But Loadgen is also doing this:

https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/load-generator/locustfile.py#L73

🤔

aepfli:
we are never creating spans, only adding ourselves to existing ones. So I am not sure, if we should extract this into an own topic, and think about what is a good span for a loadtest. When to create a new one, etc. wdyt?

julianocosta89:
Had some time again to look at this.
It seems that the Loadgen is creating the span:
image

But it looks like the context is not being propagated with the hook.

aepfli:
most likely the span is generated with `self.client.get("/")` (https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/load-generator/locustfile.py#L172) - but that is after flag evaluation, hence we do not have a span - or am i wrong here? question do you see the span correctly attached in the previous version?

julianocosta89:
Yeah, it makes sense to keep them as separate spans.

I wonder if the call shouldn't have a span and be linked to the flagd though

beeme1mr:
> I wonder if the call shouldn't have a span and be linked to the flagd though

What's the goal? The current behavior seems correct unless we want to manually create a span within the load generator.

julianocosta89:
@beeme1mr the loadgen span is already being created no?!
https://github.com/open-telemetry/opentelemetry-demo/pull/2114#issuecomment-2715906106

aepfli:
> @beeme1mr the loadgen span is already being created no?! [#2114 (comment)](https://github.com/open-telemetry/opentelemetry-demo/pull/2114#issuecomment-2715906106)

I was just assuming that `client.get` creates a log span, but if it is, this span is generated in that method, before we evaluate flags. hence i think this is an additional possible improvement, and should be most likely extracted into an own topic.

julianocosta89:
> > @beeme1mr the loadgen span is already being created no?! [#2114 (comment)](https://github.com/open-telemetry/opentelemetry-demo/pull/2114#issuecomment-2715906106)
> 
> I was just assuming that `client.get` creates a log span, but if it is, this span is generated in that method, before we evaluate flags. hence i think this is an additional possible improvement, and should be most likely extracted into an own topic.

Totally! That would indeed be a different issue, I'm just discussing here because we are at it.
But let me get this one merged

aepfli:
see: https://github.com/open-telemetry/opentelemetry-demo/pull/2119 - where i add spans around each task of the locust load generator

PR #2119, opened 2025-03-13
feat(load-generator): add spans for feature flag evaluation

aepfli:
# Changes

As the span was created before the feature flag evaluation, we are now creating a span for each task, so we can enhance it with information when needed.

image

## Merge Requirements

essential items:

### DIFF
--- a/src/load-generator/locustfile.py
+++ b/src/load-generator/locustfile.py
@@ -158,7 +158,7 @@ class WebsiteUser(HttpUser):
         self.client.post("/api/cart", json=cart_item)
 
     @task(1)
-    @tracer.start_as_current_span("checkout with items")
+    @tracer.start_as_current_span("checkout with one item")
     def checkout(self):
         # checkout call with an item added to cart
         user = str(uuid.uuid1())
--- a/src/load-generator/locustfile.py
+++ b/src/load-generator/locustfile.py
@@ -36,6 +36,7 @@ from opentelemetry.sdk.resources import Resource
 from openfeature import api
 from openfeature.contrib.provider.ofrep import OFREPProvider
 from openfeature.contrib.hook.opentelemetry import TracingHook
+from opentelemetry.trace import Tracer
 
 from playwright.async_api import Route, Request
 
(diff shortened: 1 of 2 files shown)


src/load-generator/locustfile.py
```diff
@@ -149,6 +158,7 @@ def add_to_cart(self, user=""):
         self.client.post("/api/cart", json=cart_item)
 
     @task(1)
+    @tracer.start_as_current_span("checkout with items")
```
aepfli, line 161:
```suggestion
    @tracer.start_as_current_span("checkout with one item")
```

julianocosta89:
I don't understand this PR.

The `client.get` already have an span being created automatically:

image


`opentelemetry.instrumentation.requests` instruments the outgoing calls.
My point on https://github.com/open-telemetry/opentelemetry-demo/pull/2114#issuecomment-2715906106 is that this is not linked at all with flagd.

aepfli:
i think we might have been talking about different things. i thought it is important to add the evaluation event to the span of the "flood home" action  - which is actually the important information of the evaluation.

Fetching the flags used to be a call all the time with the rpc mode. but with the ofrep mode, we are fetching a generate flag configuration regularly, and have this configuration stored locally, and never doing a call for a flag evaluation to the host system. We could eg. only poll for this changes every x seconds, currently it is every 5 seconds.

julianocosta89:
@mviitane and @puckpuck I'd like you opinion here.
I just had a chat with @aepfli and my initial concern is actually being tacked in this PR: https://github.com/open-feature/flagd/pull/1593

This one here add spans to each call of the load-gen.
On one hand, this is nice, because it gives us span events on feature flag evaluation, but on the other hand, it bloats the number of spans produced by load-gen.

In a real life scenario, I'd love to be able to see all feature flag evaluation during my user journey, but I'm not sure if we want that for load-generator.

@aepfli would be fine in closing this issue, and I'm already happy with the `flagd` one I've mentioned above.

What do you think?

puckpuck:
I don't think we need a flag for each loadgen evaluation. We can kill close this PR.

Issue #2124, opened 2025-03-15
payment service crashes with OTEL_SDK_DISABLED


rlankfo:
# Bug Report

Which version of the demo you are using? (please provide either a specific
[commit
hash](https://github.com/open-telemetry/opentelemetry-demo/commits/main)
or a specific
[release](https://github.com/open-telemetry/opentelemetry-demo/releases)).
2.0.1

## Symptom

payment service crashes when `OTEL_SDK_DISABLED` env var is `true`

**What is the expected behavior?**

Expected behavior is that the payment service honors this variable as described in the docs:
`If “true”, a no-op SDK implementation will be used for all telemetry signals. Any other value or absence of the variable will have no effect and the SDK will remain enabled.`

https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration

What do you expect to see?

payment service runs with no-op sdk implementation.

**What is the actual behavior?**

payment service crashes

```
> start
> node --require ./opentelemetry.js index.js

{"level":"info","time":1742007177599,"pid":21,"hostname":"payment-7b7b5f6dc7-jcf5n","service.name":"payment","msg":"payment gRPC server started on port 8080"}
[...]

PR #2129, opened 2025-03-20
Ensure we handle the scenario the payment service code expects there to be an active span

keyoke:
I am proposing a very minor change to handle when auto-instrumentation is disabled( similar to #2124 ) for the payment service. In this scenario, an exception is raised but I would expect it to be handled gracefully.

# Changes

Added some very simple logic to check if the current active span is undefined, if it is we create a new span & mark this as the active span. 

## Merge Requirements

essential items:

### DIFF
--- a/src/payment/index.js
+++ b/src/payment/index.js
@@ -16,6 +16,8 @@ async function chargeServiceHandler(call, callback) {
   if (!span)
   {
     span = tracer.startSpan('chargeServiceHandler');
+    // Mark this new span as active
+    tracer.setSpan(tracer.context.active(), span)
     startedSpan = true;
   }
 
--- a/src/payment/index.js
+++ b/src/payment/index.js
@@ -4,12 +4,20 @@ const grpc = require('@grpc/grpc-js')
 const protoLoader = require('@grpc/proto-loader')
 const health = require('grpc-js-health-check')
 const opentelemetry = require('@opentelemetry/api')
+const tracer = opentelemetry.trace.getTracer(process.env.OTEL_SERVICE_NAME);
 
 const charge = require('./charge')
 const logger = require('./logger')
 
 async function chargeServiceHandler(call, callback) {
-  const span = opentelemetry.trace.getActiveSpan();
+  // Check if we have an active span if not start a new one
+  let span = opentelemetry.trace.getActiveSpan();
+  let startedSpan = false;
+  if (!span)
+  {
+    span = tracer.startSpan('chargeServiceHandler');
+    startedSpan = true;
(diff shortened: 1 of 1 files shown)


julianocosta89:
@pichlermarc I would love to hear your opinion here

puckpuck:
I understand if the SDK is disabled, the service should not throw an error which is what #2124 discusses. 

I struggle to understand in what scenario we need to start a trace if one does not exists.

src/payment/index.js
```diff
@@ -4,12 +4,22 @@ const grpc = require('@grpc/grpc-js')
 const protoLoader = require('@grpc/proto-loader')
 const health = require('grpc-js-health-check')
 const opentelemetry = require('@opentelemetry/api')
+const tracer = opentelemetry.trace.getTracer(process.env.OTEL_SERVICE_NAME);
```
pichlermarc, line 7: `process.env.OTEL_SERVICE_NAME` may be `undefined`, which is not a valid name for the tracer

pichlermarc (commented):
> I struggle to understand in what scenario we need to start a trace if one does not exists.

I agree - in my opinion the way to address the problem of the crashing app would be either to:
- optional chain on the span object obtained via `getActiveSpan()` wherever it's used 
  - example: `span.setAttributes({...})` -> `span?.setAttributes({...})`
- create and end an internal span always no matter if a trace was already started or not to which you can attach any attributes and errors you like, while retaining semantics for the grpc span that's likely being modified here (since that one is strictly supposed to follow the [gRPC semconv](https://opentelemetry.io/docs/specs/semconv/rpc/grpc/#grpc-status) anyway and modifying error status may interfere with that).

src/payment/index.js
keyoke: Ok thanks understood. I have switched to optional chain as suggested above

keyoke: So I guess I misundertood a little - Should we create the internal span? I think this makes more sense as we retain the original semantics

julianocosta89:
@pichlermarc this is the issue I've mentioned to you during KubeCon.
Would you mind taking a look?

pichlermarc (approved):

pichlermarc:
@julianocosta89 - sorry for the delay. This looks good.
Thank you @keyoke for working on this 🙂

src/payment/index.js
pichlermarc: All good, the way you did it is the way to go 🙂
Thanks for adapting it :)

puckpuck (approved):

PR #2130, opened 2025-03-21
Improve logging to demo Triage and Log grouping     

mmanciop:
# Changes

1. Remove explicit attributes from log records (we want Log Grouping to extract them)
2. Mark severity as Errors for logs about product not found (LogAI does not override the severity_number set by the log bridge)
3. Remove error event, we have logs now instead
4. Add stacktrace to error log


mmanciop:
Opened against wrong repo, sorry

PR #2167, opened 2025-05-05
[chore] Downgrade gRPC

julianocosta89:
# Changes

Downgrading gRPC due to https://github.com/grpc/grpc/issues/38538


mviitane (approved):
What do you think about adding a small note about the issue? We could prevent the future upgrade to these faulty versions.

rogercoll (approved):

Kielek (approved):

Issue #912, opened 2023-06-01
Adding new service to save orders to database


luke6Lh43:
# Feature Request

Hello Team. 

I've been playing a bit with OpenTelemetry demo and started to think if there is any specific reason why we don't have any database instance (apart from Feature Flag Store)? I believe the easiest way to introduce it would be to create new service which reads orders from queue (Kafka) and save them to Database (e.g. MySQL or MongoDB). Then there could be some additional function added to Accounting Service which reads data from DB and perform some calculation. This additional feature would create additional capabilities to emphasise how great project OpenTelemetry is.

What do you think about this idea?

puckpuck:
I like this idea, but I wonder if we can use Postgres here as a shared database that can do both the Feature flag service and this new order persistence service.

luke6Lh43:
Yes, it should work too :) 

Ideally, we should have opportunity to deploy various types of DBs (MySQL, PostgreSQL, MongoDB) and enable them in helm chart, so we could give a user (vendor) opportunity to test/showcase particular DB connection. From the service perspective it shouldn't be a problem as we could use env variable to choose DB which is currently used.

Kimbohlovette:
Hello @jpkrohling can I take this issue?

Kimbohlovette:
Hello @puckpuck this is a run down of how I'm going to implement at least for what we want to achieve
- Add a database service to the docker-compose file called `database`
- Create a golang service with the following
       - kafka consumer - api to establish connection with the kafka service and listen to make order messages
       - connection to a postgresql database
       
 Of course we can add options for users to specify other database types to use in the later on.
 What do you think?

puckpuck:
@Kimbohlovette , I don't think we need a new service for this. The `accountingservice` is a Go service that already exists as a Kafka consumer for all orders. This should be a good starting point to add logic to connect to a database and read/write some records.

Also please call the database service `postgresql` or `mysql` etc.

julianocosta89:
@Kimbohlovette and @puckpuck I think we need to discuss further this before starting implementing.

AFAIK `accountingservice` is about to be replaced by an Elixir service.
And I think @joshleecreates is already looking at it.

Kimbohlovette:
> @Kimbohlovette , I don't think we need a new service for this. The `accountingservice` is a Go service that already exists as a Kafka consumer for all orders. This should be a good starting point to add logic to connect to a database and read/write some records.
> 
[...]

PR #2143, opened 2025-04-08
Add Dapr on in the Cart Service

henrikrexed:
# Changes

Please provide a brief description of the changes here.

Service impacted: cart service
I have removed the usage of the redis client to replace it with the usage of Dapr client.
I have also added the Dapr folder container : 
* Dapr configuration to enable traces
* the component directory that will hold the configuration to the backend. for cart service the default backend will be valley
* added changes in the docker-compose to have Dapr configure 

I also added the new service : sqlite that will be used in the future for he product catalog ( replace file usage with a database).

I have prepare the changes on the helm chart but  I will create a pr  once this has been merged.

## Merge Requirements

For new features contributions, please make sure you have completed the following
essential items:

* [X ] `CHANGELOG.md` updated to document new feature additions
* [ ] Appropriate documentation updates in the [docs][]
* [ ] Appropriate Helm chart updates in the [helm-charts][] 

Maintainers will not merge until the above have been completed. If you're unsure
which docs need to be changed ping the
[@open-telemetry/demo-approvers](https://github.com/orgs/open-telemetry/teams/demo-approvers).

[docs]: https://opentelemetry.io/docs/demo/
[helm-charts]: https://github.com/open-telemetry/opentelemetry-helm-charts

PR #2175, opened 2025-05-08
store order to postgresql from accounting service

cuichenli:
# Changes

Add postgresql database, for accounting service to store orders receive from kafka. The current change only reflected in docker-compose and not in the kubernetes deployment yet. It will fallback to not use database if there is no corresponding environment variable. So I think it wont break the kubernetes deployment. But if we think it is necessary to have the kubernetes change in this PR as well, I will update later. 

Fixes https://github.com/open-telemetry/opentelemetry-demo/issues/912
## Merge Requirements

essential items:


Kielek:
https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.EntityFrameworkCore used to instrument this package is not yet on stable semantic convention. It is using the old one.

Happy to see contributions to move it forward! : )

docker-compose.yml
```diff
@@ -829,3 +830,15 @@ services:
       timeout: 10s
       retries: 10
     logging: *logging
+
+  postgresql:
```
puckpuck, line 834: Please move this entire section to the "Dependent Services" section of this file. It should be in that section alphabetically based on the service name, so around line 681.

docker-compose.yml
```diff
@@ -829,3 +830,15 @@ services:
...
+    image: 'postgres:latest'
```
puckpuck, line 835: This needs to be based on an environment variable called `POSTGRES_IMAGE`, set up similarly to how we do this for Valkey.

src/accounting/init.sql
puckpuck, line 1: This file should be in a `src/postgres/` folder instead. We would also need to update docker-compose to reference the new location.

puckpuck (changes requested):
I haven't tried to build and run this yet, but some changes that we should do to stay in alignment with how the rest of the Demo is organized.

src/accounting/Consumer.cs
```diff
@@ -38,6 +58,7 @@ public void StartListening()
             {
                 try
                 {
+                    using var activity = MyActivitySource.StartActivity("order-consumed");
```
cuichenli, line 61: somehow the db activity is not automatically closing. so i added this to group the child spans. otherwise, it would result in recursively nested spans like this

cuichenli: also have raised in the dotnet repo https://github.com/open-telemetry/opentelemetry-dotnet-contrib/issues/2792

cuichenli: thanks for pointing out! did not notice npsql also have native instrumentation. have updated accordingly and it is working as expected now 
image

julianocosta89:
@cuichenli, @henrikrexed is working on #2143 that adds dapr to the Demo.
Your PR and his are adding a postgres db to the demo.
We need to decide which one will be.

I'll bring that up in the next SIG meeting.

puckpuck:
Hey @cuichenli sorry it took so long to get back to this.

I did run this PR, and I'm getting 2 traces with only a single span for each order. I would expect to get a single trace with 2 spans in it. Do you know why these 2 spans are not connected?

Screenshot 2025-06-20 at 10 41 25 PM
Screenshot 2025-06-20 at 10 41 39 PM
Screenshot 2025-06-20 at 10 41 56 PM

cuichenli:
image
added one activity and now i can see they are grouped. but this introduces a new parent span. not sure if that is acceptable. 

@puckpuck

Issue #2189, opened 2025-05-16
frontend: Duplicate server spans & high-cardinality span names on next.js spans


cedricziel:
# Bug Report

<img width="1269" alt="Image" src="https://github.com/user-attachments/assets/09888f36-1c6c-430d-987e-9632739a16a2" />

The frontend service produces a semantically incorrect trace where it has 2 kind=server spans AND it produces high-cardinality span names for spans produced by the next.js instrumentation.

## Symptom

**What is the expected behavior?**

1. server span
2. no high-cardinality spans

**What is the actual behavior?**

Please describe the actual behavior experienced.

## Reproduce

Run the demo, observe that calls to `/api/cart` produce server spans with query parameters and there's 2 server spans.

## Additional Context

Please feel free to add any other context about the problem here.

puckpuck:
There are existing rules in the deployed [Otel Collector configuration](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/otel-collector/otelcol-config.yml#L111-L113) to fix the high cardinality spans. This is a known issue with Next.js instrumentation. When you run the demo with `make start`, which uses this Collector config, you will not get the high cardinality span names.

<img width="1588" alt="Image" src="https://github.com/user-attachments/assets/234ae196-9460-4bec-a0cb-7ef07f74e2cd" />

cedricziel:
@puckpuck it still breaks for the bring your own backend when deploying the helm chart or when using a different collector - i dont think its fixed.

puckpuck:
Reading the docs now for the [Demo and Kubernetes](https://opentelemetry.io/docs/demo/kubernetes-deployment/#bring-your-own-backend), I can see why it is not obvious that you need to configure your "bring your own backend" to have certain transformation rules. We only make mention of spanmetrics, but there are also other collector configuration details that need to be applied.

@cedricziel you would need to apply this [transform processor configuration](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/otel-collector/otelcol-config.yml#L106-L113) to your own Collector.

PR #2190, opened 2025-05-16
fix(frontend): reduce cardinality in next.js spans with processor

cedricziel:
# Changes

Mangle spans so the names dont contain high-cardinality segments.

This actively prevents the scenario for the bring-your-own observability deployment type in the helm chart by fixing the span names locally

Partially fixes the concerns in #2189


puckpuck:
As mentioned in this [comment](https://github.com/open-telemetry/opentelemetry-demo/issues/2189#issuecomment-2993239983) on the original issue, this is fixed if you use the default transform rules we have in the Collector. We should keep to this pattern to fix these high cardinality span names.

cedricziel:
it still breaks for the bring your own backend when deploying the helm chart or when using a different collector - i dont think its fixed.

julianocosta89:
@cedricziel I believe we have room for improvement in the documentation of "Bring your own Collector".

In the docs we do mention how merging yaml's work: https://opentelemetry.io/docs/demo/kubernetes-deployment/#bring-your-own-backend.

Any suggestion on how we can make it clearer?

Maybe mentioning the issue and the `transform` processor?

PR #2191, opened 2025-05-16
fix(frontend): switch off the http instrumentation to avoid duplicate spans

cedricziel:
NOTE: this is a followup of #2190 in relation to #2189 

# Changes

Turn off instrumentation for incoming http requests in frontend. This avoids creating duplicate spans.

Needs to be thoroughly validated.


puckpuck:
I agree that having 2 server spans is not ideal, but disabling the HTTP auto instrumentation may not be the ideal situation here. The HTTP instrumented span contains several details that are not part of the Next.js span. 

I do agree with Juliano that this is an instrumentation issue that should be fixed there. Perhaps we should use a collector transform rule to change the span kind for the Next.js spans instead?

PR #2214, opened 2025-05-28
[shipping] actix web refactor

julianocosta89:
# Changes

This PR refactors the Shipping service to use `actix-web`.
This is one of the only instrumentation libraries present on the Rust ecosystem.

That avoids us relying on tracing from Tokio and being always one more versions behind the latest Rust release.
With that we also do not need to manually do context propagation.

With that refactoring I've also added Metrics and Logs.

Logs are already being sent via OTLP.
Metrics still need to be implemented.

Actix-web is a web framework and it doesn't  support gRPC, so I've refactored the checkout service to also call it via HTTP instead.

Tests and health checks in the docker-compose are still missing, but I want to put this PR out.

### TODO

* [x] Add custom metric.
* [x] health check on docker compose
* [x] add tests

## Merge Requirements

essential items:


src/shipping/Cargo.toml
```diff
@@ -1,44 +1,30 @@
...
+log = { version = "0.4.27", features = ["kv"] }
+serde = { version = "1.0.219", features = ["derive"] }
+tonic = "0.12.3"
+
+actix-web-opentelemetry = { version = "0.22.0", features = ["sync-middleware", "awc"] }
+opentelemetry = "0.29.0"
+opentelemetry_sdk = { version = "0.29.0", features = ["metrics", "logs"]}
+opentelemetry-otlp = { version = "0.29.0", features = ["grpc-tonic", "trace", "metrics", "logs"] }
+opentelemetry-appender-log = "0.29.0"
+opentelemetry-resource-detectors = "0.8.0"
```
scottgerring: You can bump this to 0.9.0. 
As you've made a big change with the framework, you might also like to run `cargo-machete` over the top of it to make sure you're not carrying anything in here you don't need

src/shipping/build.rs
```diff
@@ -1,10 +0,0 @@
-// Copyright The OpenTelemetry Authors
```
scottgerring, line 1: Getting rid of a `build.rs` is a nice side-effect

src/shipping/src/main.rs
```diff
@@ -1,43 +1,44 @@
...
+use health_check::health;
+
+#[actix_web::main]
+async fn main() -> std::io::Result<()> {
+    match init_otel() {
+        Ok(_) => {
+            info!("Successfully configured OTel");
+        }
+        Err(err) => {
+            warn!(
```
scottgerring: Where's this going if you can't start OTel - do you need to register some other log appender?

src/shipping/src/main.rs
```diff
@@ -1,43 +1,44 @@
...
-    let (mut health_reporter, health_service) = tonic_health::server::health_reporter();
-    health_reporter
-        .set_serving::<ShippingServiceServer<ShippingServer>>()
-        .await;
-
-    init_logger()?;
-    init_reqwest_tracing(init_tracer()?)?;
-
-    info!("OTel pipeline created");
     let port = env::var("SHIPPING_PORT").expect("$SHIPPING_PORT is not set");
```
scottgerring: You could pull forward the "this should be a `u16`" to make debugging this easier in future. e.g.
```suggestion
    let port: u16 = env::var("SHIPPING_PORT")
        .expect("$SHIPPING_PORT is not set")
        .parse()
        .expect("$SHIPPING_PORT is not a valid port");
```

I would assume at the moment this will panic down around `:42`

src/shipping/src/shipping_service.rs
```diff
@@ -16,167 +10,34 @@ use quote::create_quote_from_count;
...
+pub async fn get_quote(req: web::Json<GetQuoteRequest>) -> impl Responder {
+    let itemct: u32 = req.items.iter().map(|item| item.quantity as u32).sum();
 
-#[cfg(test)]
-mod tests {
-    use super::{
-        shop::shipping_service_server::ShippingService, shop::ShipOrderRequest, ShippingServer,
+    let quote = match create_quote_from_count(itemct).await {
+        Ok(q) => q,
+        Err(_) => return HttpResponse::InternalServerError().finish(),
```
scottgerring: Could return some error details in the response to the client

src/shipping/src/shipping_service/quote.rs
```diff
@@ -50,17 +49,21 @@ async fn request_quote(count: u32) -> Result<f64, Box<dyn std::error::Error>> {
...
+    let mut response = client
         .post(quote_service_addr)
-        .json(&reqbody)
-        .send()
-        .await?
-        .text_with_charset("utf-8")
-        .await?;
+        .trace_request()
+        .send_json(&reqbody)
+        .await
```
scottgerring, line 63: You should use `with_current_context` whenever you are going to pass control into a future, otherwise the other side of the await won't be associated with the proper calling context 

https://docs.rs/opentelemetry/latest/opentelemetry/trace/trait.FutureExt.html#method.with_current_context

src/shipping/src/shipping_service/quote.rs
```diff
@@ -50,17 +49,21 @@ async fn request_quote(count: u32) -> Result<f64, Box<dyn std::error::Error>> {
...
+        .map_err(|err| io::Error::new(io::ErrorKind::Other, err.to_string()))?;
+
+    let bytes = response
+        .body()
+        .await
```
scottgerring, line 68: `with_current_context` here too

src/shipping/src/shipping_service/quote.rs
```diff
@@ -50,17 +49,21 @@ async fn request_quote(count: u32) -> Result<f64, Box<dyn std::error::Error>> {
...
         .post(quote_service_addr)
-        .json(&reqbody)
-        .send()
-        .await?
-        .text_with_charset("utf-8")
-        .await?;
+        .trace_request()
+        .send_json(&reqbody)
+        .await
+        .map_err(|err| io::Error::new(io::ErrorKind::Other, err.to_string()))?;
```
scottgerring: This is weird - an application-level HTTP error doesn't really make sense wrapped into an IO error. You should either use a custom error type for this function and map back to that, or use anyhow. If you use anyhow, you can use it's `context` to add extra details - e.g.:

```
...
.send_json(...)
.await
.context("Failed to call quote service")
```

src/shipping/src/shipping_service/quote.rs
```diff
@@ -50,17 +49,21 @@ async fn request_quote(count: u32) -> Result<f64, Box<dyn std::error::Error>> {
...
+
+    let bytes = response
+        .body()
+        .await
+        .map_err(|err| io::Error::new(io::ErrorKind::Other, err.to_string()))?;
```
scottgerring: same here - shouldn't wrap this into an IO Error.

src/shipping/src/shipping_service/quote.rs
```diff
@@ -50,17 +49,21 @@ async fn request_quote(count: u32) -> Result<f64, Box<dyn std::error::Error>> {
...
+
+    let resp = std::str::from_utf8(&bytes)
+        .map(|s| s.to_owned())
+        .map_err(|err| io::Error::new(io::ErrorKind::Other, err))?;
```
scottgerring: same here - shouldn't wrap this into an IO Error.

src/shipping/src/shipping_service/shipping_types.rs
```diff
@@ -0,0 +1,40 @@
...
+#[derive(Debug, Deserialize)]
+pub struct GetQuoteRequest {
+    pub items: Vec<CartItem>,
+    pub address: Option<Address>,
+}
+
+#[derive(Debug, Serialize)]
+pub struct Money {
+    pub currency_code: String,
+    pub units: i64,
```
scottgerring: Could these both be unsigned?

src/shipping/src/telemetry_conf.rs
```diff
@@ -0,0 +1,85 @@
+// Copyright The OpenTelemetry Authors
```
scottgerring, line 1: I had a comment above saying - if you can't init OTel and you continue, what's your fallback for logs? I reckon you could make sure to add a console appender to the log crate or something in that case, you'll just have to structure the error handling a bit.

src/shipping/Cargo.toml
```diff
@@ -1,44 +1,30 @@
...
-tracing-opentelemetry = "0.23.0"
-tracing-subscriber = "0.3"
+actix-web = "4"
+anyhow = "1.0.98"
+awc = { version = "3.7.0", default-features = false, features = ["compress-zstd"] }
+log = { version = "0.4.27", features = ["kv"] }
+serde = { version = "1.0.219", features = ["derive"] }
+tonic = "0.12.3"
+
+actix-web-opentelemetry = { version = "0.22.0", features = ["sync-middleware", "awc"] }
```
gruebel: if you bump the OTel crates to `0.30.0` then you can also use the migrated version of the `
actix-web-opentelemetry` -> `opentelemetry-instrumentation-actix-web` which is now officially under the OTel umbrella

julianocosta89: Awesome!!!
It wasn't there when I initially sent the PR: https://cloud-native.slack.com/archives/C03GDP0H023/p1748374969278759

I'll bump those

julianocosta89:
> Hey @julianocosta89 hope this is helpful!

Really appreciate your review @scottgerring!
I'll take care of the changes later

src/shipping/src/telemetry_conf.rs
```diff
@@ -0,0 +1,83 @@
...
+                .with_tonic()
+                .build()
+                .expect("Failed to initialize tracing provider"),
+        )
+        .build();
+
+    global::set_tracer_provider(tracer_provider);
+}
+
+fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
```
cijothomas, line 40: lets make sure all 3 signals return the provider from this method, and the same can be used to effect a shutdown when app is shutting down.

src/shipping/src/telemetry_conf.rs
```diff
@@ -0,0 +1,83 @@
...
+        .with_resource(get_resource())
+        .with_batch_exporter(
+            opentelemetry_otlp::LogExporter::builder()
+                .with_tonic()
+                .build()
+                .expect("Failed to initialize logger provider"),
+        )
+        .build();
+
+    let otel_log_appender = OpenTelemetryLogBridge::new(&logger_provider);
```
cijothomas: consider using `tracing` for logs + tracing-appender here.

src/shipping/src/telemetry_conf.rs
```diff
@@ -0,0 +1,83 @@
...
+use opentelemetry_appender_log::OpenTelemetryLogBridge;
+
+use opentelemetry_resource_detectors::{OsResourceDetector, ProcessResourceDetector};
+use opentelemetry_sdk::{
+    propagation::TraceContextPropagator, resource::ResourceDetector, Resource,
+};
+
+use std::{env, str::FromStr};
+
+fn get_resource() -> Resource {
```
cijothomas, line 15: I think we should explicitly add the service.* ones too and not rely on defaults.
 
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs#L23

src/shipping/Cargo.toml
```diff
@@ -1,44 +1,31 @@
...
-tracing-subscriber = "0.3"
+actix-web = "4"
+anyhow = "1.0.98"
+awc = { version = "3.7.0", default-features = false, features = ["compress-zstd"] }
+log = { version = "0.4.27", features = ["kv"] }
+simple_logger = "5.0.0"
+serde = { version = "1.0.219", features = ["derive"] }
+tonic = "0.12.3"
+
+opentelemetry = "0.30.0"
```
cijothomas, line 18: ❤️ Thanks !

src/shipping/Cargo.toml
```diff
@@ -1,44 +1,31 @@
...
+opentelemetry_sdk = { version = "0.30.0", features = ["metrics", "logs"]}
```
cijothomas: nits: No need to list these features - logs and metrics are enabled by default!

src/shipping/src/shipping_service.rs
```diff
@@ -16,167 +10,34 @@ use quote::create_quote_from_count;
...
-                    Err(e) => panic!("error when parsing uuid: {}", e),
-                }
-            }
-            Err(e) => panic!("error when making request for tracking ID: {}", e),
-        }
-    }
+#[post("/ship-order")]
+pub async fn ship_order(_req: web::Json<ShipOrderRequest>) -> impl Responder {
+    let tid = create_tracking_id();
+    info!("Tracking ID Created: {}", tid);
```
cijothomas: I'd recommend to do structured logging wherever possible, with an event_name for every log, and attributes, and message/body if required.

Eg from Otel Rust SDK itself:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/periodic_reader.rs#L175-L177

for this specific case, the following:
```rust
    tracing::info!(name: "TracingId.Created", target: "ShippingService", tracking_id = tid);
```

(if a body/message is preferred, then add message="Tracking ID Created" too, but I don't feel its necessary as the EventName is sufficient to convey that information).

cijothomas: I suggest to ignore the above comment for this PR - you can focus on upgrading to actix, using instrumentation, and upgrade to newest otel in this PR. Fixing individual logs is a perfect candidate for a follow up, and I am happy to send a follow up with that after this is merged!

src/shipping/src/telemetry_conf.rs
cijothomas: If there is an established pattern in the demo, lets stick with that - if telemetry can't be started, then panic and fail fast, as the entire purpose of the app is to demonstrate telemetry and if it can't be enabled, then its better to fail fast.
(If this is not what other services in demo is doing, then ignore. lets stick with existing patterns. If there is no pattern, then my suggestion is fail-fast, rather than continuing without telemetry)

cijothomas (commented):
Overall looks good. Left some suggestions, which are better to be handled in follow up PRs.
Really love seeing demo moving to the latest and greatest of OTel libraries, and instrumentation libraries!

Not marking explicit approval now as PR is marked draft.

src/shipping/src/telemetry_conf.rs
julianocosta89: @cijothomas the main goal of this PR is removing the tracing dependency.

Any reason why I should use it???

src/shipping/src/telemetry_conf.rs
julianocosta89: We are passing OTEL_SERVICE_NAME and OTEL_RESOURCE_ATTRIBUTES as env vars.

We use this approach in all services to avoid hardcoding those values.

src/shipping/src/telemetry_conf.rs
cijothomas: > the main goal of this PR is removing the tracing dependency.

I guess you are removing the idea of creating Otel Spans using `tracing` crate.

For logging, either `log` or `tracing` crate can be used. There are appenders for both in the Otel repo. But most examples maintained in the Otel Rust repo are using `tracing` for logs, so I think its good if demo also uses same, though this is not a hard requirement at all. (Also Otel Rust itself is using `tracing` for its internal logs)

src/shipping/src/telemetry_conf.rs
cijothomas: perfect! Ya I forgot about the env variables.

src/shipping/src/telemetry_conf.rs
scottgerring: While I see your point @cijothomas that this is otherwise reasonably prevalent, IMHO using tracing for structured logging alone, and _explicitly not_ tracing, is confusing and unnecessary. Imagine the comment on the configuration explaining this - "yes we use it, but only for logs - no, please don't use that tracing-opentelemetry crate and certainly don't use tracings tracing macros, because context activation doesn't quite work right yet". 

This is also a chance for us to show folks how to set things up simply!

src/shipping/src/shipping_service.rs
scottgerring: as it ties back to the tracing-v-log thing, you can emit structured logs from the `log` crate as well:

```rust
// We can also use the 'kv' feature on the log crate, and add additional
// structured fields to our logs. 
info!(
    latitude = lat,
    longitude = long;
    "Fetching image"
);
```

src/shipping/src/telemetry_conf.rs
julianocosta89: @cijothomas could you help me out here a bit?

I'm not sure how to structure that in the way the code is today.
Happy to follow up in a next PR as well

cijothomas: lets do it in a follow up to not block this PR. We can copy from the examples in the Otel Rust repo.

src/shipping/src/telemetry_conf.rs
cijothomas: > how to set things up simply!
> using tracing for structured logging alone, and explicitly not tracing, is confusing and unnecessary.

I don't think there is anything about log or tracing that makes one simple or less confusing over another. The examples in Otel Rust shows this already:
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/examples/basic-otlp-http/src/main.rs

My suggestion is to use `tracing`:
1. The Otel appender for `tracing` is already marked stable.
2. It is well-tested and perf-optimized.
3. `tracing` supports "event-name" in its macros and metadata, allowing one to pass event-name like `error!(name: "my-event-name",....rest of structured logging...)`. This aligns with OpenTelemetry’s recommendation that logs include durable event names—essentially [treating logs as “Events”](https://opentelemetry.io/blog/2025/opentelemetry-logging-and-you/).

Points 1 and 2 may be temporary and are within the control of the OTel Rust community, but point 3 is more fundamental. The ability to associate a durable event name with logs is a **strong** reason for me to recommend `tracing`. It’s already used extensively in the OTel Rust SDK’s internal logging (every internal log in OTel Rust has an event-name!), and I’d love to see the demo to also follow the same pattern by using logs as events.

That said, I’ll leave it to the demo maintainers to pick the most suitable logging library.

julianocosta89:
@cijothomas and @scottgerring, I applied all suggestions, would you be able to take a 2nd look?

Shipping has now a custom metric `app.shipping.items_count`:

image

And logs with `tracing`:
Screenshot 2025-06-03 at 18 09 15

I still need to check the Dockerfile issue, but Rust code wise, it should be good to go!

src/checkout/main.go
```diff
@@ -389,14 +390,40 @@ func mustCreateClient(svcAddr string) *grpc.ClientConn {
 }
 
 func (cs *checkout) quoteShipping(ctx context.Context, address *pb.Address, items []*pb.CartItem) (*pb.Money, error) {
-	shippingQuote, err := cs.shippingSvcClient.
-		GetQuote(ctx, &pb.GetQuoteRequest{
-			Address: address,
-			Items:   items})
+	quotePayload, err := json.Marshal(map[string]interface{}{
```
cijothomas, line 393: is it intentional that checkout is also modified?

src/shipping/src/main.rs
```diff
@@ -1,43 +1,46 @@
...
+        }
+    };
+
+    let port: u16 = env::var("SHIPPING_PORT")
+        .expect("$SHIPPING_PORT is not set")
+        .parse()
+        .expect("$SHIPPING_PORT is not a valid port");
+    let addr = format!("0.0.0.0:{}", port);
+    info!(
+        name = "shipping-service",
```
cijothomas:
```suggestion
        name = "ServerStartedSuccessfully",
```

"shipping-service" would be a good fit for target = "shipping.service". (target can be thought of as "InstrumentationScope")

cijothomas: not a blockers, lets do these fixes in follow ups!

src/shipping/src/shipping_service/quote.rs
```diff
@@ -25,19 +21,23 @@ pub async fn create_quote_from_count(count: u32) -> Result<Quote, tonic::Status>
         }
     };
 
+    let meter = global::meter("items_count_meter");
```
cijothomas: for a follow up PR - create meter, counter once and re-use instead of repeatedly creating: https://github.com/open-telemetry/opentelemetry-rust/blob/main/docs/metrics.md#metrics-api

cijothomas (approved):
LGTM.
I have suggested few improvements to the logs,metrics etc to showcase best-practices - but they are better handled separately in follow up PRs!

Thanks for doing this. (If we discover any gaps in instrumentation, we can fix it in the instrumentation itself.)

cijothomas:
@julianocosta89 are we getting metrics ("http.server.duration") also for free from the actix-instrumentation?

src/checkout/main.go
julianocosta89: Yes, Shipping was using tokio and was receiving the call from checkout via grpc.

With this change I had to change the call to go over http

julianocosta89:
> @julianocosta89 are we getting metrics ("http.server.duration") also for free from the actix-instrumentation?

Not sure how to interpret that but yes: 
image

I couldn't find it in Grafana though. 🤔

scottgerring (approved):

src/shipping/src/telemetry_conf.rs
```diff
@@ -0,0 +1,79 @@
...
+                .with_tonic()
+                .build()
+                .expect("Failed to initialize tracing provider"),
+        )
+        .build();
+
+    global::set_tracer_provider(tracer_provider);
+}
+
+fn init_meter_provider() -> opentelemetry_sdk::metrics::SdkMeterProvider {
```
rogercoll, line 40: Do we need to return the meter here? Looks like it is not used in `init_otel`, thus we could avoid the clone

rogercoll (commented):
@julianocosta89 

I am getting these errors on the frontend service:

```
Error: 13 INTERNAL: shipping quote failure: failed to get shipping quote: failed to exit idle mode: invalid target address http://shipping:50050, error info: address http://shipping:50050:443: too many colons in address
    at <unknown> (.next/server/pages/api/checkout.js:1:2769)
    at new Promise (<anonymous>) {
  code: 13,
  details: 'shipping quote failure: failed to get shipping quote: failed to exit idle mode: invalid target address http://shipping:50050, error info: address http://shipping:50050:443: too many colons in address',
  metadata: [Metadata]
}
 ⨯ Error: 13 INTERNAL: shipping quote failure: failed to get shipping quote: failed to exit idle mode: invalid target address http://shipping:50050, error info: address http://shipping:50050:443: too many colons in address
    at <unknown> (.next/server/pages/api/checkout.js:1:2769)
    at new Promise (<anonymous>) {
  code: 13,
  details: 'shipping quote failure: failed to get shipping quote: failed to exit idle mode: invalid target address http://shipping:50050, error info: address http://shipping:50050:443: too many colons in address',
  metadata: [Metadata]
}

```

src/shipping/src/telemetry_conf.rs
cijothomas: we need to return meter_provider and have it saved for doing shutdown at the end. (its not doing shutdown now, needs to do it in a follow up)

julianocosta89:
@rogercoll did you rebuild the checkout service as well?

rogercoll:
> @rogercoll did you rebuild the checkout service as well?

I think so:

```
$ docker rmi -f $(docker images -aq)
$ make build
$ make start
```

Error logs:

```
$ docker logs frontend -f

> frontend@0.1.0 start
> node --require ./Instrumentation.js server.js

   ▲ Next.js 15.2.1
   - Local:        http://6a5fe276152c:8080
   - Network:      http://6a5fe276152c:8080

 ✓ Starting...
 ✓ Ready in 2.3s
Error: 14 UNAVAILABLE: Name resolution failed for target dns:http://shipping:50050
    at <unknown> (Error: 14 UNAVAILABLE: Name resolution failed for target dns:http://shipping:50050)
    at <unknown> (.next/server/pages/api/shipping.js:1:1298)
    at new Promise (<anonymous>) {
  code: 14,
  details: 'Name resolution failed for target dns:http://shipping:50050',
  metadata: [Metadata]
}
```

No logs in the shipping service, but the demo "works".

julianocosta89:
This looks weird `dns:http://shipping:50050`

julianocosta89:
@rogercoll I've removed the custom env var and left the default one.
I've just tested and got no errors:

image

Could you try it out again? 🥺

rogercoll:
> @rogercoll I've removed the custom env var and left the default one. I've just tested and got no errors:
> 
> image
> 
> Could you try it out again? 🥺

Yes, I get those traces on `jaeger` as well but those errors in the frondend proxy logs :man_shrugging: 

Maybe another @open-telemetry/demo-approvers can give it a try? Just to discard an environment issue

julianocosta89:
> Yes, I get those traces on `jaeger` as well but those errors in the frondend proxy logs 🤷‍♂️
> 
> Maybe another @open-telemetry/demo-approvers can give it a try? Just to discard an environment issue

I've managed to get the error!
It only happened when navigating manually in the Demo.

Frontend also calls Shipping and it is trying to use gRPC (which doesn't exist anymore). I'll update it.

Thanks for spotting it @rogercoll!

julianocosta89:
@rogercoll updates are in, do you have some time to take a look?

rogercoll (approved):
Great to see this happening!

Issue #1800, opened 2024-11-28
5xx errors in load generator


agardnerIT:
# Bug Report

```
helm list -A
my-otel-demo    ......   deployed    opentelemetry-demo-0.33.4     1.12.0
```

## Symptom

Lots and lots of 5xx errors.

```
Failed to load resource: the server responded with a status of 503 (Service Unavailable)
Failed to load resource: net::ERR_CONNECTION_REFUSED
Failed to load resource: net::ERR_CONNECTION_REFUSED
Failed to load resource: net::ERR_CONNECTION_REFUSED
Failed to load resource: net::ERR_CONNECTION_REFUSED
Failed to load resource: the server responded with a status of 503 (Service Unavailable)
```

**What is the expected behavior?**

No 5xx errors

What do you expect to see?

**What is the actual behavior?**

Please describe the actual behavior experienced.

## Reproduce

Could you provide the minimum required steps to resolve the issue you're seeing?

We will close this issue if:

* The steps you provided are complex.
* If we can not reproduce the behavior you're reporting.

## Additional Context

```
kubectl -n <REDACTED> describe pod/my-otel-demo-loadgenerator-867c949bd7-dcgtt
Name:             my-otel-demo-loadgenerator-867c949bd7-dcgtt
Namespace:        <REDACTED>
Priority:         0
Service Account:  my-otel-demo
Start Time:       Thu, 28 Nov 2024 14:16:51 +1000
Labels:           app.kubernetes.io/component=loadgenerator
                  app.kubernetes.io/instance=my-otel-demo
                  app.kubernetes.io/name=my-otel-demo-loadgenerator
                  opentelemetry.io/name=my-otel-demo-loadgenerator
                  pod-template-hash=867c949bd7
Annotations:      <none>
Status:           Running
IP:               <REDACTED>
IPs:
  IP:           <REDACTED>
Controlled By:  ReplicaSet/my-otel-demo-loadgenerator-867c949bd7
Containers:
  loadgenerator:
    Container ID:   containerd://344df4921a5b3b84577b3a85bed54f9c6db1ef1547c5f46a47c6798fbc24912b
    Image:          ghcr.io/open-telemetry/demo:1.12.0-loadgenerator
    Image ID:       ghcr.io/open-telemetry/demo@sha256:85c9935ff31b7ab575903fbd0b56a3161ec13e508966df25dc68fcfe7af5ec98
    Port:           8089/TCP
    Host Port:      0/TCP
[...]

PR #2374, opened 2025-07-21
fix: remove nonexistent `wait_until` param from playwright page API call

brylee10:
# Changes

The Python Playwright 1.53 `page.click` api does not have the `wait_until` parameter (docs: https://devdocs.io/playwright/api/class-page#page-click). Instead, it provides a `wait_for_load_state` function (https://devdocs.io/playwright/api/class-page#page-wait-for-load-state). This adds the `wait_for_load_state` calls. Without these calls, the API does not wait for the corresponding browser events, as intended.


julianocosta89 (approved):
Thanks for that!

It looks like this fixes #1800

brylee10:
Yes I just updated the `CHANGELOG.md`

PR #2413, opened 2025-07-31
Fix IDs  in cart logging

shantanualsi:
# Changes

Logging IDs in `cart` has been broken for a while. This PR fixes the templating to log IDs properly.
BEFORE:
image

AFTER: 
image

## Merge Requirements

essential items:

### DIFF
--- a/src/cart/src/cartstore/ValkeyCartStore.cs
+++ b/src/cart/src/cartstore/ValkeyCartStore.cs
@@ -124,7 +124,7 @@ public class ValkeyCartStore : ICartStore
     public async Task AddItemAsync(string userId, string productId, int quantity)
     {
         var stopwatch = Stopwatch.StartNew();
-        _logger.LogInformation("AddItemAsync called with userId={userId}, productId={productId}, quantity={quantity}", userId, productId, quantity);
+        _logger.LogInformation($"AddItemAsync called with userId={userId}, productId={productId}, quantity={quantity}");
 
         try
         {
@@ -173,7 +173,7 @@ public class ValkeyCartStore : ICartStore
 
     public async Task EmptyCartAsync(string userId)
     {
-        _logger.LogInformation("EmptyCartAsync called with userId={userId}", userId);
+        _logger.LogInformation($"EmptyCartAsync called with userId={userId}");
 
         try
         {
@@ -193,7 +193,7 @@ public class ValkeyCartStore : ICartStore
     public async Task<Oteldemo.Cart> GetCartAsync(string userId)
     {
         var stopwatch = Stopwatch.StartNew();
-        _logger.LogInformation("GetCartAsync called with userId={userId}", userId);
+        _logger.LogInformation($"GetCartAsync called with userId={userId}");
 
         try
         {


julianocosta89 (approved):
Thx for that @shantanualsi!

martinjt:
This is actually an incorrect change and needs reverting. I'll see if I can get that change submitted.

Logging in .NET should not use String Interpolation for a variety of reasons.

What's happening in your example here is that you're using the `body` field directly in your output, instead of interpolating the fields during the output. You can configure OpenTelemetry to do the interpolation, how you should never do it inline in the code.

Here's an output if you use interpolated strings:

```
LogRecord.Timestamp:               2025-11-20T16:01:29.3958205Z
LogRecord.CategoryName:            Program
LogRecord.Severity:                Info
LogRecord.SeverityText:            Information
LogRecord.Body:                    User b8559e9a-8c39-4576-875b-bfb840cfaad9 logged in
LogRecord.Attributes (Key:Value):
    OriginalFormat (a.k.a Body): User b8559e9a-8c39-4576-875b-bfb840cfaad9 logged in
```

Whereas this is what happens when you run it with properties:

```
LogRecord.Timestamp:               2025-11-20T16:01:29.3785160Z
LogRecord.CategoryName:            Program
LogRecord.Severity:                Info
LogRecord.SeverityText:            Information
LogRecord.Body:                    User {UserId} logged in, template
LogRecord.Attributes (Key:Value):
    UserId: b8559e9a-8c39-4576-875b-bfb840cfaad9
    OriginalFormat (a.k.a Body): User {UserId} logged in, template
```

You'll notice that the `UserId` property is included in the attributes so that it becomes searchable/filterable.

You can set `logging.IncludeFormattedMessage` on the logging so you can get that string:

```
LogRecord.CategoryName:            Program
LogRecord.Severity:                Info
LogRecord.SeverityText:            Information
LogRecord.FormattedMessage:        User 773e24d5-06f6-4580-9986-dc7d2482f186 logged in, template
LogRecord.Body:                    User {UserId} logged in, template
LogRecord.Attributes (Key:Value):
    UserId: 773e24d5-06f6-4580-9986-dc7d2482f186
    OriginalFormat (a.k.a Body): User {UserId} logged in, template
```

The body field will always include the message template rather than the full message.

There may be logging backends for `ILogger` that treat things differently, however, this is how OpenTelemetry expects it to work, and also how Microsoft/.NET designed it to work.

martincostello:
> Logging in .NET should not use String Interpolation for a variety of reasons.

There's even an analyzer that should flag any usage to stop it being introduced in the first place: [CA2254: Template should be a static expression](https://learn.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca2254)

Ideally CI here would have flagged this and failed the build and prevented the merge in the first place.

julianocosta89:
> Ideally CI here would have flagged this and failed the build and prevented the merge in the first place.

@martincostello would you be able to add a CI task to prevent those types of changes?
I reckon we do not have any .NET expert as a Demo maintainer/approver so it is tricky to know what are the best practices/recommendations for all programming languages that the Demo supports.

martincostello:
The referenced PR above should deal with it, the build for the C# projects would fail and thus block CI.

PR #2414, opened 2025-07-31
Added missing Trace propagation to accounting

hagen-p:
#Reason for change
 In its current 2.02 version, accounting  service is not using RPC trace context propagation, causing every message to become a new trace. 
 
image

After implementing  an extract on each message consume,  the accounting span is now correctly  connected  with the Checkout service via  Kafka   using RPC trace context propagation.

image

# Changes

Added the OpenTelemetry.api to Accounting.csproj
Extended the Modified StartListening loop to extract the trace header info and add it to the parent context 
Updating ChangeLog.md with relevant info

#Doc Changes
Raising PR on the Demo documentation  to indicate that accounting now has RPC Context Propagation
https://opentelemetry.io/docs/demo/telemetry-features/trace-coverage/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ process content of "Unreleased" section content will generate release notes for
 the release.
 
 ## Unreleased
-
+* [accounting] Added Trace extract on Kafka message for correlation
+  ([#2414](https://github.com/open-telemetry/opentelemetry-demo/pull/2414))
 * [chore] add GOMEMLIMIT to all Go services
   ([#2148](https://github.com/open-telemetry/opentelemetry-demo/pull/2148))
 * [product-catalog] Simplify span event name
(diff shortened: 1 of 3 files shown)


hagen-p:
Left the original lines of code consuming the Kafka message  commented out for reference.
Added the actual PR number to CHANGELOG.md

julianocosta89:
Hello @hagen-p the current propagation is actually expected.

Accounting and Fraud-Detection services do not have a parent-child relationship with checkout.

Checkout sends a message to kafka.

Accounting and Fraud-Detection read that message and create a new trace with a span link to the checkout span.

Here is what the OTel messaging semantic convention says:

https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#consumer-spans

> A single “Process” or “Receive” span can account for a single message, for a batch of messages, or for no message at all (if it is signalled that no messages were received). For each message it accounts for, the “Process” or “Receive” span SHOULD link to the message’s creation context.

hagen-p:
@julianocosta89  as you can see from the screen shots, Fraude detection did already correctly correlate based on the Kafka message. and connected it correctly to the parent span in the Trace. I have not touched that service, just made sure accounting is behaving in the same way.

I added the same functionality  to Accounting so that the full end to end interaction is visible in the trace. Also showing proper  trace correlation across the Kafka message bus
As this is a key feature for  Kafka Users, to me it make sense to have it part of the demo

Also, The Checkout service was already  Injecting the Trace context in the Kafka  messages to enable context propagation

hagen-p:
The Span link was no mentioned as part of the demo documentation,  hence my  PR,   I Think that is an omission
Anyway will close and  remove. Pr

PR #2496, opened 2025-09-02
otel_demo config

adinazugravescu:


julianocosta89:
Hello @adinazugravescu 👋🏽 

Looks like this PR was opened on the wrong repo.
In the upstream we try to keep all tools as vendor agnostic as possible.
There is a Grafana fork that uses the Grafana stack: https://github.com/grafana/opentelemetry-demo
Maybe check with them if it would make sense to contribute on their fork.

If you think this PR should be here, please feel free to re-open it, and provide a bit of context to the contribution.

Thanks again 😊

PR #2503, opened 2025-09-03
[shipping] Use cumulative metrics in shipping service to be consistent with the other services of the demo

cyrille-leclerc:
# Changes

Use cumulative metrics in shipping service to be consistent with the other services of the demo.

Note that Prometheus can't ingest delta metrics and we have an exception in the collector when using delta metrics.

I checked, metrics produced by the shipping metrics like `http.server.duration` or `app.shipping.items_count` work well in PRometheus when produced as cumulative metrics.

## Merge Requirements

essential items:

### DIFF
--- a/docker-compose.minimal.yml
+++ b/docker-compose.minimal.yml
@@ -494,9 +494,10 @@ services:
     environment:
       - SHIPPING_PORT
       - QUOTE_ADDR
-      - OTEL_EXPORTER_OTLP_ENDPOINT=http://${OTEL_COLLECTOR_HOST}:${OTEL_COLLECTOR_PORT_GRPC}
+      - OTEL_EXPORTER_OTLP_ENDPOINT
       - OTEL_RESOURCE_ATTRIBUTES
       - OTEL_SERVICE_NAME=shipping
+      - OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE
     healthcheck:
       test: ["CMD-SHELL", "timeout 1 bash -c '>/dev/tcp/localhost/${SHIPPING_PORT}'"]
       start_period: 10s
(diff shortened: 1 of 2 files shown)


cyrille-leclerc:
@julianocosta89 I think that you are the author of the piece of code that specify to use Delta temporality in the Shipping service instead of the default cumulative temporality. Do you remember the rationale?
I don't see problems testing with cumulative, is there a challenge somewhere?

See commit diff https://github.com/open-telemetry/opentelemetry-demo/commit/00142af74736a421022f6aa00d5d23089d52b23e#diff-6e0dcf073aecdc8a72dcc3dc9264e65fe0b91c859b4017bde948bf48e90a72fdR45

julianocosta89:
hey @cyrille-leclerc thanks for that!
The rationale is that I simply forgot about it 😅  haha.

The temporality should be configured via env var and not via code.
I'll add to your PR the `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` on the `shipping` service.

Thanks for raising this

cyrille-leclerc:
Thanks @julianocosta89,
FYI I didn't mention it in this PR but I'm looking with @ywwg at making it easy to switch the entire OTel Demo from cumulative to delta in order to test the ongoing work of the Prometheus community to support delta temporality. We will be happy to pair with you on this topic as we see this is also in your radar.
An item I have in mind for example is to switch OTel Collector infrastructure monitoring receivers like HostMetrics or PostgreSQL receivers to delta temporality, I don't know if the `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` env var would work for this.

julianocosta89:
@cyrille-leclerc the `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` should do what you are trying to achieve.

The whole demo should be able to be configured just with this env var to switch from cumulative to delta.

julianocosta89 (approved):

PR #2526, opened 2025-09-09
Refactor cart service to use Swift

slashmo:


slashmo:
Apologies for the spam, I accidentally targeted `upstream` instead of my fork.

julianocosta89:
@slashmo no worries.
I wonder if that is working?

I'd love to discuss this and maybe add a swift service to the demo.

slashmo:
> I'd love to discuss this and maybe add a swift service to the demo.

I'd be more than happy to contribute it upstream as well, yes! The only caveat is that it'd be using [swift-otel/swift-otel](https://github.com/swift-otel/swift-otel) instead of [open-telemetry/opentelemetry-swift](https://github.com/open-telemetry/opentelemetry-swift) because it's a bit more geared towards server-side Swift.

julianocosta89:
Interesting 🤔 
I didn't know about this swift-otel repo at all.
Is there any reason this work is not done on upstream?

slashmo:
The main reason back then when I first created it is that it was coming from a very different angle compared to the upstream. The upstream SDK is a full-on OTel SDK and at least then very much focused on iOS, not cross-platform Swift. Nowadays, the difference is that Swift OTel is pretty much just an OTLP exporter for Swift's [logging](https://github.com/apple/swift-log), [metrics](https://github.com/apple/swift-metrics), and [tracing](https://github.com/apple/swift-distributed-tracing) libraries. I think the upstream has some of these integrations as well now but it still provides it's own API in addition to them. The server-side Swift ecosystem usually depends on these three libraries directly instead of using the OTel SDK, which means as an end-user you get telemetry exported from them for free when using Swift OTel or other implementations of these libraries' protocols.

julianocosta89:
@slashmo I've added this thread to our agenda to discuss it on our next SIG meeting  (Sept 24).
Feel free to join us, if you'd like!

PR #2550, opened 2025-09-16
Update Cart Service to use AzMon OTel Distro

mattmccleary:
This should be all the changes required to switch the cart service to using the AzMOn OTel Distro


mattmccleary:
My error, this should've been in the Azure fork.

PR #2577, opened 2025-09-22
[flagd-ui] fix application start order for mix release

dottorblaster:
Sometimes the Flagd-UI doesn't wait for the OTEL exporter erlang application to start properly. As per official documentation, this should fix the issue.


julianocosta89:
Thank you @dottorblaster !

PR #2632, opened 2025-10-07
WIP: native otel names

ywwg:
DO NOT MERGE

This is a brute-force rename of all the metrics to their native otel equivalents.  There may still be errors.  For testing only

Signed-off-by: Owen Williams <owen.williams@grafana.com># Changes

### DIFF
--- a/src/grafana/provisioning/alerting/cart-service-alerting.yml
+++ b/src/grafana/provisioning/alerting/cart-service-alerting.yml
@@ -78,11 +78,11 @@ groups:
             The 95th percentile response time for operation {{ index $labels "service.namespace"
                 }}/{{ index $labels "service.name" }} "{{ index $labels "http.request.method" }} {{
                 index $labels "http.route" }}" has been
-                above xxx seconds for 2 minutes on {{ index $labels "service.instance.id"}}. Current
+                above 0.0001 seconds for 2 minutes on {{ index $labels "service.instance.id"}}. Current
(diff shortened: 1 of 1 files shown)


julianocosta89:
@ywwg do you want to keep this open?
Should we close it and re-open later?
Should we review it?

Let us know

ywwg:
I do not intend to merge this, so no review is needed.  I think the right way to support native names would be to come up with a general way of generating these dashboard files with code, and allowing the user to select which translation scheme to use at code gen time. Do you think that is a practical approach for this repo?

julianocosta89:
> I do not intend to merge this, so no review is needed. I think the right way to support native names would be to come up with a general way of generating these dashboard files with code, and allowing the user to select which translation scheme to use at code gen time. Do you think that is a practical approach for this repo?

I do not understand what you mean.

Actually, looking at the PR, I'm not quite sure I get what you are trying to solve here.
Is this a dashboard visualization issue?

AFAIK all dashboards are working, no?!

ywwg:
Because we have customers who want to make everything "otel native", including the naming conventions, we are exploring ways to support them. I am testing the flow of not translating otel metric names to prometheus style, which is an option that otel+prometheus now support. So I wanted to see how many changes were needed to the dashboards to support the original otel metric and label names (names with dots and without suffixes). There is still an open question about how to support both types of customers, because we see about a 50/50 split between those who are comfortable with prometheus style and those who want the newer otel style.

PR #2158, opened 2025-04-28
Genai demo

aabmass:
# Changes

Fixes https://github.com/open-telemetry/opentelemetry-demo/issues/2156

## Merge Requirements

For new features contributions, please make sure you have completed the following
essential items:

* [ ] `CHANGELOG.md` updated to document new feature additions
* [ ] Appropriate documentation updates in the [docs][]
* [ ] Appropriate Helm chart updates in the [helm-charts][]

Maintainers will not merge until the above have been completed. If you're unsure
which docs need to be changed ping the
[@open-telemetry/demo-approvers](https://github.com/orgs/open-telemetry/teams/demo-approvers).

[docs]: https://opentelemetry.io/docs/demo/
[helm-charts]: https://github.com/open-telemetry/opentelemetry-helm-charts

PR #2663, opened 2025-10-17
Add a Product Review service with GenAI-powered summaries

dmitchsplunk:
# Changes

This PR proposes adding Customer Product Reviews to the Astronomy Shop application, using Generative AI to summarize the reviews for each product.  This addition will allow the community to demonstrate OpenTelemetry capabilities for instrumenting Generative AI interactions within the Astronomy Shop application. 

Summary of changes: 

* Adds a new Python-based Product Review service with two functions: getProductReviews(productId) and getProductReviewSummary(productId).
* Introduces a Python-based LLM service that mocks OpenAI’s Chat Completions API to generate AI summaries of product reviews.
* Stores customer reviews in a MySQL database.
* Updates the front-end product page with a new Reviews section, including an AI-generated summary and individual reviews.
* Instruments GenAI interactions using opentelemetry-instrumentation-openai-v2 to capture relevant spans and attributes.

Here's a screenshot of the new Customer Reviews section of the product page: 

Astronomy Shop - Product Reviews

And here's an example trace showing the Product Review Summary flow: 

Product Review Summary Trace

The LLM service supports two new feature flags: 

* `llmInaccurateResponse`: when this feature flag is enabled the LLM service returns an inaccurate product summary for product ID L9ECAV7KIM
* `llmRateLimitError`: when this feature flag is enabled, the LLM service intermittently returns a RateLimitError with HTTP status code 429

If the direction looks good, I’ll follow up with documentation and Helm chart changes. In the meantime, I’d welcome early feedback.

## Merge Requirements

essential items:

* [x] Appropriate documentation updates in the [docs][] -> [Docs PR](https://github.com/open-telemetry/opentelemetry.io/pull/8294)
* [x] Appropriate Helm chart updates in the [helm-charts][] -> [Helm chart PR](https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1920)

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -14,7 +14,7 @@ permissions:
 
 jobs:
   build_images:
-    permissions: # required by the reusable workflow
+    permissions:  # required by the reusable workflow
       contents: read
       packages: write
     uses: ./.github/workflows/component-build-images.yml
--- a/src/frontend-proxy/envoy.tmpl.yaml
+++ b/src/frontend-proxy/envoy.tmpl.yaml
@@ -61,7 +61,7 @@ static_resources:
                             prefix_rewrite: "/"
                             upgrade_configs:
                               - upgrade_type: websocket
-                        - match: { prefix: "/" } # Default/catch-all route - keep last since prefix:"/" matches everything
+                        - match: { prefix: "/" }  # Default/catch-all route - keep last since prefix:"/" matches everything
                           route: { cluster: frontend }
                 http_filters:
                   - name: envoy.filters.http.fault
--- a/test/tracetesting/shipping/empty-quote.yaml
+++ b/test/tracetesting/shipping/empty-quote.yaml
@@ -26,14 +26,14 @@ spec:
           "items": []
         }
   specs:
-  - name: It called GetQuote successfully
-    selector: span[tracetest.span.type="http" name="/get-quote" http.request.method="POST"]
-    assertions:
-    - attr:http.response.status_code = 200
-  - name: It returned a valid empty quote
-    selector: span[tracetest.span.type="general" name="Tracetest trigger"]
-    assertions:
-    - attr:tracetest.response.status = 200
-    - attr:tracetest.response.body | json_path '$.cost_usd.currency_code' = "USD"
-    - attr:tracetest.response.body | json_path '$.cost_usd.units' = 0
-    - attr:tracetest.response.body | json_path '$.cost_usd.nanos' = 0
+    - name: It called GetQuote successfully
+      selector: span[tracetest.span.type="http" name="/get-quote" http.request.method="POST"]
+      assertions:
+        - attr:http.response.status_code = 200
+    - name: It returned a valid empty quote
+      selector: span[tracetest.span.type="general" name="Tracetest trigger"]
+      assertions:
+        - attr:tracetest.response.status = 200
+        - attr:tracetest.response.body | json_path '$.cost_usd.currency_code' = "USD"
+        - attr:tracetest.response.body | json_path '$.cost_usd.units' = 0
+        - attr:tracetest.response.body | json_path '$.cost_usd.nanos' = 0
--- a/test/tracetesting/shipping/order.yaml
+++ b/test/tracetesting/shipping/order.yaml
@@ -31,12 +31,12 @@ spec:
           ]
         }
   specs:
-  - name: It called ShipOrder successfully
-    selector: span[tracetest.span.type="http" name="/ship-order" http.request.method="POST"]
-    assertions:
-    - attr:http.response.status_code = 200
-  - name: It returned a tracking ID
-    selector: span[tracetest.span.type="general" name="Tracetest trigger"]
-    assertions:
-    - attr:tracetest.response.status = 200
-    - attr:tracetest.response.body | json_path '$.tracking_id' != ""
+    - name: It called ShipOrder successfully
+      selector: span[tracetest.span.type="http" name="/ship-order" http.request.method="POST"]
+      assertions:
+        - attr:http.response.status_code = 200
+    - name: It returned a tracking ID
+      selector: span[tracetest.span.type="general" name="Tracetest trigger"]
+      assertions:
+        - attr:tracetest.response.status = 200
+        - attr:tracetest.response.body | json_path '$.tracking_id' != ""
--- a/test/tracetesting/shipping/quote.yaml
+++ b/test/tracetesting/shipping/quote.yaml
@@ -31,13 +31,13 @@ spec:
           ]
         }
   specs:
-  - name: It called GetQuote successfully
-    selector: span[tracetest.span.type="http" name="/get-quote" http.request.method="POST"]
-    assertions:
-    - attr:http.response.status_code = 200
-  - name: It returned a valid quote
-    selector: span[tracetest.span.type="general" name="Tracetest trigger"]
-    assertions:
-    - attr:tracetest.response.status = 200
-    - attr:tracetest.response.body | json_path '$.cost_usd.currency_code' = "USD"
-    - attr:tracetest.response.body | json_path '$.cost_usd.units' > 0
+    - name: It called GetQuote successfully
+      selector: span[tracetest.span.type="http" name="/get-quote" http.request.method="POST"]
+      assertions:
+        - attr:http.response.status_code = 200
+    - name: It returned a valid quote
+      selector: span[tracetest.span.type="general" name="Tracetest trigger"]
+      assertions:
+        - attr:tracetest.response.status = 200
+        - attr:tracetest.response.body | json_path '$.cost_usd.currency_code' = "USD"
+        - attr:tracetest.response.body | json_path '$.cost_usd.units' > 0


julianocosta89:
Hello @dmitchsplunk 👋🏽 

this looks cool!
Would you be able to join the SIG meeting this Wed?
Here are the SIG meeting notes: https://docs.google.com/document/d/16f-JOjKzLgWxULRxY8TmpM_FjlI1sthvKurnqFz9x98/edit?tab=t.0#heading=h.b9liq2yzxczd

I've added an entry to discuss this PR.
My main concern about it is the fact that this is just for OpenAI.
We are a vendor agnostic project and we try to keep it like that, I think here it is the same problem we are having with adding a serverless service to the demo.

We had a PR from @aabmass some time ago (but it was a draft PR and closed due to inactivity) that also added a GenAI service: https://github.com/open-telemetry/opentelemetry-demo/pull/2158

dmitchsplunk:
Thanks @julianocosta89!  Happy to join the SIG meeting on Wednesday to discuss this PR. 

I understand the need to keep things vendor agnostic, and I chose to use the OpenAI API format because most LLM providers offer OpenAI-compatible endpoints.  And while Anthropic and Google have their own API formats, the industry trend is toward LLMs either using or supporting OpenAI's API format.

dmitchsplunk:
Hi @julianocosta89 - thanks for the discussion earlier today and for your support on this PR.  I've made the following changes: 

1. The updates to `kubernetes/opentelemetry-demo.yaml` have been reverted (since they're auto-generated from the Helm chart). 

2. The Product Reviews service now uses the existing Postgres database.  I introduced new `accounting` and `reviews` schemas, and updated the Accounting service to reference the new schema name.  The MySQL database I added earlier has been removed. 

3. I updated the solution to ensure it works with a real OpenAI API compatible LLM.  This is driven via configuration.  In the `.env` file we point to the mock LLM service by default: 

````
# LLM service
LLM_DOCKERFILE=./src/llm/Dockerfile
LLM_HOST=llm
LLM_PORT=8000

# Mock LLM - to use a real OpenAI API-compatible LLM, comment the next 3 lines
LLM_BASE_URL=http://${LLM_HOST}:${LLM_PORT}/v1
LLM_MODEL=astronomy-llm
OPENAI_API_KEY=dummy
````

If we want to use OpenAI instead, we would change the config to: 

````
LLM_BASE_URL=https://api.openai.com/v1
LLM_MODEL=gpt-4o-mini
OPENAI_API_KEY=<replace with API key>
````

I'll work on the Helm chart updates next.

**Comment by julianocosta89** [comment]
@dmitchsplunk I took a look at the Ui and I didn't quite understand it.
I know that it is all mockup data, but maybe we could make it more accurate.

One of the products has a rating of 1.8, but 80% of the ratings are 5 stars and 20% 4 stars 😅 
image


Another thing that I I'd rather have is some way to actually interact with the chat.
Currently the summary and the chat call is done without any user interaction.
Can we change that?
WDYT?

**Comment by dmitchsplunk** [comment]
Hi @julianocosta89, the 1.8 star review summary is intentionally inaccurate, and is controlled by the following feature flag: 

`llmInaccurateResponse`: when this feature flag is enabled the LLM service returns an inaccurate product summary for product ID L9ECAV7KIM

It was intended to allow observability platforms to demonstrate how they evaluate the quality and accuracy of LLM responses, but perhaps it should be disabled by default to avoid confusion.

**Comment by dmitchsplunk** [comment]
I think user interaction with the chat would be a valuable feature. For example, we could add a chat interface that allows customers to ask questions like: 

- What kind of telescope would you recommend for a child? 
- What products do you have with ratings of 4.5 stars of higher? 
- What was the total amount of order xyz? 

However, I think this functionality goes beyond the original scope of the PR, which was focused on adding product reviews and AI-generated summaries.  So I'd suggest we address this as a follow-up PR, if possible.

**Comment by julianocosta89** [comment]
@dmitchsplunk thx for the reply!
I didn't pay attention to the feature flag as yes, the default is always off 😅 

Regarding the implementation I now have mixed feelings about it.
Having a call to the AI every time we open a product page seems unnecessary.
I was thinking. An easy solution would be a button, that the user can click to get the summary, but TBH I'd love to have something more interactive.

IDK if you have ever seen the Rufus AI, the AI Amazon has in their product pages.
It's pretty simple.
- No conversation state management - Each question is independent
- No message history - Simpler UI and backend
- Product-scoped context - Only need current product data
- Cleaner UX - Users see one Q&A at a time

I know this is totally different from your PR, but I'd love to hear your opinion.

**Comment by dmitchsplunk** [comment]
@julianocosta89 this is a great idea.  I've started working this into the code, and here's what it looks like so far: 

Customer Reviews with Interactive AI

The average score can be calculated without AI, so we can always display that, just above the product reviews themselves.  

But now, if the user wants to see a summary of product reviews, they'll have to request it.  

Since the LLM will be a mock in most cases, there are only a few questions that it will know how to answer.  

Please let me know if this is aligned with your thinking, and I'll continue with the implementation.

**Comment by julianocosta89** [comment]
@dmitchsplunk I loved that!

**Comment by dmitchsplunk** [comment]
Hi @julianocosta89 - I've completed the implementation.  When the mock LLM is used, it will respond to the three fixed questions, but will say it doesn't know the answer for ad-hoc questions.  When a real LLM is used, it has the ability to fetch additional product info and the product reviews, and will answer whatever questions it can with that info.  Please try it out when you have a chance. 

Product AI Assistant

**Comment by julianocosta89** [comment]
@dmitchsplunk thank you!
I'm flying today, but I'll take a look whenever I have a couple of minutes.

Really nice addition to the demo!
Im excited to test it out! 🥳

**Comment by julianocosta89** [comment]
@dmitchsplunk I've updated your PR with a fix to work with OpenAI.
I've used Claude to fix it, and I've tested with and without OpenAI.

### The Problem
The original code only handled the first tool call (tool_calls[0]), but when OpenAI's API returns multiple tool calls (e.g., both `fetch_product_reviews` and `fetch_product_info`), you must provide a response for each `tool_call_id`. The API was rejecting your request because it was missing responses for the additional tool calls.

### The Solution
The updated code now:
- Processes all tool calls in a loop instead of just the first one
- Appends the assistant's message once before processing any tool calls
- Appends a tool response for each tool call with the correct tool_call_id
- Consolidates the final user prompt to avoid duplication based on which tool was called
- Makes a single final LLM call with all tool results included

### Key Changes
- Changed from `tool_call = tool_calls[0]` to `for tool_call in tool_calls`:
- Moved `messages.append(response_message)` outside the loop so it's only added once
- Each tool call now gets its response appended to the messages array
- Simplified the flow to eliminate redundant code paths for different tool types

Screenshot 2025-10-29 at 18 12 29

Screenshot 2025-10-29 at 18 14 07

**Comment by julianocosta89** [comment]
I've also moved the configuration of OpenAI token to `.env.override`.
With that, users do not need to keep comment/uncomment code.

If they want to use OpenAI, they can uncomment the override file and that will take care of overriding the values from .env file.

File: src/flagd/demo.flagd.json [hunk 1]
```diff
@@ -1,6 +1,24 @@
 {
   "$schema": "https://flagd.dev/schema/v0/flags.json",
   "flags": {
+    "llmInaccurateResponse": {
```
julianocosta89, line 4: This is just implemented to work with the fake llm service.
Not sure if we can force an hallucination with a public API llm🤔

dmitchsplunk: Right, the feature flag is used by the mock LLM service.  It may be possible to change the system prompt to tell OpenAI to make something up.

dmitchsplunk:
@julianocosta89 thanks for the fix for the multiple tool calls, the questions I was testing with must have resulted in single tool calls only.  

I made a small change to the mock LLM service to ensure it still returns product reviews successfully.

dmitchsplunk:
> I've also moved the configuration of OpenAI token to `.env.override`. With that, users do not need to keep comment/uncomment code.
> 
> If they want to use OpenAI, they can uncomment the override file and that will take care of overriding the values from .env file.

Is there anything else I need to do?  I tried updating .env.override and restarting the app with docker compose, but it didn't pick up the changes.

src/flagd/demo.flagd.json
dmitchsplunk: Or, when the feature flag is activated, and a review for that specific product is requested, we could direct the request to the mock LLM service.

dmitchsplunk:
> > I've also moved the configuration of OpenAI token to `.env.override`. With that, users do not need to keep comment/uncomment code.
> > If they want to use OpenAI, they can uncomment the override file and that will take care of overriding the values from .env file.
> 
> Is there anything else I need to do? I tried updating .env.override and restarting the app with docker compose, but it didn't pick up the changes.

Please disregard, I got it working with the following command: 

`docker compose --env-file .env --env-file .env.override up --force-recreate --remove-orphans --detach --build `

dmitchsplunk:
Hi @julianocosta89 , I've created draft PRs for the Helm chart and documentation updates: 

https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1920
https://github.com/open-telemetry/opentelemetry.io/pull/8294

Please let me know if there any other changes you'd like me to make as part of this PR.

julianocosta89:
Hello @dmitchsplunk I was at a conference the past week, but I'll be able to take a look latest next week.

I'm still concerned about the feature flags we are introducing with this PR, as they won't work with OpenAI.

We should either remove the feature flag, or think about a different way to illustrate that with OpenAI.

dmitchsplunk:
Hi @julianocosta89 - I've updated the service so that the feature flags work even when the real OpenAI API is used. 

For the llmRateLimitError feature flag, I changed the logic such that, when the feature flag is active and the random number is less than 0.5, then the mock LLM is always called, even if the app is configured to use the real OpenAI API.  The mock LLM then returns the rate limit error (HTTP status code 429).  The UI will simply show that the system is unable to process the request at this time. 

429 Error in UI

429 Rate Limit Error

For the llmInaccurateResponse feature flag, I modified the prompt sent to OpenAI when this feature flag is enabled to ask it to return an inaccurate response.  It works well with my testing, please let me know what you think.

julianocosta89 (approved):

julianocosta89:
@dmitchsplunk I've spent some time fixing the tracetest and finally got it to work.
Not because of your changes, on yours I just had to update the method name.

The rest was all on me

dmitchsplunk:
Thanks for all your help in getting this merged @julianocosta89 !  I'll go ahead and move the Helm chart and docs PRs out of draft stage:  

https://github.com/open-telemetry/opentelemetry-helm-charts/pull/1920
https://github.com/open-telemetry/opentelemetry.io/pull/8352

PR #2710, opened 2025-11-05
OpenShift version

maximilianoPizarro:
# Changes

## Merge Requirements

essential items:


julianocosta89:
@maximilianoPizarro thx for that contribution!
Would it be possible to contribute that via https://github.com/open-telemetry/opentelemetry-helm-charts?

This file is generated from the `make generate-kubernetes-manifests` command.
And this one on its turn is just generating the manifests from the helm chart.

maximilianoPizarro:
Hi @julianocosta89 , thanks for the feedback! I've added a YAML file for the OpenShift scenario within the Kubernetes folder, **opentelemetry-demo-openshift.yaml**. I rolled back the YAML file generated by the `make generate-kubernetes-manifests` command so they wouldn't overlap for this initial scope. What do you think about saving a version? I'll also be analyzing the integration in the Helm Chart repository. I'd like to add this version as a backup. Have a great week!

julianocosta89:
@maximilianoPizarro moving forward with the Demo we have decided that we will not maintain the k8s-manifests anymore.
We will document the way to generate them using helm.

During the last SIG meeting we agreed that we would create a container containing Helm and if the user wants to deploy the Demo without using Helm, he could run `make generate-kubernetes-manifests` and this command would use the Helm container to generate the k8s manifests.

With that in mind, if we want an OpenShift manifest file, we would need to add it in the Helm chart and then have a param like: `make generate-kubernetes-manifests DISTRO=OpenShift` or something like that.

maximilianoPizarro:
> @maximilianoPizarro moving forward with the Demo we have decided that we will not maintain the k8s-manifests anymore. We will document the way to generate them using helm.
> 
> During the last SIG meeting we agreed that we would create a container containing Helm and if the user wants to deploy the Demo without using Helm, he could run `make generate-kubernetes-manifests` and this command would use the Helm container to generate the k8s manifests.
> 
> With that in mind, if we want an OpenShift manifest file, we would need to add it in the Helm chart and then have a param like: `make generate-kubernetes-manifests DISTRO=OpenShift` or something like that.

The [opentelemetry-demo-openshift.yaml](https://github.com/maximilianoPizarro/opentelemetry-demo/blob/main/kubernetes/opentelemetry-demo-openshift.yaml) file was manually generated and tested on OpenShift 4.19. 

It does not overwrite the Helm process or modify the original file; it's part of an alternative method for starting the demo. 

If it's helpful (for those unfamiliar with Helm), we'll merge it; otherwise, we'll close the pull request. 

Thanks in advance.

puckpuck:
@maximilianoPizarro we want to have better support for OpenShift, but this PR is not the right approach for it. 

As @julianocosta89 mentioned earlier, the intention is to move away from maintaining a Kubernetes manifest in the demo and instead use a `make` command to generate them. We should do that first, then discuss how to apply the necessary changes to the output so that it works in OpenShift.

This also needs to be done in a way that can be reused with each release, and as we make changes to the demo. Maintenance of this will be important.

With all that being said, I will close this PR as it stands. We can discuss using similar approaches for OpenShift after we move to using a Docker container to generate k8s manifests.

PR #2720, opened 2025-11-09
[flagd-ui] add back legacy REST APIs to empower programmatic usage

dottorblaster:
# Changes

Since we switched to Elixir, Phoenix and LiveView for the FlagD UI we lost the ability to programmatically interact with the API of the service to script behaviors. Since we got a couple feedbacks about that, this PR restores the REST APIs that were removed in the rewrite process due to not being necessary anymore thanks to LiveView.

## Merge Requirements

essential items:

### DIFF
--- a/src/cart/src/Dockerfile
+++ b/src/cart/src/Dockerfile
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 # https://mcr.microsoft.com/v2/dotnet/sdk/tags/list
-FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:8.0 AS builder
+FROM --platform=$BUILDPLATFORM mcr.microsoft.com/dotnet/sdk:10.0 AS builder
 ARG TARGETARCH
 
 WORKDIR /usr/src/app/
@@ -30,7 +30,7 @@ RUN dotnet publish ./src/cart.csproj -r linux-musl-$TARGETARCH --no-restore -o /
 # -----------------------------------------------------------------------------
 
 # https://mcr.microsoft.com/v2/dotnet/runtime-deps/tags/list
-FROM mcr.microsoft.com/dotnet/runtime-deps:8.0-alpine3.20
+FROM mcr.microsoft.com/dotnet/runtime-deps:10.0-alpine3.22
 
 WORKDIR /usr/src/app/
 COPY --from=builder /cart/ ./
--- a/src/cart/src/cart.csproj
+++ b/src/cart/src/cart.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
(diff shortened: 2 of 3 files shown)


julianocosta89:
hey @dottorblaster this looks interesting.
Could you add an example to the readme on how to use it?

I think it would also be worth adding something to the OTel.io docs

dottorblaster:
Thanks for the comment @julianocosta89, I added some instructions to the README to explain how to use the API, I don't know where could I put the documentation on the otel.io docs, any pointer? :grimacing:

src/flagd-ui/README.md
```diff
@@ -19,3 +19,173 @@ The application can be run with the rest of the demo using the documented
...
+    },
+    "recommendationCacheFailure": {
+      "defaultVariant": "off",
+      "description": "Fail recommendation service cache",
+      "state": "ENABLED",
+      "variants": {
+        "off": false,
+        "on": true
+      }
+    }
```
julianocosta89:
```suggestion
    ...
```

julianocosta89: Removing the whole file just to avoid the need to update it every time we have a new FF.
I think just a snippet is enough to illustrate it to the user.

src/flagd-ui/README.md
```diff
@@ -19,3 +19,65 @@ The application can be run with the rest of the demo using the documented
...
+
+You can also write a new settings file by sending a new configuration inside
+the `data` field of a POST request body.
+
+Bear in mind that _all_ the data will be rewritten by this write operation.
+
+```sh
+$ curl --header "Content-Type: application/json" \
+  --request POST \
+  --data '{"data": {"productCatalogFailure": {"defaultVariant": "off", "description": "Fail product catalog service on a specific product"}}...' \
```
julianocosta89:
```suggestion
  --data '{"data": {"$schema":"https://flagd.dev/schema/v0/flags.json","flags":{"adFailure":{"defaultVariant":"on","description":"Fail ad service","state":"ENABLED","variants":{"off":false,"on":true}}...' \
```

julianocosta89: I've tried without passing the schema and everything crashed

julianocosta89:
> Thanks for the comment @julianocosta89, I added some instructions to the README to explain how to use the API, I don't know where could I put the documentation on the otel.io docs, any pointer? 😬

I think this page would be the best: https://github.com/open-telemetry/opentelemetry.io/blob/main/content/en/docs/demo/feature-flags/_index.md

julianocosta89:
@dottorblaster I've noticed something weird.

I've updated all FF to `on` via `write` API call.
Tested and everything was working fine (or not working as expected 😅).

With the demo running, I've reverted the git changes on the `demo.flagd.json` file.
The Flagd-ui didn't update the values to `off`, but the Demo started running normally as the FF are all `off` now.

Any ideas?

dottorblaster:
The Flagd-ui keeps an in-memory representation of the state for performance reasons, so if you modify just the file without going through the UI it just won't reflect the changes unless you restart the service :sweat_smile:

PR #2794, opened 2025-12-05
Added Weaver configuration for all telemetry, and a generated mkdocs site

martinjt:
# Changes

This adds a new service called `telemetry-docs` which serves a documentation site for the custom telemetry documented by the demo across all the services.

It also includes a weaver registry for the all those attributes.

This does not include a live-check interface, that will come later, nor does this include codegen for these attributes or metrics.

## Merge Requirements

essential items:


julianocosta89:
@martinjt can we configure live-check to ensure the telemetry schema is always up to date?

I just don't want that to be another burden for maintainers.

Also, the k8s manifests shouldn't be part of this PR. That is generated via helm update whenever we have a new release

martinjt:
We can, i just didnt want to make it part of this PR since its already got a lot of moving parts.

I can remove the k8s bits.

julianocosta89:
@martinjt I had a chat with @puckpuck in the last SIG meeting and we were discussing this PR.
As an OTel user, I couldn't understand why would we run the documentation as part of the Demo.

Shouldn't we keep this PR just with the `/telemetry-schema` and have the telemetry docs as part of the https://opentelemetry.io/docs/demo/ docs?

Something like how the semconv docs is currently done.

martinjt:
My view is that we need to show people how to use Weaver for the things it's built for (Markdown documentation, and telemetry validation). The easiest of those to see is generating markdown. We show them how that's done in this repo, not be having another repo generate them and display them.

If this repo had to have something else host the docs, that wouldn't be something that people could clone, run and iterate on. We need to show people how they can use weaver, give them a place to test it, just like we do with all the languages and their instrumentations.

It's totally valid that the otel docs could ALSO use the schema to generate some more docs that are static and updated on a specific schedule. The repo, the helm chart and the images, should be able to run to allow the user to see all of them together, it shouldn't be that I run the demo, then consult the otel website for the docs (which could be for a different version).

As an application developer, I would likely have a way of viewing the changes of the documentation, I'd maybe have something in my pipeline generate github pages. Either way, it's local to the repo and something that a developer can generate

julianocosta89:
What if we keep the doc service as a different docker compose file? This would allow users to test weaver, but the docs services wouldn't always start with the Demo.
I've added an entry to discuss this at the SIG meeting tomorrow (Wed 14 Jan). (I won't be able to attend it, but happy to accept whatever is agreed by the majority).

julianocosta89 (approved):

julianocosta89:
@martinjt I've touched your PR, hope you don't mind.
Here are the things I've changed:

- Removed the k8s manifests (those are updated via Helm only when we have a new release)
- Reduced image size from 236MB to 102MB, by switching from mkdocs serve with Python runtime to static site build + nginx
  - With that we also gained traces from the telemetry docs service.
- Set static Weaver version on the Dockerfile to avoid breaking changes from `latest`.
- Bumped Python version to 3.14.
- Added the new service to the build images workflow.

This should be good to go, but as I've changed a couple of things, I'll wait for your review.

julianocosta89:
Hey @martinjt just a kind bump 😊

docker-compose.yml
```diff
@@ -377,6 +379,8 @@ services:
         condition: service_started
       flagd-ui:
         condition: service_started
+      telemetry-docs:
+        condition: service_started
```
ps48, line 382: Since the envoy template is shared with minimal docker compose. We should add these in minimal compose or the envoy template will fail to resolve `${TELEMETRY_DOCS_HOST}` and `${TELEMETRY_DOCS_PORT}`.

src/telemetry-docs/templates/markdown/readme.md.j2
```diff
@@ -0,0 +1,47 @@
...
+
+1. **Logical Attribute Groups** - Semantic domain-focused attribute definitions in `common.yaml`
+2. **Service Registries** - Service-specific attribute and metric definitions
+
+Each service references logical attribute groups using the `ref` syntax, creating a separation between logical groupings and service-specific usage patterns.
+
+## Schema Files
+
+- `telemetry-schema/registry_manifest.yaml` - Registry metadata
+- `telemetry-schema/attributes/common.yaml` - Logical attribute group definitions
```
ps48: We don't have `common.yaml`, Should we remove this?

telemetry-schema/attributes/order.yaml
```diff
@@ -0,0 +1,61 @@
...
+        brief: Number of items in cart
+        stability: stable
+        note: The count of items currently in the shopping cart
+        examples: [0, 3, 7]
+      - id: app.payment.amount
+        type: string
+        brief: Payment amount
+        stability: stable
+        note: The monetary amount being charged
+        examples: ["29.99", "99.95", "150.00"]
```
ps48: Nice, this helped me to easily see a difference in instrumentation. `app.payment.amount` in order.yaml is `type: string` but `app.shipping.amount `is type: double` amounts should consistently use the same type. Probably needs fix in a separate PR. I already find this useful 👍🏽

telemetry-schema/attributes/shipping.yaml
```diff
@@ -0,0 +1,45 @@
...
+        brief: Shipping cost amount
+        stability: stable
+        note: The monetary amount charged for shipping
+        examples: [5.99, 15.00]
+      - id: app.shipping.cost.total
+        type: string
+        brief: Total shipping cost
+        stability: stable
+        note: The total cost of shipping for an order (string representation)
+        examples: ["10.50", "25.00"]
```
ps48: Same inconsistency for cost totals. `app.shipping.cost.total` is `type: string` while `app.quote.cost.total` is `type: double`

src/telemetry-docs/mkdocs.yml
```diff
@@ -0,0 +1,56 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+site_name: OpenTelemetry Demo Telemetry Schema
+site_description: Semantic conventions and telemetry schema for the OpenTelemetry Demo application
+site_author: OpenTelemetry Demo Contributors
+site_url: https://opentelemetry-demo.example.com/telemetry/
```
ps48: Should we change this to `http://localhost:8080/telemetry/`

martinjt:
@julianocosta89 @ps48 I've pushed an update to this PR that:

- **Upgrades Weaver from v0.21.2 to v0.22.1** and migrates the schemas to the v2 definition format (`file_format: definition/2`)
- Attribute files now use the flat v2 `attributes:` syntax with `key:` instead of nested `groups:`
- Service files use v2 `attribute_groups:` with `visibility: public`
- Renamed `registry_manifest.yaml` → `manifest.yaml` (old name is deprecated in v0.22)
- Updated the jq filters in `weaver.yaml` to handle the new resolved registry ID format

Also addressed the review comments:
- ✅ Added `TELEMETRY_DOCS_HOST`/`TELEMETRY_DOCS_PORT` to `docker-compose.minimal.yml` so the envoy template resolves correctly
- ✅ Fixed `readme.md.j2` — removed the `common.yaml` reference, updated to reflect actual file structure (`manifest.yaml`, `attributes/*.yaml`, `services/*.yaml`, `metrics/*.yaml`)
- ✅ Updated `mkdocs.yml` site_url to `http://localhost:8080/telemetry/`
- The type inconsistencies (string vs double for amounts/costs) are noted and can be addressed in a follow-up PR

Everything builds and serves correctly with the static nginx setup.

julianocosta89:
@martinjt can you check the easyCLA issue?
It may be related to claude being a co-author of the PR

martinjt:
That should be done now @julianocosta89

julianocosta89:
@martinjt I ran `weaver registry resolve` and got a couple of messages:

```
⚠ Invalid attribute definition detected while resolving '"model/metrics/
  │ product_reviews.yaml"' (group_id='metric.product_reviews.ai_assistant',
  │ attribute_id='product.id'). Missing stability field.
```

I've added stability to the metrics ([97abcc7](https://github.com/open-telemetry/opentelemetry-demo/pull/2794/commits/97abcc7f7ad832613a8b63fd697abaa19ee6c211) and bumped semconv to `1.40`).

Now I'm only getting:

```
⚠ File format `definition/2` is not yet stable: model/services/***
```

But I guess this is expected.

src/telemetry-docs/templates/markdown/attribute_macros.j2
jsuereth, line 1: Something for later - This markdown generation is nice! Maybe it could be contributed to `opentelemetry-weaver-packages` at some point for general re-use?

telemetry-schema/README.md
```diff
@@ -0,0 +1,21 @@
+# OpenTelemetry Demo Telemetry Schema
+
+This directory contains the semantic conventions and telemetry schema for the
+OpenTelemetry Demo application, defining all custom attributes and metrics used
+across the demo services.
+
+## Structure
+
+The schema is organized into three directories:
```
jsuereth, line 9: I guess we don't have span definitions yet, due to limitations in tracking span type?

docker-compose.yml
```diff
@@ -922,6 +925,29 @@ services:
       - "${PROMETHEUS_PORT}:${PROMETHEUS_PORT}"
     logging: *logging
 
+  # Telemetry Documentation
+  telemetry-docs:
```
jsuereth, line 928: Love this idea for the demo!

julianocosta89 (approved):
I'd get this one merged and we can iterate from here

PR #2822, opened 2025-12-17
[chore] update memory limits

puckpuck:
# Changes

updates memory limits to better support the demo running over a long period of time (7+ days). Also updates Prometheus retention to 7d (from 1h), which is fine even with the lower memory limit. This change in retention will align with K8s deployments.


julianocosta89 (approved):
Thx @puckpuck!
I've also bumped the opensearch that was crashing at startup sometimes.

PR #2869, opened 2026-01-08
Use PromQL info function instead of resource attribute promotion

aknuds1:
# Changes

Switch from promoting resource attributes to metric labels to using the experimental PromQL `info()` function, which joins metrics with `target_info` at query time. This reduces metric cardinality in Prometheus by storing resource attributes once in `target_info` rather than on every metric series, aligning with [Prometheus recommendations](https://prometheus.io/blog/2025/12/16/introducing-info-function/).

Bear also in mind that Prometheus might in the future store OTel resource attributes as native metadata — this PR would prepare for that because `info()` should keep working.

The motivation is that I'm a Prometheus OTLP endpoint code owner, and the `info` PromQL function creator, and would like for the demo to reflect Prometheus' recommendation to include OTel resource attributes via `info` instead of through promotion :) Relying too much on promotion has some serious downsides, e.g. high cardinality.

### Prometheus configuration (`src/prometheus/prometheus-config.yaml`)
- Remove most promoted resource attributes, keeping only `kafka.cluster.alias`, `collector.instance.id`, and `host.name` (needed for hostmetrics which lack `service.name`)
- Upgrade Prometheus to v3.11.1 (includes `info()` bug fix [prometheus/prometheus#17817](https://github.com/prometheus/prometheus/pull/17817))
- Enable `--enable-feature=promql-experimental-functions`

### OTel Collector configuration (`src/otel-collector/otelcol-config.yml`)
- Add `resource/postgresql` processor to set `service.name` on PostgreSQL receiver metrics
- Add `transform/postgresql` processor to construct unique `service.instance.id` per PostgreSQL resource scope (database, table, index), preventing duplicate `target_info` entries
- Add `metric_statements` to `transform/sanitize_spans` to set default `service.instance.id` for services that don't provide one
- Add dedicated `metrics/postgresql` pipeline
- Add `transform/sanitize_spans` to the main metrics pipeline

### Docker Compose (`docker-compose.yml`, `.env`)
- Set stable `service.instance.id` for checkout, product-catalog, and flagd (Go/flagd services that generate random IDs from SDK)
- Upgrade Prometheus image to v3.11.1

### Grafana dashboards and alerts
- Migrate all PromQL queries in APM, PostgreSQL, and OpenTelemetry Collector dashboards to use `info()` for resource attribute access
- Migrate cart service alert to use `info()`
- Dashboard variable queries now use `label_values(target_info, ...)`
- Remove broken `pg_stat_bgwriter_*` queries from PostgreSQL dashboard Buffers panel (pre-existing upstream issue, but made verification difficult)

### Kubernetes deployment (`kubernetes/`)
- Add `values.yaml` with Prometheus v3.11.1, `info()` feature flag, minimal resource attribute promotion, and collector transform `metric_statements` for PostgreSQL and duplicate `target_info` prevention
- Add `values-kind.yaml` with Kind-specific overrides
- Add `deploy.sh` for deploying to any k8s cluster (requires `--context`)
- Add `deploy-kind.sh` for local Kind cluster deployment
- Add `kind-config.yaml` with port mapping

## Merge Requirements

essential items:


docker-compose.yml
```diff
@@ -147,7 +147,7 @@ services:
       - GOMEMLIMIT=16MiB
       - OTEL_EXPORTER_OTLP_ENDPOINT
       - OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE
-      - OTEL_RESOURCE_ATTRIBUTES
+      - OTEL_RESOURCE_ATTRIBUTES=${OTEL_RESOURCE_ATTRIBUTES},service.instance.id=checkout
```
cyrille-leclerc: `service.instance.id` is expected to be generated by SDKs or derived from the K8s environment, moreover, it should be a GUID.
Specs: https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/#service-instance-id
K8s naming specs: https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/

src/otel-collector/otelcol-config.yml
```diff
@@ -152,6 +152,14 @@ processors:
...
   resourcedetection:
     detectors: [env, docker, system]
+  resource/postgresql:
+    attributes:
+      - key: service.name
+        value: postgresql
+        action: upsert
+      - key: service.instance.id
+        value: ${env:POSTGRES_HOST}
+        action: upsert
```
cyrille-leclerc: Reading the `service.name` specs [here](https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/#service-name), we could try to broaden the requirements in specs to also define `service.name` in infrastructure monitoring use cases and convince OTel Collector Receiver maintainers to adopt this but today, no infrastructure monitoring receiver produces `service.name` or `service.instance.id`

src/otel-collector/otelcol-config.yml
```diff
@@ -160,6 +168,11 @@ processors:
           # could be removed when https://github.com/vercel/next.js/pull/64852 is fixed upstream
           - replace_pattern(name, "\\?.*", "")
           - replace_match(name, "GET /api/products/*", "GET /api/products/{productId}")
+    metric_statements:
+      - context: resource
+        statements:
+          # Set service.instance.id to service.name if not already set (needed for Prometheus info() joins)
+          - set(attributes["service.instance.id"], attributes["service.name"]) where attributes["service.instance.id"] == nil and attributes["service.name"] != nil
```
cyrille-leclerc, line 245: We would have collision if the same service type (eg a redis) is running multiple times. For infra monitoring metrics, we commonly use attributes like `host.name`... to differentiate the instances.

CLAUDE.md
```diff
@@ -0,0 +1,136 @@
+# CLAUDE.md
```
jmichalek132: not sure you want to commit this file?

CLAUDE.md
```diff
@@ -0,0 +1,136 @@
...
+**Why:**
+- Reduces metric cardinality in Prometheus
+- Resource attributes are stored once in `target_info` rather than on every metric
+- The `info()` function joins metrics with `target_info` at query time
+
+**Note:** Requires Prometheus with `--enable-feature=promql-experimental-functions`.
+
+## Forking Notes
+
+This is a Grafana Labs fork of the upstream OpenTelemetry demo. The main branch is `grafana`.
```
jmichalek132: this comment is wrong

CLAUDE.md
aknuds1: Good point @jmichalek132 - it's useful while working out the PR however - I would remove it before finalizing the PR.

CLAUDE.md
aknuds1: Not sure how Claude ended up thinking that. Thanks, will remove it.

jmichalek132:
Overall looks good outside of what @cyrille-leclerc already pointed out, did you test this locally (given lot of changes to the queries even just re-formatting) do all of the panel still show metrics? Would be potentially nice to show screenshots of it.

aknuds1:
> did you test this locally (given lot of changes to the queries even just re-formatting) do all of the panel still show metrics? 

@jmichalek132 I did some simple testing locally, but I already don't know the demo much, so I'm not a very effective tester :/ Do you know the demo well enough to look for discrepancies?

I did fix the bugs I could find from checking the APM and PostegreSQL dashboards, on Docker Compose.

CLAUDE.md
aknuds1: Done.

aknuds1:
As discussed offline with @cyrille-leclerc, it might be better to implement `instance` label synthesis in Prometheus' OTLP endpoint, based on user configuration, instead of in OTel Collector config (because this would be a hurdle to users).

src/otel-collector/otelcol-config.yml
```diff
@@ -152,6 +152,25 @@ processors:
...
+          # postgresql.index.name as resource attributes, creating multiple target_info
+          # entries with the same identifying labels. By including these in service.instance.id,
+          # each scope gets a unique target_info, allowing info() to work correctly.
+          - set(attributes["service.instance.id"], Concat([attributes["service.name"], "/", attributes["postgresql.database.name"], "/", attributes["postgresql.table.name"], "/", attributes["postgresql.index.name"]], "")) where attributes["postgresql.index.name"] != nil
+          - set(attributes["service.instance.id"], Concat([attributes["service.name"], "/", attributes["postgresql.database.name"], "/", attributes["postgresql.table.name"]], "")) where attributes["postgresql.table.name"] != nil and attributes["postgresql.index.name"] == nil
+          - set(attributes["service.instance.id"], Concat([attributes["service.name"], "/", attributes["postgresql.database.name"]], "")) where attributes["postgresql.database.name"] != nil and attributes["postgresql.table.name"] == nil
+          - set(attributes["service.instance.id"], attributes["service.name"]) where attributes["postgresql.database.name"] == nil
```
ldufr: After talking with @aknuds1, we notice that I had a similar idea, though not using `service.instance.id` (or `instance` from Prometheus side), but a different `resource.uid`.

One advantage of using `service.instance.id` is that it works out of the box for older Prometheus, but it adds additional translation which might create issues. For instance, it does some translation which may be surprising for a user, but it's not clear what will happen if the entity data model is supported. Should the user remove this translation and break the ui?

Using a different attributes has the advantage that while not changing the status-quo, will work better with the introduction of the entity data model and is also a bit more clear. `service.instance.id` is not always enough, sometime it requires `service.name` and `service.namespace` (i.e., `job`).

aknuds1: We will probably drop this client side synthesis though, in favour of something on the Prometheus OTLP side instead.

ldufr: Ok, and presumably, the synthesis wouldn't be changing `instance`, but an other label, correct?

aknuds1: @ldufr The synthesis should generate the `instance` label, why not? If the OTLP endpoint cannot generate `target_info`, because the identifying resource attribute triplet isn't present (`service.namespace`, `service.name`, `service.instance.id`), the idea is to have a potential fallback for determining another identifying resource attribute subset (from which to generate `target_info` and `job` and `instance` labels).

cyrille-leclerc:
Can we verify that Prometheus alerts report enough context, not only `service.instance.id` but also `host.name` , `k8s.cluster.name`, k8s.pod.name`...
We have this context today.

aknuds1:
> Can we verify that Prometheus alerts report enough context

How can we verify this @cyrille-leclerc? I would appreciate any help :D

aknuds1:
> Can we verify that Prometheus alerts report enough context

@cyrille-leclerc Can you check now? Claude helped me implement what I think is a fix for the `CartAddItemHighLatency` alert.

aknuds1:
I've fixed the OTel Collector dashboard too.

puckpuck:
@aknuds1 sorry for the lack of activity on this.

I'd like to get it merged, can you Resolve Conflicts, and I'll run it through a full end-end test to make sure.

aknuds1:
Thanks @puckpuck - I'm fixing it up. I will let you know when it's ready to review.

aknuds1:
@puckpuck I've finalized the PR, and checked the dashboards in Docker Compose and k8s modes. AFAICT they look fine. Please go ahead with your review :)

Do you want me to also make PRs to update [docs](https://opentelemetry.io/docs/demo/) and [Helm charts](https://github.com/open-telemetry/opentelemetry-helm-charts)?

kubernetes/deploy-kind.sh
puckpuck, line 1: All of these files in Kubernetes should be removed. Ultimately we are moving all K8s support to be based on using our Helm chart. The existing manifest file here also needs to be removed since this folder really just causes overall confusion to other contributors.

src/otel-collector/otelcol-config.yml
```diff
@@ -160,6 +160,44 @@ processors:
     spike_limit_percentage: 25
   resourcedetection:
     detectors: [env, docker, system]
+  resource/postgresql:
```
puckpuck: I understand why we need this, but I feel like this is alot of transform required to get common infra metrics working with the info function in PromQL. 

What about Kafka, or Redis metrics. Or metrics from other infra, so we need to ensure they all have unique service.instance.id attributes as well?

src/otel-collector/otelcol-config.yml
puckpuck: I agree, we don't want to set the id to a name that may have a collision. This needs to be set to the pod/container/host name which should be unique.

src/otel-collector/otelcol-config.yml
```diff
@@ -216,8 +259,12 @@ service:
       processors: [resourcedetection, memory_limiter, transform/sanitize_spans]
       exporters: [otlp_grpc/jaeger, debug, spanmetrics]
     metrics:
-      receivers: [docker_stats, httpcheck/frontend-proxy, hostmetrics, nginx, otlp, postgresql, redis, spanmetrics, kafkametrics]
-      processors: [resourcedetection, memory_limiter]
+      receivers: [docker_stats, httpcheck/frontend-proxy, hostmetrics, nginx, otlp, redis, spanmetrics, kafkametrics]
+      processors: [resourcedetection, transform/sanitize_spans, memory_limiter]
+      exporters: [otlp_http/prometheus, debug]
+    metrics/postgresql:
```
puckpuck: We should not expose multiple metrics pipelines, this will cause issues with people that fork or extend the demo. Let's get everything into the standard pipelines (traces, metrics, logs), and use filters in the the processors themselves.

Also fwiw, `memory_limiter` should always be the first processor in the list. I noticed it is not here, but it should be.

puckpuck (changes requested):

cyrille-leclerc:
I’m aligned with @puckpuck. This PR does a great job highlighting the challenges around OTel infrastructure metrics resource attributes with Prometheus, and it reinforces the value of the OTel Demo as a realistic environment to understand how these pieces fit together.

That said, it feels a bit early to replace Prometheus resource attribute promotion in the OTel Demo with the `info` function, especially since attributes like `service.name` and `service.instance.id` are not consistently present today.

It might be helpful to align as a group on a few milestones that could guide a confident transition:

- Evolve the OTel semantic conventions to make `service.name` and `service.instance.id` required, including for infrastructure components like databases and for the operating system. Otherwise, metrics lacking these attributes would lose all context, including `host.name` or `k8s.*` attributes that are critical and commonly present on Prometheus-style metrics ([resources](https://opentelemetry.io/docs/specs/semconv/resource/service/))
- Update Prometheus documentation to gradually discourage resource attribute promotion in favor of the `info` function ([docs](https://prometheus.io/docs/guides/opentelemetry/#promoting-resource-attributes))
- Ensure OTel Collector Contrib receivers reliably emit `service.name` and `service.instance.id` in line with semconv, including support for environment-based configuration such as Kubernetes annotations ([spec](https://opentelemetry.io/docs/specs/semconv/non-normative/k8s-attributes/))
- Validate that this model resonates with practitioners. From what I remember at the last OTel Unconference, the community tended to prefer resource attribute promotion over joins with `target_info`

Curious to hear what others think about this direction.

puckpuck:
Yes, we should use this as an initiative to push the underlying ecosystem (ie: infra metrics receivers) to support the upcoming standard. Perhaps something we can discuss at the next SIG meeting to understand who else needs to be notified and coordinated with.

kubernetes/deploy-kind.sh
aknuds1: I see, thanks for making me aware.

aknuds1: I thought about it. The kubernetes/deploy-kind.sh script is useful for me to install into a local k8s cluster and test my changes. It's already based around Helm, what's the argument against it? One would think it's generally useful for testing OTel demo in k8s mode?

src/otel-collector/otelcol-config.yml
aknuds1: Please see my [Postgres receiver issue](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45347). As @lmolkova [points out](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45347#issuecomment-3863117562), I think its current modeling of resources and metrics is broken. Not just due to emitting a non-uniquely identifying `service.instance.id`, but more generally due to putting attributes on the resource level which should be on the metrics level (causing there to be multiple distinct resources per `service.instance.id`, and unnecessary need to promote resource attributes to metric labels).

At least one of the maintainers (@antonblock) [agrees](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45347#issuecomment-3915658217) that the issue's proposed solution is good.

Do you think we should hold off on this PR until the Postgres issue is fixed?

aknuds1:
> That said, it feels a bit early to replace Prometheus resource attribute promotion in the OTel Demo with the info function, especially since attributes like service.name and service.instance.id are not consistently present today.

@cyrille-leclerc If `service.instance.id` is missing, which Prometheus depends on, it can be solved through a processor though. Can you please provide your opinion @lmolkova, on the right solution? Should the OTel Demo use a processor to insert `service.instance.id` when corresponding OTel receivers don't already do it themselves?

@puckpuck Please also note that when the OTel Entity Data Model becomes standard, Prometheus will no longer need `service.instance.id` to identify resources, as entities will flag which resource attributes are identifying. I'm not sure when the Entity Data Model will go into production though, which makes the whole thing uncertain.

aknuds1:
> Perhaps something we can discuss at the next SIG meeting to understand who else needs to be notified and coordinated with.

@puckpuck In respect to this, please see my [other comment](https://github.com/open-telemetry/opentelemetry-demo/pull/2869#discussion_r3064844210).

src/otel-collector/otelcol-config.yml
aknuds1: I've tried to fix this in the latest commit.

src/otel-collector/otelcol-config.yml
aknuds1: > This needs to be set to the pod/container/host name which should be unique.

How is this implementable under Docker Compose @puckpuck?

puckpuck:
I added this PR to the agenda as something to discuss in the next Demo SIG meeting on 4/15 @ 11am New York time.

In the spirit of specifications and standards moving forward we want to support this, but we need to make sure it is ready enough today for us to support it.

aknuds1:
Appreciate it @puckpuck.

puckpuck:
We have decided to not move forward with using this yet. Ideally we can do this without the need for custom transforms for this to work. We would also like to see the Prometheus community come forward with advocate for this being the right configuration forward.

I created issue #3264 to track adding this capability in the future.

aknuds1:
> We would also like to see the Prometheus community come forward with advocate for this being the right configuration forward.

@puckpuck Could you elaborate on what this means in practice? Not sure what the expectation is.

PR #2885, opened 2026-01-12
Fix metrics temporality env var interpolation in docker-compose

### LINKED ISSUE #2884: OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE from .env not applied to services
# Bug Report

## Symptom

Prometheus logs show "invalid temporality and type combination" errors for many metrics, causing them to be rejected:

```
level=ERROR msg="Error appending remote write" err="invalid temporality and type combination for metric \"http.server.request.duration\"; invalid temporality and type combination for metric \"rpc.server.duration\"; ..."
```

This affects most services' RED metrics (Rate, Errors, Duration), making the APM dashboard unusable for services like cart, checkout, ad, etc.

## Root Cause

The `docker-compose.yml` uses the syntax:
```yaml
environment:
```

This syntax passes the variable **from the shell environment**, not from the `.env` file. When the shell doesn't have this variable set, Docker Compose doesn't apply the `.env` value - instead, the value baked into the container images (`delta`) takes precedence.

The `.env` file correctly sets:
```
OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE=cumulative
```

But this value is never applied to the containers.

**Verification:**
```bash
# Shows delta (wrong - from image)
docker exec cart env | grep OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE

# .env file has cumulative (correct)
grep TEMPORALITY .env
```

## Fix

Change all occurrences in `docker-compose.yml` from:
```yaml
```

To explicit interpolation:
```yaml
- OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE=${OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE}
```

This ensures Docker Compose uses the value from `.env` file.

## Impact

Without this fix:
- **21 services** have metrics rejected by Prometheus
- HTTP server metrics (`http.server.request.duration`) from .NET services (cart, accounting) are lost
- RPC server metrics (`rpc.server.duration`) from Go/Java services are lost
- All JVM, .NET, Go, Node.js, and Python runtime metrics are lost
- Kafka metrics are lost
- APM dashboard only works for flagd

## Reproduce

1. `make start`
2. Check container env: `docker exec cart env | grep TEMPORALITY` → shows `delta`
3. Check Prometheus logs: `docker logs prometheus 2>&1 | grep "invalid temporality"`
4. Observe many metrics being rejected

## Environment

- Docker Compose v2.x
- macOS / Linux
- Demo version: main branch (commit ebb1818)

aknuds1:
## Summary

Fixes #2884

The `docker-compose.yml` uses the syntax `- OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` which passes the variable from the shell environment, not from the `.env` file. When the shell doesn't have this variable set, Docker Compose doesn't apply the `.env` value - instead, the value baked into the container images (`delta`) takes precedence. See [reference](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#ways-to-set-variables-with-interpolation) on how environment variable interpolation works with .env files.

This causes Prometheus to reject metrics with "invalid temporality and type combination" errors, affecting HTTP/RPC server metrics from most services.

## Changes

Changed all occurrences from:
```yaml
```

To explicit interpolation:
```yaml
- OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE=${OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE}
```

## Test plan

1. `make start`
2. Verify `docker exec cart env | grep TEMPORALITY` shows `cumulative`
3. Verify `docker logs prometheus 2>&1 | grep "invalid temporality"` shows no errors
4. Check APM dashboard shows metrics for cart, checkout, ad, etc.


julianocosta89:
This change shouldn't be necessary.
All the env vars come from the .env file, and when the env var has the same name as whatever is set in the .env file, the value from the .env file is used.

@aknuds1 Could you elaborate further on the issue?
Most OTLP endpoints and all resource attributes are being set without ${value} being explicitly set.
Everything comes from .env file.

aknuds1:
I will try again tomorrow.

aknuds1:
@julianocosta89 I tried to reproduce yesterday's problem (delta temporality being configured in the demo stack) this morning, but couldn't. I've inferred that the probable cause was `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE=delta` being exported in my shell for some reason unknown to me (I didn't set it knowingly). I got confused by the Docker Compose documentation making it look like there's a behavioral difference between the syntaxes. Sorry for the noise!

Issue #2905, opened 2026-01-16
checkout service uses non-standard span event name "error" instead of "exception"


neptunian:
# Bug Report

**Version:** commit `main`

## Symptom

The Go `checkout` service uses `span.AddEvent("error", ...)` instead of the semantic convention compliant event name `"exception"`. 

https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/checkout/main.go#L305

This causes problems for downstream consumers.  In our case, the Elastic APM processor does not recognize these as exception events and fails to add enrichment fields. This breaks error grouping and visibility in downstream APM tooling used in Kibana.  

**What is the expected behavior?**

Per the [OTel exception semantic conventions](https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-spans/), span exception events should use `"exception"` as the event name, or use the SDK helper `span.RecordError()` which sets the correct name automatically. https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/span.go#L544-L567

**What is the actual behavior?**

The checkout service uses:
`span.AddEvent("error", trace.WithAttributes(semconv.ExceptionMessageKey.String(err.Error())))`

## Reproduce

1. Run the demo with payment service failure enabled
2. Query logs index for checkout exception events
3. Observe `event_name: "error"` instead of `event_name: "exception"`

ahfuckit:
Wonder why that might be... maybe the guy who wrote it did that on purpose so he could identify people using his code without proper attribution 🤔

PR #2906, opened 2026-01-16
[checkout]: use span.RecordError instead of AddEvent(error)

neptunian:
fix: https://github.com/open-telemetry/opentelemetry-demo/issues/2905

span.RecordError uses the semconv-compliant exception event name and automatically adds exception.type.

# Changes

The checkout service was using span.AddEvent("error", ...) which does not follow OpenTelemetry semantic conventions. This causes downstream processors that expect the standard "exception" event name to miss these errors.

Changed to span.RecordError(err) which:
- Uses the correct "exception" event name per semconv
- Automatically includes exception.type attribute
- Is the idiomatic Go SDK approach for recording errors

### DIFF
--- a/src/accounting/Accounting.csproj
+++ b/src/accounting/Accounting.csproj
@@ -9,17 +9,17 @@
 	</PropertyGroup>
 
 	<ItemGroup>
-		<PackageReference Include="Confluent.Kafka" Version="2.12.0" />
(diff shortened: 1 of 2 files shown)


julianocosta89:
@neptunian thanks for that, would you be able to sign the easyCLA?
Without it we can't accept your contribution.

neptunian:
@julianocosta89 Looking into it, thanks!

PR #3014, opened 2026-02-19
feat: Replace mock LLM with local SmolLM2-135M inference

ps48:
# Changes

 ## Summary
  - Replace the mock LLM service with SmolLM2-135M running locally via `llama-cpp-python`, removing the need for an external API while still exposing an OpenAI-compatible `/v1/chat/completions` endpoint
  - Simplify `product_reviews_server.py` by dropping the custom tool-calling flow in favor of a direct chat completion with review context in the prompt
  - Add manual OpenTelemetry GenAI semantic-convention spans and metrics to the LLM service (token usage histogram, operation duration)
  - Pin OTel library versions in `src/llm/requirements.txt` to 1.39.1 / 0.60b1
  - Update comments in `.env`, `.env.override`, and `src/product-reviews/README.md` to reflect the new default model

## Merge Requirements

essential items:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ the release.
 
 ## Unreleased
 
+* [llm] Replace mock LLM with local SmolLM2-135M inference via llama-cpp-python
+  ([#3014](https://github.com/open-telemetry/opentelemetry-demo/pull/3014))
 * [accounting] fix memory leak with dbcontext
   ([#2876](https://github.com/open-telemetry/opentelemetry-demo/pull/2876))
 * [chore] Upgrade OTel Collector to v0.145.0 with :warning: breaking change:
--- a/src/llm/Dockerfile
+++ b/src/llm/Dockerfile
@@ -5,7 +5,7 @@
 FROM docker.io/library/python:3.12-slim-bookworm AS build
 
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends build-essential g++ curl && \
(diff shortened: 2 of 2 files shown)


kylehounslow:
Real inference, nice!  One thought on the instrumentation side:

The `product_reviews_server.py` already uses the `openai` Python client to talk to this service. Instead of hand-rolling the GenAI spans and metrics in the LLM server, we could instrument the client side with [`opentelemetry-instrumentation-openai-v2`](https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation-genai/opentelemetry-instrumentation-openai-v2). That would:

1. Dogfood OTel's own instrumentation library in OTel's own demo, which feels like the right place to showcase it
2. Stay in sync with semconv changes as the library evolves, vs maintaining attribute names by hand
3. Drop ~60 lines of manual OTel setup from `src/llm/app.py` (TracerProvider, MeterProvider, manual span creation, histogram recording, the resource attribute parsing, etc.)

The LLM service can stay simple: Flask + llama-cpp-python + Flask auto-instrumentation. The GenAI telemetry makes more sense on the client side where the `openai` SDK lives.

Setup would just be:
```python
from opentelemetry.instrumentation.openai_v2 import OpenAIInstrumentor
OpenAIInstrumentor().instrument()
```

What do you think?

ps48:
Will close this PR based on the discussion in the last SIG demo meeting, will create a new PR for updating gen_ai attributes.

PR #3136, opened 2026-03-20
Add GHCR cleanup workflow to delete old container images

AnjaliMishra1st:
This PR introduces a GitHub Actions workflow to clean up old container images from GHCR.

- Runs daily using a scheduled workflow
- Deletes older container image versions
- Keeps the latest 5 versions

This helps reduce storage clutter and maintain a clean registry.


osullivandonal:
@AnjaliMishra1st nice work on adding this job, just some things to consider here.

This cleanup job is quite aggressive, the otel demo release cadence is usually every few months when a number of changes have accumulated. This job would run every day which give that release cadence is very high. However, we do have a nightly build of images. Also another thing to consider is users of older images in the github container registry, the demo is used a lot by different users to showcase how to instrument a cloud service, we may break a lot of demo's by removing older versions of images. 

For example using `min-versions-to-keep: 5` would only keep less than a week of images in the nightly builds.

It would be great to hear from the maintainers here @open-telemetry/demo-maintainers .

AnjaliMishra1st:
Thanks for the helpful feedback!
I’ve updated the workflow to run weekly instead of daily and increased the number of versions to keep for safer retention.
Please let me know if this looks better or if further adjustments are needed.

AnjaliMishra1st:
Hi maintainers 👋
Just a gentle follow-up on this PR. Based on earlier feedback, I’ve updated the workflow to run weekly and increased the number of retained versions to make it safer for users.
Could you please review when you have time? I’d really appreciate your feedback.

@open-telemetry/demo-maintainers
@osullivandonal
Thanks again! 🙂

julianocosta89 (commented):
@AnjaliMishra1st sorry for the delay, last week was KubeCon and I'm still playing catch!
Regarding this PR I think we need to adjust a couple of things.

The Demo currently releases nightly builds, but we also have releases.
If I understood this gh workflow correctly, that will clean up everything that is older than 20 versions, but it doesn't differentiate releases from nightly builds.
Which means that if we run this today, we would remove versions 2.2.0, 2.1.3, and so on.

Ideally the clean up would run just for the nightly builds.

AnjaliMishra1st:
Hi @julianocosta89 👋🏽
Thanks for the detailed feedback — that makes sense!
I understand that the workflow should not delete release versions and should only target nightly builds.
I’ll update the workflow to restrict cleanup to nightly images only and keep all release versions safe.

I’ll push the changes soon. Thanks again! 🙂

puckpuck:
Is there a way we can run this in some kind of dry-run mode first to see what images it would delete?

AnjaliMishra1st:
Hi @julianocosta89 @puckpuck
Thank you for the helpful feedback! I’ve updated the workflow to include a dry-run mode and to restrict cleanup to nightly builds only, ensuring release versions remain safe.
Whenever you have time, I’d really appreciate your review and any further suggestions. Thank you! 🙂

puckpuck (approved):

puckpuck:
This looks fine. As it's configured right now nothing destructive will happen, and at 3am it will execute in dry_run mode producing a list of images that it would delete.

If that looks good, we can do another PR to update the dry_run setting to be false.

puckpuck:
@AnjaliMishra1st it's failing the yamllint and checklicense checks. We need to add a license copyright header to the file (see other GH action workflows for an example), and yamllint is likely extra white space or empty lines. The failed check output will have better details on the offending entries.

.github/workflows/cleanup-ghcr.yml
```diff
@@ -0,0 +1,64 @@
...
+        required: false
+        default: "true"
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
```
Kielek: All other GH actions pins versions to the exact commit + tag comment. 
It should be applied also here.

.github/workflows/cleanup-ghcr.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: Install GitHub CLI
+        run: sudo apt-get update && sudo apt-get install gh -y
```
Kielek: are you sure that it is needed? I think that GitHub CLI should be available by default on all-defualt runmers.

.github/workflows/cleanup-ghcr.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: Authenticate GitHub CLI
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | gh auth login --with-token
```
Kielek: I think that this one is also not-necessary. In OTel .NET repository there is usage of gh-cli, but there is no need to login.

.github/workflows/cleanup-ghcr.yml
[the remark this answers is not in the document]
Kielek: After hash add some comment related to version, see other files

AnjaliMishra1st: Good catch, I’ve updated this to use a pinned commit SHA and added a version comment for consistency with other workflows.

.github/workflows/cleanup-ghcr.yml
AnjaliMishra1st: You're right, I’ve removed the installation step since GitHub CLI is already available on default runners.

.github/workflows/cleanup-ghcr.yml
AnjaliMishra1st: Thanks for pointing this out, removed the unnecessary authentication step.

AnjaliMishra1st:
Hi @Kielek @julianocosta89 @puckpuck
Thanks for the helpful feedback!
I’ve made the suggested updates:
- Pinned checkout action with SHA
- Removed unnecessary GitHub CLI setup
- Added required permissions
- Fixed dry-run logic and pagination
- Ensured safe deletion with break and tag handling
I’ve kept the package name as-is for now—happy to adjust if needed.
Please let me know if anything else should be improved. Thanks!

AnjaliMishra1st:
Hi maintainers
Just a gentle follow-up on this PR.
Thank you again for all the helpful feedback so far — I’ve addressed the suggested changes, including pagination, permissions, pinned action SHA, dry-run logic, and safer nightly-only cleanup handling.
I noticed the PR is currently approved but still blocked due to an unresolved conversation and a workflow awaiting maintainer approval.
Whenever you have time, could someone please help review the remaining conversation thread and approve the pending workflow so we can move this forward?
I’d really appreciate your guidance. Thank you 🙂

rogercoll (commented):
Apologize for my late review, but what do you think if instead of having a workflow that cleans up old container images we just override the tag of the released nightly images? The idea is to use a tag that references the day of the week, giving us a rolling set of the most recent seven images.
(we would need to manually remove all the current images)

AnjaliMishra1st:
Thank you for the great suggestion — I really like this approach.
Using rolling day-of-week tags for nightly images seems safer and simpler than periodic cleanup.
Would you prefer that I update this PR for the new tagging strategy, or open a follow-up PR focused on the nightly build workflow?

Thank you again for the guidance!

rogercoll:
> Thank you for the great suggestion — I really like this approach. Using rolling day-of-week tags for nightly images seems safer and simpler than periodic cleanup. Would you prefer that I update this PR for the new tagging strategy, or open a follow-up PR focused on the nightly build workflow?
> 
> Thank you again for the guidance!

I have no preference, but maybe a new one will have a cleaner history, and we can reference this one as per the decision.

 Thank you

AnjaliMishra1st:
Thank you everyone for the valuable feedback and guidance throughout this PR.
Based on the latest discussion, I agree that using rolling day-of-week tags for nightly images is a safer and cleaner approach than maintaining a separate cleanup workflow.
To keep the history clean and make the implementation easier to review, I’ll open a new PR focused on updating the nightly build workflow with rolling tags, and reference this PR there as part of the design decision.

Thank you again for all the support and suggestions.

julianocosta89:
@rogercoll what do you think?

AnjaliMishra1st:
Hi maintainers 
Thanks for the feedback and discussion. Based on the decision to move to rolling weekday tags, I’ll close this PR and continue with the new approach in a separate PR.
Appreciate all the guidance!

Issue #3140, opened 2026-03-23
Add astronomy shop agent which can navigate users in astronomy shop applications.


fali007:
# Feature Request

Along with the existing micro service based astronomy shop application, I would like to add:

1. Agentic version of astronomy shop where the agent accepts Natural Language queries instead of HTTP load from load generator.
2. Astronomy shop MCP server (The agent will use appropriate tool(s) to satisfy the user request.)
3. A chatbot for humans to send requests to agent.
4. A NL load generator analogous to existing load generator which can send NL queries to agent.

**Is your feature request related to a problem?**
In the age of LLM automations, it would be good to have a test ecosystem where people can test the capabilities of the agents and discover the new challenges arising from this shift.
The advantage of converting an existing micro service system is that we will get the ground truth traces/logs for a similar agentic execution.

**Describe the solution you'd like:**
Implementation of the agent and tools for astronomy shop.

Agent
a) Users can provide a yaml configuration for a simple Agent (system prompt + tools) which can be extended to a MultiAgent configuration. The advantage is that users can test out and verify the different architectural advantages. (Eg: Splitting tools across agents might help in reducing the token usage)
b) A static single agent application

Tools
a) Astronomy Shop MCP server - Advantage is easy to use server which is ready made to use. Possible disadvantage is lack of flexibility to modify tool descriptions.
b) User can provide the tools as a config file and we can build the tools and tool serving(MCP or native langGraph tools) dynamically. Advantage is greater degree of flexibility.

LoadGenerator
a) Manually crafted queries analogous to existing load generator HTTP queries. Toggle feature for agent and existing load.

LLM
a) Configure the agent to use any LLM provider.
b) Configure a VCR caching of LLM responses to reuse previously seen responses. This feature will help new users without LLM's with an easy onboarding and reduce the token cost.

**Describe alternatives you've considered.**

Mentioned above

## Additional Context

<img width="4703" height="6360" alt="Image" src="https://github.com/user-attachments/assets/0e1666a2-eb18-4267-b899-f06117141116" />

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

PR #3148, opened 2026-03-24
Draft :  Add agent astronomy shop and astronomy shop MCP server

fali007:
# Changes
This is a Draft PR to demo changes to address #3140 

## Merge Requirements

essential items:


src/mcp/src/mcp_server/tools.py
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/python
```
ps48: Looks like this is a dupe for `src/agents/tools.py` Can we just keep one?

src/agent/src/agents/tools.py
```diff
@@ -0,0 +1,140 @@
...
+        res = requests.get(url)
+        res.raise_for_status()
+        return res.json()
+    except Exception as e:
+        return f"Error while fetching currency list: {e}"
+
+
+def convert_currency(from_currency: str, to_currency: str, amount: float):
+    """Convert between currencies."""
+    url = f"http://{BASE_URL}/api/currency/convert?from={from_currency}&to={to_currency}&amount={amount}"
```
ps48: Where is this route hosted? I see that the `src/frontend/pages/api/currency.ts` is the only handler to GET returning currencyCodes[]`.  Currency conversion is done internally via gRPC (CurrencyGateway.convert) and is never exposed over HTTP.

src/agent/src/agents/tools.py
```diff
@@ -0,0 +1,140 @@
...
+        res = requests.get(url, params=params)
+        res.raise_for_status()
+        return res.json()
+    except Exception as e:
+        return f"Error fetching recommendations: {e}"
+
+
+def get_shipping_quote(address_id: str):
+    """Get estimated shipping cost for a given address."""
+    url = f"http://{BASE_URL}/api/shipping/quote/{address_id}"
```
ps48: `/api/shipping/quote/{address_id}` does not exist. The actual endpoint is GET /api/shipping with query parameters itemList (JSON-stringified CartItem[]), currencyCode, and address (JSON-stringified Address). Should be something like:

```
url = f"http://{BASE_URL}/api/shipping"
params = {"itemList": json.dumps(items), "currencyCode": currency, "address": json.dumps(address)}
```

src/mcp/src/mcp_server/tools.py
```diff
@@ -0,0 +1,140 @@
...
+        res = requests.post(url, json=data)
+        res.raise_for_status()
+        return res.json()
+    except Exception as e:
+        return f"Error while adding product to cart: {e}"
+
+
+def get_cart():
+    """Retrieve the current contents of a user's cart."""
+    url = f"http://{BASE_URL}/api/cart"
```
ps48: Do we need session id here https://github.com/open-telemetry/opentelemetry-demo/blob/5b7074107ca79537552c967698b789358e994d13/src/frontend/pages/api/cart.ts#L17 
```
  def get_cart(sessionId: str):
      url = f"http://{BASE_URL}/api/cart"
      res = requests.get(url, params={"sessionId": sessionId})
```

src/agent/src/agents/llm.py
```diff
@@ -0,0 +1,39 @@
...
+
+class ChatLLM(ChatOpenAI):
+    def __init__(self, **kwargs):
+        model_name = os.getenv("LLM_MODEL", "default")
+        use_vcr = os.getenv("USE_VCR", False) == "True"
+        cassette_name = ""
+        if use_vcr:
+            cassette_name = f"{model_name.replace('/', '_')}_casette.yaml"
+        if "http_async_client" not in kwargs:
+            kwargs["http_async_client"] = httpx.AsyncClient(verify=False)
```
ps48: Should we gate this behind an env var (e.g., LLM_TLS_VERIFY). ?  LLM provider is external is usually external so should we default to `verify=True`

src/agent/requirements.txt
```diff
@@ -0,0 +1,19 @@
...
+mcp==1.22.0
+fastmcp>=2.13.1
+langchain_mcp_adapters==0.1.14
+langchain_community==0.4.1
+fastapi==0.120.4
+opentelemetry-api==1.38.0
+opentelemetry-sdk==1.38.0
+opentelemetry-semantic-conventions==0.59b0
+opentelemetry-instrumentation-langchain<0.53.0
+fastmcp==2.13.1
```
ps48: looks like a dupe of line 9 where we have `fastmcp>=2.13.1`

src/mcp/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+ENV PYTHONDONTWRITEBYTECODE=1 \
+    PYTHONUNBUFFERED=1
+WORKDIR /app
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    build-essential \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY src/agent/requirements.txt .
```
ps48: This should be `src/mcp/requirements.txt`

src/chatbot/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+ENV PYTHONDONTWRITEBYTECODE=1 \
+    PYTHONUNBUFFERED=1
+WORKDIR /app
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    build-essential \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY src/agent/requirements.txt .
```
ps48: This should be `src/chat_interface/requirements.txt`

src/agent/src/agents/agents.py
```diff
@@ -0,0 +1,109 @@
...
+        if self.mcp_server:
+            await self.mcp_server.cleanup()
+
+    async def handle_prompt(self, request: ChatRequest):
+        return await self.run_agent(request.message)
+
+    async def get_tool_list(self):
+        mcp_enabled = os.getenv("MCP_ENABLED", "False") == "True"
+        if mcp_enabled and self.mcp_server is not None:
+            await self.mcp_server.connect_to_mcp_server(self.mcp_server_url)
```
ps48: `get_tool_list()` calls `connect_to_mcp_server()` on every `/prompt` request, creating a new MCP session each time. The connection should be established once during the lifespan startup and reused.

src/mcp/src/mcp_server/tools.py
```diff
@@ -0,0 +1,140 @@
...
+    url = f"http://{BASE_URL}/api/products/{product_id}"
+    try:
+        res = requests.get(url)
+        res.raise_for_status()
+        return res.json()
+    except Exception as e:
+        return f"Error while fetching product {product_id}: {e}"
+
+
+def checkout(checkout_person):
```
ps48: Let's add a Pydantic model or at minimum dict type. LLM tool-calling frameworks need type info to generate the tool schema. Without it, the agent won't know what structure to pass

docker-compose.yml
```diff
@@ -778,6 +773,105 @@ services:
         condition: service_started
     logging: *logging
 
+  # Agent
+  agent:
```
ps48: The agent, mcp, and chatbot services are missing OTEL_SERVICE_NAME and OTEL_RESOURCE_ATTRIBUTES environment variables that all other services define. For an observability demo, telemetry data from these services won't be properly labeled.

docker-compose.yml
```diff
@@ -778,6 +773,105 @@ services:
...
+      - AGENT_PORT
+      - AGENT_ENDPOINT
+      - CHATBOT_PORT
+    depends_on:
+      jaeger:
+        condition: service_started
+      otel-collector:
+        condition: service_started
+      product-catalog:
+        condition: service_started
```
ps48: The chatbot only talks to the agent service, but its depends_on lists jaeger, otel-collector, and product-catalog. It should depend on agent instead.

.env
```diff
@@ -192,3 +192,21 @@ JAEGER_GRPC_PORT=4317
...
+# Agent
+GRAPH_RECURSION_LIMIT=25
+MCP_ENDPOINT=mcp
+MCP_PORT=8011
+MCP_ENABLED=False
+AGENT_ENDPOINT=agent
+AGENT_PORT=8010
+CHATBOT_PORT=7860
+LLM_BASE_URL=
+LLM_MODEL=Azure/gpt-4o
```
ps48: `LLM_MODEL=Azure/gpt-4o` the OpenTelemetry demo shouldn't default to a vendor-specific model. Please consider leaving this empty or using a placeholder with a comment explaining how to configure it.

src/agent/src/agents/tools.py
```diff
@@ -0,0 +1,140 @@
...
+    params = {"contextKeys": category}
+    try:
+        res = requests.get(url, params=params)
+        res.raise_for_status()
+        return res.json()
+    except Exception as e:
+        return f"Error fetching ads: {e}"
+
+
+def add_to_cart(userId: str, productId: str, quantity: int = 1):
```
ps48: The parameter names mix camelCase and snake_case: userId, productId (camelCase) vs product_id, from_currency, to_currency (snake_case). Since these become tool parameter names visible to the LLM, inconsistency can cause the agent to use wrong names. Please pick one convention (snake_case is Pythonic).

src/chatbot/run.py
```diff
@@ -0,0 +1,33 @@
...
+from dotenv import load_dotenv
+from src.chat_interface.chat_interface import ChatAgentUI, get_chat_ui_config
+
+logging.basicConfig(level=logging.INFO)
+
+load_dotenv()
+
+
+async def start_servers():
+    """Run both the LangGraph Agent and the MCP server concurrently."""
```
ps48: This looks like a stale docstring. We can say this runs the chat UI server?

src/mcp/run.py
```diff
@@ -0,0 +1,40 @@
...
+load_dotenv()
+
+Traceloop.init(
+    app_name="AstronomyShopAgentMCP",
+    api_endpoint=os.getenv("OTEL_EXPORTER_OTLP_ENDPOINT", "localhost:4317"),
+)
+
+
+async def start_servers():
+    """Run both the LangGraph Agent and the MCP server concurrently."""
```
ps48: Same here, This looks like a stale docstring. We can say this runs the MCP server?

ps48 (commented):
@fali007 Added some initial set of comments. Also can you please rebase you PR on top of main to remove the conflicts!

docker-compose.yml
fali007: Added for 3 new services

docker-compose.yml
fali007: Corrected the dependency for MCP and chatbot

.env
fali007: Removed

src/mcp/run.py
fali007: Made the change

src/chatbot/run.py
fali007: Made the change

src/agent/src/agents/tools.py
fali007: Changed

src/mcp/src/mcp_server/tools.py
fali007: Added details of request model to tool description.

src/agent/src/agents/llm.py
fali007: Added behind env variable.

src/mcp/src/mcp_server/tools.py
fali007: Added session_id as request param

src/mcp/src/mcp_server/tools.py
fali007: Sure. Made the change

src/agent/src/agents/tools.py
fali007: Made the change

src/agent/src/agents/tools.py
fali007: Removed the tool

src/agent/src/agents/agents.py
fali007: Changed this to initialize the session in lifespan and reuse it for each request

src/agent/src/agents/tools.py
```diff
@@ -0,0 +1,140 @@
...
+def empty_cart(user_id: str):
+    """Empty the shopping cart for a user."""
+    url = f"http://{BASE_URL}/api/cart/empty/{user_id}"
+    try:
+        res = requests.get(url)
+        res.raise_for_status()
+        return res.json()
+    except Exception as e:
+        return f"Error while emptying cart: {e}"
+
```
ps48: Looks like a different URL & HTTP method. The frontend API has no `GET /api/cart/empty/{user_id}` route. It should be:

```
def empty_cart(user_id: str):
    """Empty the shopping cart for a user."""
    url = f"http://{BASE_URL}/api/cart"
    data = {"userId": user_id}
    try:
        res = requests.delete(url, json=data)
        res.raise_for_status()
        return res.json()
    except Exception as e:
        return f"Error while emptying cart: {e}"

File: src/agent/src/agents/tools.py [hunk 18]
```diff
@@ -0,0 +1,140 @@
...
+    params = {"productIds": product_id}
+    try:
+        res = requests.get(url, params=params)
+        res.raise_for_status()
+        return res.json()
+    except Exception as e:
+        return f"Error fetching recommendations: {e}"
+
+
+def get_shipping_quote(items, currency_code, address):
```
ps48: `requests.get(url, params)` passes the dict as the positional params argument, which technically works, but the real issue is that the shipping API is a POST endpoint expecting a JSON body, not a GET with query-string-encoded JSON. 

```
def get_shipping_quote(items, currency_code, address):
    """Get estimated shipping cost for a given address."""
    url = f"http://{BASE_URL}/api/shipping"
    data = {"items": items, "currencyCode": currency_code, "address": address}
    try:
        res = requests.post(url, json=data)
        res.raise_for_status()
        return res.json()
    except Exception as e:
        return f"Error fetching shipping quote: {e}"
```
Also in the test script, items should be a list with productId (camelCase) to match the API schema:

`items = [{"productId": product_id, "quantity": 2}]`

docker-compose.yml
```diff
@@ -779,6 +774,105 @@ services:
...
+      - CHATBOT_PORT
+      - OTEL_RESOURCE_ATTRIBUTES=${OTEL_RESOURCE_ATTRIBUTES},service.criticality=low
+      - OTEL_SERVICE_NAME=chatbot
+    depends_on:
+      agent:
+        condition: service_started
+    logging: *logging
+
+  # Postgresql used by Accounting service
+  postgresql:
```
ps48: `postgresql:` service key before astronomy-db. This may cause the docker compose to fail.

src/agent/src/agents/agents.py
```diff
@@ -0,0 +1,108 @@
...
+
+    async def get_tool_list(self):
+        mcp_enabled = os.getenv("MCP_ENABLED", "False") == "True"
+        if mcp_enabled and self.mcp_server is not None:
+            return await load_mcp_tools(self.mcp_server.session)
+        else:
+            tool_list = [
+                add_to_cart,
+                checkout,
+                convert_currency,
```
ps48: `convert_currency` is in the tool list but is never imported or defined. Will this crash at runtime when `MCP_ENABLED=False`

src/agent/src/agents/tools.py
```diff
@@ -0,0 +1,140 @@
...
+        return res.json()
+    except Exception as e:
+        return f"Error while fetching cart: {e}"
+
+
+def empty_cart(user_id: str):
+    """Empty the shopping cart for a user."""
+    url = f"http://{BASE_URL}/api/cart/empty/{user_id}"
+    try:
+        res = requests.get(url)
```
ps48: Should this be `DELETE /api/cart` with a JSON body `{"userId": user_id}`.

src/mcp/requirements.txt
```diff
@@ -0,0 +1,13 @@
...
+httpx==0.28.1
+langchain==1.0.3
+mcp==1.22.0
+fastmcp>=2.13.1
+fastapi==0.120.4
+opentelemetry-api==1.38.0
+opentelemetry-sdk==1.38.0
+opentelemetry-semantic-conventions==0.59b0
+opentelemetry-instrumentation-langchain<0.53.0
+fastmcp==2.13.1
```
ps48: can we please remove the dupe `fastmcp`

src/chatbot/requirements.txt
```diff
@@ -0,0 +1,5 @@
+requests==2.32.5
+dotenv==0.9.9
```
ps48: Should this be `python-dotenv` instead? `dotenv` I believe is deprecated. https://github.com/pedroburon/dotenv

.env
```diff
@@ -194,6 +194,24 @@ PROMETHEUS_PORT=9090
...
+GRAPH_RECURSION_LIMIT=25
+MCP_ENDPOINT=mcp
+MCP_PORT=8011
+MCP_ENABLED=False
+AGENT_ENDPOINT=agent
+AGENT_PORT=8010
+CHATBOT_PORT=7860
+LLM_BASE_URL=
+LLM_MODEL=
+USE_VCR=True
```
ps48: `.env` has `LLM_BASE_URL=` and `LLM_MODEL=` blank, and `USE_VCR=True` by default which expects VCR cassette files that are gitignored and don't exist. First-time users will get confusing errors. Please consider adding a comment section explaining the setup.

docker-compose.yml
```diff
@@ -779,6 +774,105 @@ services:
...
+      - APPLICATION_ENDPOINT=frontend:8080
+      - OTEL_RESOURCE_ATTRIBUTES=${OTEL_RESOURCE_ATTRIBUTES},service.criticality=low
+      - OTEL_SERVICE_NAME=mcp
+    depends_on:
+      agent:
+        condition: service_started
+    logging: *logging
+
+  # ChatBot
+  chatbot:
```
ps48: The chatbot service (port 7860) has no route configured in the envoy frontend-proxy, so it's only accessible via Docker's random port mapping. Other UI services (Jaeger, Grafana, LoadGen) are all exposed through port 8080.

.gitignore
```diff
@@ -61,4 +61,8 @@ src/shipping/target/
 test/tracetesting/tracetesting-vars.yaml
 
 !src/currency/build
+src/cart/src/Program.cs.bak
+src/flagd/demo.flagd.json.bak
```
ps48: Can we please remove these, if not needed?

fali007: Removed

docker-compose.yml
fali007: Added to envoy to expose chatbot on `http://localhost:8080/chatbot/`

.env
fali007: Added a comment to explain the `USE_VCR` env variable.

src/chatbot/requirements.txt
fali007: Changed this to `python-dotenv` in agent, chatbot and mcp

src/mcp/requirements.txt
fali007: Removed.

src/agent/src/agents/tools.py
fali007: Corrected this

src/agent/src/agents/agents.py
fali007: Missed this while removing the tool. Removed it now.

docker-compose.yml
fali007: Corrected this by removing

src/agent/src/agents/tools.py
fali007: For me get endpoint is what is giving response

```
def get_shipping_quote(items, currency_code, address):
    url = f"http://{BASE_URL}/api/shipping"

    if isinstance(items, dict):
        items = [items]

    normalised_items = []
    for it in items:
        if not isinstance(it, dict):
            return f"Error fetching shipping quote: invalid item {it!r}"
        product_id = it.get("productId") or it.get("product_id")
        quantity = it.get("quantity", 1)
        if not product_id:
            return "Error fetching shipping quote: each item must include productId"
        normalised_items.append({"productId": product_id, "quantity": quantity})

    params = {
        "itemList": json.dumps(normalised_items),
        "currencyCode": currency_code,
        "address": json.dumps(address),
    }
    try:
        res = requests.get(url, params=params)
        if not res.ok:
            body = res.text.strip() or "<empty body>"
            return (
                f"Shipping quote failed with HTTP {res.status_code}: {body}. "
                "Hint: itemList must be a JSON array of {productId, quantity} objects."
            )
        return res.json()
    except Exception as e:
        return f"Error fetching shipping quote: {e}"
```

src/agent/src/agents/tools.py
fali007: Corrected this

docker-compose.yml
```diff
@@ -779,6 +778,104 @@ services:
...
 
+  # Agent
+  agent:
+    image: ${IMAGE_NAME}:${DEMO_VERSION}-agent
+    container_name: agent
+    build:
+      context: ./
+      dockerfile: ${AGENT_DOCKERFILE}
+      cache_from:
+        - ${IMAGE_NAME}:${IMAGE_VERSION}-agent
```
ps48: Heads up — when I ran `make start`, the build reused layers from the published `2.2.0-agent` image, which still has the old `convert_currency` import. The first `/prompt` then crashes with `NameError: name 'convert_currency' is not defined`, even though the source on this branch is correct.

Could we drop `cache_from` for the new services (agent here, plus mcp at L829 and chatbot at L858) until clean images are published, or pin to a tag we know is in sync? Happy to help test if useful.

src/agent/src/agents/llm.py
```diff
@@ -0,0 +1,41 @@
...
+        use_vcr = os.getenv("USE_VCR", False) == "True"
+        llm_tls_verify = os.getenv("LLM_TLS_VERIFY", True) == "True"
+        cassette_name = ""
+        if use_vcr:
+            cassette_name = f"{model_name.replace('/', '_')}_casette.yaml"
+        if "http_async_client" not in kwargs:
+            kwargs["http_async_client"] = httpx.AsyncClient(verify=llm_tls_verify)
+        kwargs.setdefault("openai_api_base", os.getenv("LLM_BASE_URL"))
+        kwargs.setdefault("model", model_name)
+        kwargs.setdefault("api_key", os.getenv("API_KEY"))
```
ps48: On a fresh container the agent crashed at startup with `OpenAIError: Missing credentials`. It looks like `ChatOpenAI` (`langchain_openai==1.0.1`) reads `OPENAI_API_KEY` from the environment during pydantic validation, before the `kwargs.setdefault("api_key", ...)` we pass in is checked.

Would you be open to either renaming the env var to `OPENAI_API_KEY` in `.env`/compose, or also exporting `OPENAI_API_KEY=${API_KEY}` in the agent service block? That should make the demo work end-to-end without a manual step.

src/agent/src/agents/patch_vcr.py
```diff
@@ -0,0 +1,65 @@
...
+
+def clean_response(response):
+    headers = response["headers"]
+    headers.pop("content-encoding", None)
+    headers.pop("transfer-encoding", None)
+    return response
+
+
+VCR = vcr.VCR(
+    cassette_library_dir="fixtures/vcr_cassettes",
```
ps48: I noticed the agent Dockerfile copies `src/agent/src` and `run.py` but not `src/agent/fixtures/`, and the compose service doesn't mount it either. Since `cassette_library_dir` is resolved relative to `/app`, any cassette a user drops into `src/agent/fixtures/vcr_cassettes/` is invisible to the container — and with `record_mode="new_episodes"`, VCR ends up writing a fresh cassette inside the ephemeral container fs.

Could we either bind-mount `./src/agent/fixtures:/app/fixtures` in compose or `COPY` the directory in the Dockerfile, and mention the location in the README? This one took me a while to debug, so wanted to flag.

src/agent/src/agents/llm.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class ChatLLM(ChatOpenAI):
+    def __init__(self, **kwargs):
+        model_name = os.getenv("LLM_MODEL", "default")
+        use_vcr = os.getenv("USE_VCR", False) == "True"
+        llm_tls_verify = os.getenv("LLM_TLS_VERIFY", True) == "True"
+        cassette_name = ""
+        if use_vcr:
+            cassette_name = f"{model_name.replace('/', '_')}_casette.yaml"
```
ps48: Tiny one — looks like a typo (single `s`, should be `cassette`). Worth fixing now before any cassette filenames get baked in.

ps48 (commented):
Hey @fali007 — really cool addition, the agent + MCP + chatbot trio is a lovely demo of what an instrumented LLM workflow looks like. I tried to run it end-to-end against the cassette today and ran into a few rough edges; leaving notes inline below.

A few of these (the `cache_from` / `API_KEY` / cassette-mount ones) are blockers I'd suggest tackling first since they're what kept me from getting a clean replay. The rest are mostly polish / robustness suggestions — totally happy to chat through any of them. Thanks for the work on this!

docker-compose.yml
```diff
@@ -779,6 +778,104 @@ services:
...
+          memory: 500M
+    restart: unless-stopped
+    ports:
+      - "${AGENT_PORT}"
+    environment:
+      - OTEL_EXPORTER_OTLP_ENDPOINT=${OTEL_COLLECTOR_HOST}:${OTEL_COLLECTOR_PORT_GRPC}
+      - GRAPH_RECURSION_LIMIT
+      - MCP_ENDPOINT=mcp
+      - MCP_PORT=8011
+      - MCP_ENABLED=True
```
ps48: `MCP_ENABLED` and `USE_VCR` are hardcoded to `True` here (and L807 below), but `.env` defaults to `MCP_ENABLED=False` / `USE_VCR=True`. The hardcoded values silently win, so toggling `.env` has no visible effect — confusing for users following the README.

Could we use `${MCP_ENABLED}` / `${USE_VCR}` here so the `.env` and `.env.override` files actually drive behavior? It'd make the configuration story easier to follow.

src/agent/run.py
```diff
@@ -0,0 +1,37 @@
...
+import os
+
+from dotenv import load_dotenv
+from src.agents.agents import Agent
+from traceloop.sdk import Traceloop
+
+logging.basicConfig(level=logging.INFO)
+
+load_dotenv()
+
```
ps48: Since the demo's main goal is showing OTel in action, it'd be great if the new services emitted spans for their HTTP path too. Right now `Traceloop.init` is doing the LangChain instrumentation, but inbound `/prompt` requests and outbound `requests.*` calls to `product-catalog` / `cart` / `frontend` aren't traced.

Adding `opentelemetry-instrumentation-fastapi` and `opentelemetry-instrumentation-requests` (and the equivalent in `chatbot/run.py`) would close the loop and make the agent flow appear as one connected trace in Jaeger. Happy to suggest snippets if helpful.

src/agent/run.py
```diff
@@ -0,0 +1,37 @@
...
+Traceloop.init(
+    app_name="AstronomyShopAgent",
```
ps48: With `OTEL_SERVICE_NAME=agent` set in compose, the agent will show up under both names in Jaeger — `AstronomyShopAgent` (Traceloop spans) and `agent` (FastAPI/requests spans, once those are instrumented). Traceloop respects `OTEL_SERVICE_NAME`, so we could drop the `app_name=` kwarg here (and the equivalent in `mcp/run.py`) and let the env var win. Keeps the trace view tidy.

src/agent/src/agents/tools.py
```diff
@@ -0,0 +1,175 @@
+#!/usr/bin/python
+
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+import json
+import os
+
+import requests
```
ps48: Each tool function uses sync `requests` inside async FastAPI handlers, so every tool call blocks the uvicorn event loop while waiting on the downstream service. Under any concurrency, prompts will serialize.

`httpx.AsyncClient` is already in `requirements.txt` — would you be open to switching the tools over, or wrapping the calls in `asyncio.to_thread`? Either approach should be fine.

src/agent/src/agents/tools.py
```diff
@@ -0,0 +1,175 @@
...
+BASE_URL = os.getenv("APPLICATION_ENDPOINT", "localhost:8080")
+
+
+def get_ads(category: str):
+    """Fetch promotional ads for Astronomy Shop homepage.
+    Eg : category: `telescopes` or `travel`"""
+    url = f"http://{BASE_URL}/api/data"
+    params = {"contextKeys": category}
+    try:
+        res = requests.get(url, params=params)
```
ps48: None of the `requests.get/post/delete` calls in this file pass a `timeout=`. If a downstream is slow the worker hangs indefinitely. Could we add a default (and optionally make it env-tunable)? Something like `requests.get(url, timeout=10)` would do the job.

src/agent/src/agents/patch_vcr.py
```diff
@@ -0,0 +1,65 @@
...
+    record_mode="new_episodes",
```
ps48: With `match_on=["method","uri","body"]`, any small change to a tool signature or the system prompt causes a body mismatch, which silently appends a new episode to the cassette and also hits the live LLM. For a demo aimed at deterministic replay, would `"none"` (or `"once"`) be a better default, with explicit re-recording when fixtures need to refresh? Just a thought — happy to chat through trade-offs.

.env
```diff
@@ -194,6 +194,29 @@ PROMETHEUS_PORT=9090
...
+# USE_VCR flags helps in caching the LLM requests and responses which will reuse the response in the future. 
+# Turn this to False for calling configured LLM always
+USE_VCR=True
+LLM_TLS_VERIFY=True
+AGENT_DOCKERFILE=src/agent/Dockerfile
+MCP_DOCKERFILE=src/mcp/Dockerfile
+CHATBOT_DOCKERFILE=src/chatbot/Dockerfile
+AGENT_ADDR=${AGENT_ENDPOINT}:${AGENT_PORT}
+APPLICATION_ENDPOINT=frontend:8080
+OTEL_EXPORTER_OTLP_INSECURE=true
```
ps48: This makes `OTEL_EXPORTER_OTLP_INSECURE=true` a global default for the whole demo. Since only the new Python services need it (because `traceloop-sdk` reads OTEL env vars directly), would it be cleaner to scope it to the agent/mcp/chatbot service blocks in `docker-compose.yml`? Avoids accidentally affecting the rest of the stack.

src/agent/src/agents/agents.py
```diff
@@ -0,0 +1,107 @@
...
+    get_shipping_quote,
+    get_supported_currencies,
+    list_products,
+)
+from traceloop.sdk.decorators import workflow
+
+
+class ChatRequest(BaseModel):
+    message: str
+    history: List[Dict] = []
```
ps48: The chatbot UI sends a `history` field, but `handle_prompt` only forwards `request.message` (line 60), so the conversation is effectively single-turn even though the UI implies multi-turn.

Either threading history into `agent.ainvoke({"messages": [...]})` or removing the field would make the contract match the behavior — whichever direction you'd prefer.

src/chatbot/src/chat_interface/chat_interface.py
```diff
@@ -0,0 +1,95 @@
...
+            server_port=self.config.uiPort,
+            root_path=self.config.rootPath,
+        )
+
+
+def get_chat_ui_config():
+    chat_ui_config = ChatUiConfig(
+        uiBaseUrl=os.getenv("CHATBOT_ENDPOINT", "0.0.0.0"),
+        uiPort=int(os.getenv("CHATBOT_PORT", "7860")),
+        sessionId=str(uuid.uuid4()),
```
ps48: This generates one `session_id` per chatbot process, so concurrent demo users would share state. Gradio exposes `request: gr.Request` in handler functions with `request.session_hash` — would it be reasonable to use that for a per-session ID?

src/agent/src/agents/patch_vcr.py
```diff
@@ -0,0 +1,65 @@
...
+        content = response.read()
+
+    return {
+        "status": {"code": response.status_code, "message": response.reason_phrase},
+        "headers": dict(response.headers),
+        "body": {"string": content},
+    }
+
+
+vcr.stubs.httpx_stubs._to_serialized_response = patched_to_serialized_response
```
ps48: The `httpx_stubs._to_serialized_response` patch is applied at import time regardless of `USE_VCR`, and is fragile across vcrpy versions. Could we gate it behind an `if USE_VCR:` check so the agent doesn't carry a patched HTTP stack when VCR isn't in use?

src/agent/src/agents/mcp_client.py
```diff
@@ -0,0 +1,28 @@
...
+        read, write, _ = await self.exit_stack.enter_async_context(stream_context)
+        session_context = ClientSession(read, write)
+        self.session = await self.exit_stack.enter_async_context(session_context)
+        await self.session.initialize()
+
+    async def cleanup(self):
+        try:
+            await self.exit_stack.aclose()
+        except Exception as e:
+            print(f"Error closing connection : {e}")
```
ps48: Tiny one — switching to `logging.error(...)` would keep this consistent with the rest of the codebase and feed into the OTel log pipeline.

docker-compose.yml
fali007: Made the change.

src/agent/src/agents/patch_vcr.py
fali007: Added 
```
volumes:
      - ./src/agent/fixtures:/app/fixtures
```
Because I think adding to image might increase the size of image. I can provide baseline cassettes in a followup PR when we onboard agentic load generator.

src/agent/src/agents/llm.py
fali007: Made the change by setting `OPENAI_API_KEY=${API_KEY}`.

src/agent/src/agents/llm.py
fali007: Thanks, Made the change.

docker-compose.yml
fali007: Made the change

src/agent/run.py
fali007: Sure. Can you help me with the scripts?

src/agent/run.py
fali007: I will remove `AstronomyShopAgent` `app_name` for agent and MCP once I add `FastAPI/requests` instrumentation.

src/agent/src/agents/tools.py
fali007: Converted to httpx client.

src/agent/src/agents/tools.py
fali007: Converted to 
```
TIMEOUT = httpx.Timeout(10.0)
```

src/agent/src/agents/patch_vcr.py
fali007: Ive changed it to once. Do you think making it an ENV will provide more value?

.env
fali007: Sure. Ive removed it from .env

src/agent/src/agents/agents.py
fali007: Ive kept the history for future additions when we have multi-turn load generator which I've been thinking about. Apparently the current load-generator doesn't support multi turn load.
It makes sense to add history to messages since users can still use chatbot to have multi turn conversations.

src/chatbot/src/chat_interface/chat_interface.py
fali007: Yes. Ive changed it accordingly

src/agent/src/agents/patch_vcr.py
fali007: Added the check.

src/agent/src/agents/mcp_client.py
fali007: Changed to `logging.error`

src/agent/run.py
fali007: Ive added a v1, can you validate it once?

src/agent/run.py
fali007: I am not able to get the service name in the trace. Will remove it once I find a solution.

.env
```diff
@@ -194,6 +194,28 @@ PROMETHEUS_PORT=9090
...
+AGENT_PORT=8010
+CHATBOT_HOST=chatbot
+CHATBOT_PORT=7860
+CHATBOT_ROOT_PATH=/chatbot
+LLM_BASE_URL=
+LLM_MODEL=
+API_KEY=
+# USE_VCR flags helps in caching the LLM requests and responses which will reuse the response in the future. 
+# Turn this to False for calling configured LLM always
+USE_VCR=True
```
ps48: Default `.env` is internally inconsistent and breaks the out-of-box demo. With `USE_VCR=True` but `API_KEY=`, `LLM_BASE_URL=`, and `LLM_MODEL=` all empty, the agent crashes on the first request with `openai.OpenAIError: Missing credentials` because `ChatOpenAI` validates creds eagerly before VCR replay can happen.

If the intent is "works without a real LLM out of the box", default a non-empty placeholder (e.g. `API_KEY=dummy`) and pre-fill `LLM_MODEL`/`LLM_BASE_URL` to match one of the recorded cassettes. Otherwise the cassette files never get exercised on `make start`. Verified locally.

src/agent/src/agents/llm.py
```diff
@@ -0,0 +1,41 @@
...
+import httpx
+from langchain_openai import ChatOpenAI
+from src.agents.patch_vcr import VCR
+
+
+class ChatLLM(ChatOpenAI):
+    def __init__(self, **kwargs):
+        model_name = os.getenv("LLM_MODEL", "default")
+        use_vcr = os.getenv("USE_VCR", False) == "True"
+        llm_tls_verify = os.getenv("LLM_TLS_VERIFY", True) == "True"
```
ps48: Bug: when `LLM_TLS_VERIFY` is unset, `os.getenv("LLM_TLS_VERIFY", True)` returns the **bool** `True`, then `True == "True"` is `False`, so `llm_tls_verify` becomes `False` and TLS verification is silently disabled — the opposite of what the README documents.

Use a string default and case-insensitive compare, e.g. `os.getenv("LLM_TLS_VERIFY", "True").lower() == "true"`. Same pattern applies to the `USE_VCR` check on line 16 (`MCP_ENABLED` in `agents.py` is also affected).

src/agent/src/agents/llm.py
```diff
@@ -0,0 +1,41 @@
...
+        cassette_name = ""
+        if use_vcr:
+            cassette_name = f"{model_name.replace('/', '_')}_cassette.yaml"
+        if "http_async_client" not in kwargs:
+            kwargs["http_async_client"] = httpx.AsyncClient(verify=llm_tls_verify)
+        kwargs.setdefault("openai_api_base", os.getenv("LLM_BASE_URL"))
+        kwargs.setdefault("model", model_name)
+        kwargs.setdefault("api_key", os.getenv("API_KEY"))
+        super().__init__(**kwargs)
```
ps48: `super().__init__(**kwargs)` calls `ChatOpenAI`'s validator, which raises `OpenAIError: Missing credentials` when `API_KEY` is empty — before `_generate`/`_agenerate` ever run. So `USE_VCR=True` alone does not bypass the LLM credential check; the agent never reaches replay.

Either (a) inject a dummy `api_key` when `use_vcr` is true, or (b) skip parent `__init__` and stub the OpenAI client when in VCR mode. Without one of these the cassette path is unreachable on a fresh checkout.

src/agent/src/agents/llm.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class ChatLLM(ChatOpenAI):
+    def __init__(self, **kwargs):
+        model_name = os.getenv("LLM_MODEL", "default")
+        use_vcr = os.getenv("USE_VCR", False) == "True"
+        llm_tls_verify = os.getenv("LLM_TLS_VERIFY", True) == "True"
+        cassette_name = ""
+        if use_vcr:
+            cassette_name = f"{model_name.replace('/', '_')}_cassette.yaml"
```
ps48: Cassette name derived from `LLM_MODEL.replace('/', '_')` is case-sensitive on Linux: `LLM_MODEL=Azure/gpt-4o` -> `Azure_gpt-4o_cassette.yaml` but `azure/gpt-5.5` -> `azure_gpt-5.5_cassette.yaml`. The two files in this PR differ only by case. Either lower-case before mapping, or document the exact `LLM_MODEL` strings in the README so users don't end up missing-cassette.

src/agent/src/agents/agents.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class Agent:
+    def __init__(self):
+        self.app = FastAPI(lifespan=self.lifespan)
+        self.app.post("/prompt")(self.handle_prompt)
+        self.agentRecursionLimit = int(os.getenv("GRAPH_RECURSION_LIMIT", "25"))
```
ps48: `self.agentRecursionLimit` is read from `GRAPH_RECURSION_LIMIT` here but never passed to `agent.ainvoke(...)` on line 92, so the env var has no effect. Pass it through, e.g. `await agent.ainvoke({"messages": history}, config={"recursion_limit": self.agentRecursionLimit})`.

src/agent/src/agents/agents.py
```diff
@@ -0,0 +1,108 @@
...
+                get_product,
+                get_recommendations,
+                get_shipping_quote,
+                get_supported_currencies,
+                list_products,
+            ]
+            return [tool(t) for t in tool_list]
+
+    @workflow(name="astronomy_shop_agent_workflow")
+    async def run_agent(self, input_prompt, history=[]):
```
ps48: Mutable default argument: `history=[]` is shared across all calls to `run_agent`. Combined with `history.append(...)` on line 91 this means each request appends to the previous request's list. Use `history: List[Dict] | None = None` and initialize per call.

src/agent/src/agents/agents.py
```diff
@@ -0,0 +1,108 @@
...
+            system_prompt="You are a helpful assistant. Be concise and accurate.",
+        )
+        try:
+            history.append({"role": "user", "content": input_prompt})
+            result = await agent.ainvoke(
+                {"messages": history}
+            )
+            return {"response": result}
+        except Exception as e:
+            raise HTTPException(status_code=500, detail=str(e))
```
ps48: Re-raising as `HTTPException(500, detail=str(e))` discards the traceback and the original exception type. Since this is decorated with `@workflow`, the trace span loses the failure context too. Consider `raise HTTPException(...) from e` or letting Traceloop record the exception via `trace.get_current_span().record_exception(e)` before raising.

src/agent/src/agents/patch_vcr.py
```diff
@@ -0,0 +1,67 @@
...
+
+def clean_response(response):
+    headers = response["headers"]
+    headers.pop("content-encoding", None)
+    headers.pop("transfer-encoding", None)
+    return response
+
+
+VCR = vcr.VCR(
+    cassette_library_dir="fixtures/vcr_cassettes",
```
ps48: `cassette_library_dir="fixtures/vcr_cassettes"` is relative to the process CWD. It only resolves because the Dockerfile happens to set `WORKDIR /app` and compose mounts the fixtures volume there. Local `python run.py` runs from `src/agent/` and won't find the cassettes.

Make it absolute: `cassette_library_dir=os.path.join(os.path.dirname(os.path.dirname(__file__)), "..", "fixtures", "vcr_cassettes")` (or similar).

src/agent/src/agents/patch_vcr.py
```diff
@@ -0,0 +1,67 @@
...
+    record_mode="new_episodes",
```
ps48: `record_mode="new_episodes"` silently falls through to the real network whenever a request body doesn't match a recorded one — turning what should be a deterministic demo into a flaky outbound call.

Verified locally: the recorded `tools` block in `Azure_gpt-4o_cassette.yaml` contains a `convert_currency` function that's no longer in `tools.py`, so live requests never match. With `new_episodes`, VCR then makes a real call to `ete-litellm.bx.cloud9.ibm.com`, which from CI/most users just hangs.

Suggest `record_mode=os.getenv("VCR_RECORD_MODE", "none")` for the demo and re-record cassettes against the current tools list. With `match_on=["method","uri","body"]`, body equality is also fragile against any tool-description tweak — consider a custom matcher that ignores volatile fields (`tools`, `x-stainless-*`) once you re-record.

src/agent/src/agents/patch_vcr.py
```diff
@@ -0,0 +1,67 @@
...
+    except Exception:
+        pass
+    return request
+
+
+def clean_response(response):
+    headers = response["headers"]
+    headers.pop("content-encoding", None)
+    headers.pop("transfer-encoding", None)
+    return response
```
ps48: Before publishing these cassettes upstream, please scrub provider metadata that's leaking in the recorded responses:

- response headers: `x-litellm-call-id`, `x-litellm-key-max-budget`, `x-litellm-key-spend`, `x-litellm-model-id`, `x-litellm-model-api-base`, `x-litellm-response-cost*`, `llm_provider-azureml-model-session`, `x-ratelimit-*`
- request `host` header / `uri`: rewrite from `ete-litellm.bx.cloud9.ibm.com` to a generic `api.openai.com` (or similar) so users aren't directed at IBM infra when records are bypassed
- `x-stainless-*` request headers (OS, runtime version, package version) — they tie the cassette to the recording machine and make body-matching brittle

Easiest fix: add the response-header scrubbing here in `clean_response`, and add a `before_record_request` step that rewrites the URI.

ps48 (commented):
Inline comments on the agent/MCP/chatbot additions plus VCR validation. End-to-end test summary:

- Built and ran `agent`, `mcp`, `chatbot` from this branch.
- With committed defaults (`USE_VCR=True`, `API_KEY=`, `LLM_BASE_URL=`, `LLM_MODEL=`), the first `POST /prompt` 500s with `openai.OpenAIError: Missing credentials` from inside `ChatLLM()` — `ChatOpenAI` validates credentials eagerly, before VCR can replay anything.
- Setting placeholder credentials (`API_KEY=dummy`, `LLM_BASE_URL=https://ete-litellm.bx.cloud9.ibm.com`, `LLM_MODEL=Azure/gpt-4o`) lets the agent boot, but the cassettes never match: their recorded `tools` block contains a `convert_currency` tool that is **not** in the current `tools.py`, so the request body diverges. With `record_mode="new_episodes"`, VCR silently falls through to the real IBM endpoint.
- I verified the VCR mechanics work — feeding the cassette's exact recorded body through `VCR.use_cassette(...)` returns a recorded 200 with no network call. The replay path itself is fine; the cassettes are stale relative to the shipped tools.
- Cassettes also contain internal hostnames/metadata (`ete-litellm.bx.cloud9.ibm.com`, `eastus2.api.cognitive.microsoft.com`, `x-litellm-*`, `llm_provider-azureml-model-session`, `x-litellm-key-spend`, etc.) that should be sanitized before merge to upstream.

src/agent/run.py
```diff
@@ -0,0 +1,44 @@
...
+    api_endpoint=os.getenv("OTEL_EXPORTER_OTLP_ENDPOINT", "localhost:4317"),
+)
+
+RequestsInstrumentor().instrument()
+HTTPXClientInstrumentor().instrument()
```
ps48: `Traceloop.init` already instruments `httpx` and `requests` internally. Calling `RequestsInstrumentor().instrument()` and `HTTPXClientInstrumentor().instrument()` afterwards produces a `WARNING:opentelemetry.instrumentation.instrumentor: Attempting to instrument while already instrumented` at startup (visible in `docker logs agent`). Drop these two lines or guard with `is_instrumented_by_opentelemetry`.

src/chatbot/run.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def _configure_tracing() -> None:
+    resource = Resource.create(
+        {
+            "service.name": os.getenv("OTEL_SERVICE_NAME", "chatbot"),
+        }
+    )
+    provider = TracerProvider(resource=resource)
+    provider.add_span_processor(BatchSpanProcessor(OTLPSpanExporter()))
```
ps48: `OTLPSpanExporter()` defaults to `http://localhost:4317`. The `chatbot` service in `docker-compose.yml` does **not** set `OTEL_EXPORTER_OTLP_ENDPOINT`, so traces from the chatbot will silently never reach the collector.

Either pass `endpoint=os.getenv("OTEL_EXPORTER_OTLP_ENDPOINT", ...)` here, or (preferred for consistency with `agent`/`mcp`) add `OTEL_EXPORTER_OTLP_ENDPOINT=${OTEL_COLLECTOR_HOST}:${OTEL_COLLECTOR_PORT_GRPC}` and `OTEL_EXPORTER_OTLP_INSECURE=true` to the `chatbot` service env block.

src/chatbot/src/chat_interface/chat_interface.py
```diff
@@ -0,0 +1,94 @@
...
+class ChatAgentUI:
+    def __init__(self, chat_ui_config: ChatUiConfig):
+        self.config = chat_ui_config
+
+    def chat_with_agent(self, message, history, request: gr.Request):
+        try:
+            session_id = request.session_hash or self.config.sessionId
+            payload = {
+                "message": message,
+                "session_id": session_id,
```
ps48: `session_id` is sent in the payload but the agent's `ChatRequest` model only declares `message` and `history`. With Pydantic strict-extra-fields it would 422; today it's silently dropped. Either add `session_id` to `ChatRequest` and use it for memory/threading, or remove it here.

src/chatbot/src/chat_interface/chat_interface.py
```diff
@@ -0,0 +1,94 @@
...
+
+        chatbot_ui = gr.ChatInterface(
+            fn=self.chat_with_agent,
+            title=config["title"],
+            description=config["description"],
+            examples=config["examples"],
+            chatbot=gr.Chatbot(height="70vh"),
+        )
+
+        chatbot_ui.launch(
```
ps48: Gradio's default `launch()` enables analytics, which causes the chatbot to call out to `api.gradio.app` and `huggingface.co/api/telemetry/*` on every startup (verified in container logs). For an upstream OpenTelemetry demo this looks bad. Pass `analytics_enabled=False` to `chatbot_ui.launch(...)` and/or set `GRADIO_ANALYTICS_ENABLED=False` in the compose env.

src/frontend-proxy/envoy.tmpl.yaml
```diff
@@ -54,17 +57,34 @@ static_resources:
...
+                        - match: { path: "/chatbot" }
+                          redirect: { path_redirect: "/chatbot/" }
+                        - match: { prefix: "/chatbot/" }
+                          route:
+                            cluster: chatbot
+                            prefix_rewrite: "/"
+                            timeout: 0s
+                            upgrade_configs:
+                              - upgrade_type: websocket
+                        - match: { prefix: "/" }
```
ps48: Stray `- match: { prefix: "/" }` with no route/redirect action. Either it's a no-op (best case) or, depending on Envoy version, fails route validation. Looks like a merge artifact — remove this line.

Also, the `/profiles` block right below appears to be from #3333 (firepit profiles) and seems out of scope for this PR.

src/mcp/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    build-essential \
+    && rm -rf /var/lib/apt/lists/*
+
+COPY src/mcp/requirements.txt .
+
+RUN pip install --no-cache-dir -r requirements.txt
+
+COPY src/mcp/src src
+COPY src/agent/src/agents/tools.py src/mcp_server
```
ps48: Cross-package copy: `COPY src/agent/src/agents/tools.py src/mcp_server` makes the MCP image silently dependent on a file in another service's tree. If `tools.py` is edited and only the agent image is rebuilt, MCP keeps serving stale tool definitions with no warning.

Move the shared tools to a top-level shared module (e.g. `src/shared/tools.py`) that both Dockerfiles `COPY` from, or publish them as a small package both services install. As-is, this is the same drift problem the cassette has.

docker-compose.yml
```diff
@@ -781,6 +780,101 @@ services:
...
+    restart: unless-stopped
+    ports:
+      - "${AGENT_PORT}"
+    volumes:
+      - ./src/agent/fixtures:/app/fixtures
+    environment:
+      - OTEL_EXPORTER_OTLP_ENDPOINT=${OTEL_COLLECTOR_HOST}:${OTEL_COLLECTOR_PORT_GRPC}
+      - GRAPH_RECURSION_LIMIT
+      - MCP_ENDPOINT=mcp
+      - MCP_PORT=8011
```
ps48: `MCP_ENDPOINT=mcp` and `MCP_PORT=8011` are hardcoded here even though both are already defined in `.env`. Pass the variable name only (`- MCP_ENDPOINT`, `- MCP_PORT`), matching the surrounding entries — otherwise changing the port in `.env` won't propagate.

docker-compose.yml
```diff
@@ -781,6 +780,101 @@ services:
...
+      - MCP_ENABLED=${MCP_ENABLED}
+      - AGENT_ENDPOINT
+      - AGENT_PORT
+      - LLM_BASE_URL
+      - LLM_MODEL
+      - API_KEY
+      - USE_VCR=${USE_VCR}
+      - OPENAI_API_KEY=${API_KEY}
```
ps48: `API_KEY` (line 809) and `OPENAI_API_KEY=${API_KEY}` here are duplicated for the same value. The agent code only reads `API_KEY`; this extra `OPENAI_API_KEY` is just propagating the secret to a second var that nothing reads. Drop one.

docker-compose.yml
```diff
@@ -781,6 +780,101 @@ services:
...
+      - OTEL_EXPORTER_OTLP_ENDPOINT=${OTEL_COLLECTOR_HOST}:${OTEL_COLLECTOR_PORT_GRPC}
+      - MCP_ENDPOINT
+      - MCP_PORT
+      - OTEL_EXPORTER_OTLP_INSECURE=true
+      - APPLICATION_ENDPOINT=frontend:8080
+      - OTEL_RESOURCE_ATTRIBUTES=${OTEL_RESOURCE_ATTRIBUTES},service.criticality=low
+      - OTEL_SERVICE_NAME=mcp
+    depends_on:
+      agent:
+        condition: service_started
```
ps48: `mcp` only depends on `agent`, but the MCP tools (copied from `src/agent/src/agents/tools.py`) call `frontend:8080` (cart, products, recommendations, etc.). On a cold start MCP can come up before `frontend`/`product-catalog` and the first prompt fails with connection errors. Add `frontend` (or at minimum `product-catalog`) to `depends_on`.

docker-compose.yml
```diff
@@ -781,6 +780,101 @@ services:
...
+    restart: unless-stopped
+    ports:
+      - "${CHATBOT_PORT}"
+    environment:
+      - AGENT_PORT
+      - AGENT_ENDPOINT
+      - CHATBOT_PORT
+      - CHATBOT_ROOT_PATH
+      - OTEL_RESOURCE_ATTRIBUTES=${OTEL_RESOURCE_ATTRIBUTES},service.criticality=low
+      - OTEL_SERVICE_NAME=chatbot
```
ps48: `chatbot` env block is missing `OTEL_EXPORTER_OTLP_ENDPOINT` (and `OTEL_EXPORTER_OTLP_INSECURE`). Combined with `chatbot/run.py` calling bare `OTLPSpanExporter()`, traces from the chatbot service are exported to `http://localhost:4317` inside its own container and dropped. Add the same two env entries the `agent` service has.

fali007: Missed this. Added to docker compose file.

docker-compose.yml
fali007: Yeah. You are right. Ive made the change

docker-compose.yml
fali007: Ive added this to address this issue {[link](https://github.com/open-telemetry/opentelemetry-demo/pull/3148/#discussion_r3232045985)}

docker-compose.yml
fali007: Sure. Ive made the change

fali007:
`gpt-4o` cassette is incomplete. Please use model as  `gpt-5.5` or `claude-opus-4-7`

src/agent/src/agents/llm.py
fali007: MCP_ENABLED uses "False" as default and can ge changed to "True". Ive made the other change.

.env
fali007: Made the change. Added a dummy value which works out of the box. Made the change to match only with request body and remove the url used in the request while saving cache to prevent comparing it in the future.

src/agent/src/agents/llm.py
fali007: Added dummy API_KEY in .env which will resolve this issue.

fali007:
```LLM_BASE_URL=https://local-llm.com
# set LLM_MODEL as azure/gpt-5.5 or claude-opus-4-7 to reuse resposes in VCR cache
LLM_MODEL=
API_KEY=sk-dummy
```

Set model to one of the above and API_KEY can be _unset_ as well. It will reuse the responses from respective cassette file.

src/agent/src/agents/llm.py
fali007: Ive added this to readme.

src/agent/src/agents/agents.py
fali007: Sorry. I missed this. Added the suggested fix.

src/agent/src/agents/agents.py
fali007: Made the change

src/agent/src/agents/agents.py
fali007: Added the change by raising the exception from e

src/agent/src/agents/patch_vcr.py
fali007: Do we plan on supporting running this locally using python ?

src/agent/src/agents/patch_vcr.py
fali007: Added this change and replaced old cache with new.

src/agent/src/agents/patch_vcr.py
fali007: Made the change and tested.

src/agent/run.py
fali007: Dropped those two lines

src/chatbot/run.py
fali007: Made the change.

src/chatbot/src/chat_interface/chat_interface.py
fali007: Commented the `session_id`. Kept it if it might be required in the future enhancements.

src/chatbot/src/chat_interface/chat_interface.py
fali007: Added `analytics_enabled=False` .

src/frontend-proxy/envoy.tmpl.yaml
fali007: Removed the line and placed it under the fire pit change.

src/mcp/Dockerfile
fali007: I agree with the VCR problem. Ive moved the tools under shared folder.

fali007:
There is a minor change with the tool list when MCP enabled and MCP disabled mode. That was the reason for cache miss. Ive added cache for both cases in the current cassettes.

src/agent/run.py
fali007: Once I drop `HTTPXClientInstrumentor().instrument()`, I see the spans missing in MCP_ENABLED=False mode. Ive removed the `requests` instrumentation but kept `httpx`. I don't see the warning when httpx is kept.

.github/workflows/assign-reviewers.yml
osullivandonal, line 1: Are the changes to this folder (.github/workflows) part of the work to add agent astronomy shop and MCP server? Considering there are a lot of changes in this PR, could we ensure that this PR is atomic and only includes the required changes in it? This helps to reduce the PR reviewers burden.

fali007:
Closing this PR will create new one.

PR #3185, opened 2026-04-01
o11y(email): remove recipient email from telemetry

niwoerner:
# Changes

The semantic convention recommend to avoid capturing PII data unless absolutely necessary. ([source](https://opentelemetry.io/docs/security/handling-sensitive-data/#sensitive-data-considerations) 

This PR stops capturing recipient email addresses in telemetry and uses order.id for correlation instead, which should be sufficient for troubleshooting.


puckpuck:
Part of me wants to keep the email attribute on the span data, and add a redaction processor to the collector instead. I think adding a comment in code for the email service that this is deliberate would be a good thing too.

Removing the entry in the log makes sense though since that's kinda unstructured and uglier to clean up in a collector.

Kielek (approved):
@puckpuck, I would say that we should show good pattern only here. If you would like to show some sensitive data redaction by the collector, we could introduce separate, fake "sensitive.data.demo.fake.attribute" and use it for such purposes.

I think that someone can inspire/copy paste the email example to the application.

cyrille-leclerc:
I like @Kielek’s proposal: it highlights a common real-world but undesirable use case, shows how to mitigate it, and clearly sets the expectation that it is an antipattern.

puckpuck:
Yeah, let's do that. I'll create an issue for us to add a clearly fake but should be redacted attribute, and a processor to redact it.

PR #3107, opened 2026-03-13
Add full and minimal docker profiles

kerneljack:
# Changes

This adds on to the work done by @julianocosta89 in [#2396](https://github.com/open-telemetry/opentelemetry-demo/pull/2396) to add **docker profiles** to the Demo app.

I have tested the functionality of the Demo app itself and the Grafana Demo dashboard using both the **full** and **minimal** profiles.

You can use `make start` for the **full** profile and `make start-minimal` for the **minimal** profile.

I have attempted to fix the tests and most of them work except for one.

## Merge Requirements

For new features contributions, please make sure you have completed the following
essential items:

* [x] `CHANGELOG.md` updated to document new feature additions
* [ ] Appropriate documentation updates in the [docs][]
* [ ] Appropriate Helm chart updates in the [helm-charts][]

Maintainers will not merge until the above have been completed. If you're unsure
which docs need to be changed ping the
[@open-telemetry/demo-approvers](https://github.com/orgs/open-telemetry/teams/demo-approvers).

[docs]: https://opentelemetry.io/docs/demo/
[helm-charts]: https://github.com/open-telemetry/opentelemetry-helm-charts

PR #3229, opened 2026-04-09
feat(docker): modular Docker Compose with layered -f files

### LINKED ISSUE #2195: Create OpenTelemetry Demo Light Brach
# Feature Request

The current OpenTelemetry demo requires significant resources to run all services, making it difficult to deploy in environments with limited capacity—such as GitHub Codespaces—without using a large instance.
The purpose of this feature request is to create a new branch In the project, named opentelemetry-demo-light providing a lighter version of the demo.

It would be helpful to provide a lightweight version of the demo that includes only a minimal set of core services, such as:

- frontend
- productcatalogservice
- cartservice
- checkoutservice
- paymentservice

This would make it easier for users to explore OpenTelemetry in constrained environments and lower the barrier to entry for experimentation and learning.

puckpuck:
Closes #2195
Supersedes #3107 (takes a `-f` layering approach instead of profiles)

## Summary

Replaces the monolithic `docker-compose.yml` with four composable files:

| File | Contents |
|---|---|
| `compose.yaml` | Core/minimal services — always the base |
| `compose.full.yaml` | Kafka group (accounting, fraud-detection, kafka) |
| `compose.observability.yaml` | Jaeger, Prometheus, OpenSearch, Grafana |
| `compose.extras.yaml` | Empty stub for vendor/fork customizations |

The OTel Collector config is split into similar layers (`otelcol-config.yml`, `otelcol-config-full.yml`, `otelcol-config-observability.yml`, `otelcol-config-extras.yml`) and loaded via `--config` flags in the same order.

Four `make` targets cover the combinations:

```
make start                  # full + observability (default)
make start-minimal          # core + observability
make start-no-o11y          # full, no observability stack
make start-minimal-no-o11y  # core only
```

Forks can update the otherwise empty `compose.extras.yaml` and `otelcol-config-extras.yml` overrides to add custom o11y backends or demo services without touching upstream files.


julianocosta89:
Please, let's not have 4 different collector config files and 4 different compose files.
This will be a mess maintaining and configuring.

Why can't we use docker profiles?
We can keep it all in one single compose file and just configure which profile will be executed via `make`.

puckpuck:
The issue is when you have the same service that is run multiple different ways (collector, checkout, recommendation, etc) based on which profile it belongs to. To make this work with profiles you still need multiple -f files, so with profiles we are now mixing the -f CLI option and the profile construct.

This approach just doesn't use profiles at all and stays with -f, but in a cleaner way than we have today.

rogercoll:
> Please, let's not have 4 different collector config files and 4 different compose files. This will be a mess maintaining and configuring.
> 
> Why can't we use docker profiles? We can keep it all in one single compose file and just configure which profile will be executed via `make`.

I would also prefer not having 4 different compose files, but we might require them either way with Docker Profiles. Profiles can't patch services across boundaries, thus services like checkout needs KAFKA_ADDR and `depends_on: kafka` only in full mode, and otel-collector needs `depends_on: jaeger/opensearch` only with the observability mode. Profiles select which services to start, but they can't conditionally add env vars or dependencies to a service that's always on.

See that in https://github.com/open-telemetry/opentelemetry-demo/pull/3107 a new `compose.full.yaml` was needed for the `full` profile.

compose.yaml
```diff
@@ -0,0 +1,757 @@
...
+      - OTEL_COLLECTOR_PORT_HTTP
+      - OTEL_RESOURCE_ATTRIBUTES=${OTEL_RESOURCE_ATTRIBUTES},service.criticality=critical
+      - OTEL_SERVICE_NAME=frontend-proxy
+      - ENVOY_PORT
+      - ENVOY_ADDR
+      - ENVOY_ADMIN_PORT
+      - FLAGD_HOST
+      - FLAGD_PORT
+      - FLAGD_UI_HOST
+      - FLAGD_UI_PORT
```
osullivandonal, line 316: Looks like we need to add these env vars to the `front-end` proxy to get this to work:

```yaml
  - TELEMETRY_DOCS_HOST
  - TELEMETRY_DOCS_PORT
```

Without the above when I run any `make start-*` the front-end proxy restarts with this error:

```bash
: Proto constraint validation failed (BootstrapValidationError.StaticResources: embedded message failed validation | caused by StaticResourcesValidationError.Clusters[9]: embedded message failed validation | caused by ClusterValidationError.LoadAssignment: embedded message failed validation | caused by ClusterLoadAssignmentValidationError.Endpoints[0]: embedded message failed validation | caused by LocalityLbEndpointsValidationError.LbEndpoints[0]: embedded message failed validation | caused by LbEndpointValidationError.Endpoint: embedded message failed validation | caused by EndpointValidationError.Address: embedded message failed validation | caused by AddressValidationError.SocketAddress: embedded message failed validation | caused by SocketAddressValidationError.Address: value length must be at least 1 characters)
```

osullivandonal:
Even with the added maintenance burden of multiple compose files, I do like the idea of the `compose.extras.yaml` file to allow forks to easily specify there backends, o11y setup. Makes it a lot easier than the current setup.

Makefile
```diff
@@ -14,6 +14,28 @@ ADDLICENSE = $(TOOLS_DIR)/$(ADDLICENSE_BINARY)
...
+DOCKER_COMPOSE_FILES_OBSERVABILITY=-f compose.observability.yaml
+DOCKER_COMPOSE_FILES_EXTRAS=-f compose.extras.yaml
+
+# Default: full demo + observability stack + extras stub
+DOCKER_COMPOSE_FILES=$(DOCKER_COMPOSE_FILES_FULL) $(DOCKER_COMPOSE_FILES_OBSERVABILITY) $(DOCKER_COMPOSE_FILES_EXTRAS)
+
+# Java Workaround for macOS 15.2+ and M4 chips (see https://bugs.openjdk.org/browse/JDK-8345296)
+ifeq ($(shell uname -m),arm64)
+	ifeq ($(shell uname -s),Darwin)
+		DOCKER_COMPOSE_ENV+= --env-file .env.arm64
```
ps48: Can we please rebase the PR onto main. The Makefile references `--env-file .env.arm64` which was removed in #3237. On macOS arm64, make start fails with couldn't find env file: `.env.arm64`.

Otherwise, tested locally all four targets (start, start-minimal, start-no-o11y, start-minimal-no-o11y) work correctly after creating an empty `.env.arm64` workaround. All services come up healthy and telemetry flows through the collector.

puckpuck: I updated the Makefile to remove all references to this, and incorporate all other recent changes that were made to it on main.

compose.yaml
```diff
@@ -167,7 +185,8 @@ services:
     image: ${IMAGE_NAME}:${DEMO_VERSION}-email
     container_name: email
     build:
-      context: ./src/email
+      context: ./
```
ps48, line 188: Should change the build context here, is this change intentional?

compose.yaml
ps48, line 317: Should we also add the missing `telemetry-docs` service. PR #2794 added telemetry-docs to docker-compose.yml on main. The frontend-proxy still passes `TELEMETRY_DOCS_HOST/TELEMETRY_DOCS_PORT` env vars and has envoy routing for it, but the service definition and this `depends_on` entry are both missing from all four compose layers. 

Should it be added as a service in compose.yaml and listed here in `frontend-proxy.depends_on`?

Makefile
```diff
@@ -154,12 +170,12 @@ clean-images:
 
 .PHONY: run-tests
 run-tests:
-	$(DOCKER_COMPOSE_CMD) $(DOCKER_COMPOSE_ENV) -f docker-compose-tests.yml run frontendTests
-	$(DOCKER_COMPOSE_CMD) $(DOCKER_COMPOSE_ENV) -f docker-compose-tests.yml run traceBasedTests
+	$(DOCKER_COMPOSE_CMD) $(DOCKER_COMPOSE_ENV) $(DOCKER_COMPOSE_FILES) -f docker-compose-tests.yml run frontendTests
+	$(DOCKER_COMPOSE_CMD) $(DOCKER_COMPOSE_ENV) $(DOCKER_COMPOSE_FILES) -f docker-compose-tests.yml run traceBasedTests
```
ps48: `docker-compose-tests.yml` is deleted in this PR but still referenced here (and in run-tracetesting at line `176`, and stop at line `230`). These targets will fail. 

We should either recreate a test compose file that layers on the new files, or remove/update these targets.

compose.observability.yaml
```diff
@@ -0,0 +1,171 @@
...
+      - "${PROMETHEUS_PORT}:${PROMETHEUS_PORT}"
+    logging:
+      driver: "json-file"
+      options:
+        max-size: "5m"
+        max-file: "2"
+
+  # OpenSearch
+  opensearch:
+    container_name: opensearch
```
ps48, line 95: Missing `image: ${IMAGE_NAME}:${DEMO_VERSION}-opensearch`, it was present in the old `docker-compose.yml` but not carried over. Without it, docker compose pull can't fetch a pre-built image.

src/ad/build.gradle
```diff
@@ -21,7 +21,7 @@ version = "0.1.0-SNAPSHOT"
 
 def opentelemetryVersion = "1.61.0"
 def opentelemetryInstrumentationVersion = "2.27.0"
-def grpcVersion = "1.81.0"
+def grpcVersion = "1.79.0"
```
puckpuck: without this reverted change the ad service would not run properly. This feels like a regression from a Dependabot update.

Makefile
puckpuck: Existing tests are currently broken/unreliable. We should remove this target, and update tests through another PR, like this [one](https://github.com/open-telemetry/opentelemetry-demo/pull/3356).

julianocosta89:
I found an issue that I'm okay in living with it.

At Datadog I usually run the Collector with an extra feature flag enabled. To that to work in the current setup I need to add all `commands` to the `extra` file.
If I run `full` mode it dual-ship to Datadog and Demo Obs stack

BUT if I run `no-o11y`, the Collector crashes because of `--config=/etc/otelcol-config-observability.yml` is not part of the files.

So f I want users to dual-ship I need to recommend one thing, but if I want to showcase only Datadog, then I can remove the `--config=/etc/otelcol-config-observability.yml` from my extra file.

This is a bit annoying for local review (I dual ship all the time), but as I don't see an easier fix I'm okay in living with that

julianocosta89:
I also have to remove the components the collector pipelines I'm changing  🫠 

```
Error: invalid configuration: service::pipelines::logs: references exporter "opensearch" which is not configured
Error: invalid configuration: service::pipelines::metrics: references exporter "otlp_http/prometheus" which is not configured
```

src/otel-collector/otelcol-config-full.yml
```diff
@@ -0,0 +1,46 @@
...
+receivers:
+  kafkametrics:
+    scrapers:
+      - brokers
+      - topics
+      - consumers
+    brokers:
+      - ${KAFKA_ADDR}
+    collection_interval: 10s
+  postgresql:
```
julianocosta89: postgresql is always part of the Demo. This receiver needs to be moved to the main otelcol-config

src/otel-collector/otelcol-config-observability.yml
```diff
@@ -0,0 +1,51 @@
...
+    sending_queue:
+      # Explicitly set due to bug: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/45016
+      num_consumers: 10
+      queue_size: 1000
+      batch:
+
+service:
+  pipelines:
+    traces:
+      exporters: [debug, otlp_grpc/jaeger, spanmetrics]
```
ps48: References the connector as spanmetrics, but the connector defined in otelcol-config.yml:181 (and used everywhere else) is span_metrics. The collector enters a restart loop:
```
  Error: invalid configuration: service::pipelines::traces:  references exporter "spanmetrics" which is not configured
```



```suggestion
      exporters: [debug, otlp_grpc/jaeger, span_metrics]
```

src/otel-collector/otelcol-config.yml
ps48: The always-loaded base wires the logs pipeline to the opensearch exporter, but opensearch is only defined in `otelcol-config-observability.yml`. So any mode that doesn't include the observability layer fails immediately:

```
Error: invalid configuration: service::pipelines::logs: references exporter "opensearch" which is not configured
```
  
**Fix**: set the base service.pipelines.logs.exporters to [debug] and have otelcol-config-observability.yml add opensearch back via the override layer (same pattern it already uses for traces/metrics).

Makefile
ps48: These targets will fail with "no such file" the moment they run. They need to be updated to the new compose files.  `-f docker-compose.yml`

julianocosta89 (approved):

julianocosta89:
@puckpuck I went through everything.
I updated the `postgresql` env vars and added `/telemetry` to all start modes messages.

I think everything that was raised by @ps48 was addressed by you, so I believe we should be good to merge!

I'll give you the honours 😅

PR #3238, opened 2026-04-10
ci: restrict image push and FOSSA to open-telemetry repo

puckpuck:
## Summary

- Image builds in `component-build-images.yml` now run in all forks (useful for Dockerfile validation), but GHCR login, Docker Hub login, and the actual image push are gated on `github.repository == 'open-telemetry/opentelemetry-demo'` to avoid missing-secret failures in forks
- FOSSA scanning job is similarly guarded since it requires an org-level API key unavailable in forks

## Problem

Forks of this repo experience CI failures on every push to `main`:
- `Checks` / `build_and_push_images` — fails with `Username and password required` because `DOCKER_USERNAME`/`DOCKER_PASSWORD` org secrets are not available in forks
- `FOSSA scanning` — fails with `Input required and not supplied: api-key` for the same reason

## Test plan

- [ ] Verify `Checks` workflow passes on push to `main` in a fork (build runs, push is skipped)
- [ ] Verify `FOSSA scanning` job is skipped in forks
- [ ] Verify both workflows still function correctly in the upstream repo
- [x] Manually verified changes

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.github/workflows/component-build-images.yml
+++ b/.github/workflows/component-build-images.yml
@@ -184,13 +184,13 @@ jobs:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
-        if: ${{ inputs.push && github.repository == 'open-telemetry/opentelemetry-demo' }}
+        if: ${{ inputs.push && !github.event.repository.fork }}
       - name: Log in to Docker Hub
         uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121  # v4.1.0
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
-        if: ${{ inputs.push && github.repository == 'open-telemetry/opentelemetry-demo' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/component-build-images.yml
```diff
@@ -184,13 +184,13 @@ jobs:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
-        if: ${{ inputs.push }}
+        if: ${{ inputs.push && github.repository == 'open-telemetry/opentelemetry-demo' }}
```
Kielek: Consider `github.event.repository.fork == false`
ref: https://github.com/open-telemetry/opentelemetry-dotnet/blob/bfa5afc69832efb8253c73237b022696225501f1/.github/workflows/fossa.yml#L14

puckpuck: Oh, that's a much cleaner solution.

We use the github.repository == '...' syntax in multiple workflows already, so I will create another PR to bring all thoses ones to this standard as well.

Kielek (approved):

PR #3239, opened 2026-04-10
fix(telemetry-schema): enforce app.* prefix for all custom metric attributes

julianocosta89:
## Summary

Fixes found during review of #2794. Addresses all outstanding review comments and a few additional issues spotted in the schema and templates.

- **`product.id` → `app.product.id`**: removed the local attribute definition from `metrics/product_reviews.yaml` and updated both metric counter calls in `product_reviews_server.py` to use the canonical ref, consistent with how span attributes are set in the same file
- **`recommendation.type` → `app.recommendation.type`**: same pattern — removed local definition from `metrics/recommendation.yaml`, added canonical definition to `attributes/misc.yaml`, added ref to `services/recommendation.yaml`, and updated `recommendation_server.py`
- **Monetary type fixes** (`app.payment.amount`, `app.shipping.cost.total`): changed from `string` to `double` in the schema (addresses ps48's review comments); updated payment service (JS) and shipping service (Rust) to emit numeric values instead of formatted strings
- **Grafana dashboard**: updated PromQL label `recommendation_type` → `app_recommendation_type` in `demo-dashboard.json` to match the renamed attribute (Prometheus normalises `.` to `_`)
- **Template cleanup**: removed 5 unused macros from `attribute_macros.j2` (`find_lineage`, `name_with_link`, `name`, `display_name`, `humanize`, `sentence_case`)
- **Template banner paths**: fixed incorrect auto-generated comment paths in `service.md.j2` and `registry.md.j2`


julianocosta89:
@martinjt this PR is pointing to your PR.
LMK what you think.

I was pushing to your PR, but the force pushes wiped out the changes I sent.

puckpuck:
Should we be moving to use a `demo.` prefix instead of `app.` since app. is getting set for mobile apps?

I think we need this PR to become consistent, but we should probably follow with a much larger PR that renames the prefix for all telemetry including dashboard references.

puckpuck (approved):

julianocosta89:
> Should we be moving to use a `demo.` prefix instead of `app.` since app. is getting set for mobile apps?
> 
> 
> 
> I think we need this PR to become consistent, but we should probably follow with a much larger PR that renames the prefix for all telemetry including dashboard references.

Yes!
I think @martinjt brought that up some time ago.
`app.` is now reserved, and we should move away from it.

PR #3254, opened 2026-04-13
ci: replace linkspector with lychee for link checking

puckpuck:
## Summary

- Replaces `umbrelladocs/action-linkspector` with `lycheeverse/lychee-action@v2.8.0` in CI
- Removes `@umbrelladocs/linkspector` npm dev dependency
- Adds `.lychee.toml` config preserving equivalent settings (URL excludes, accepted status codes, gitignore support)
- Sets a browser-like `user_agent` to avoid false 403s from CDN bot filters (e.g. Splunk community)
- Updates `make checklinks` to call `lychee` directly (no npm install needed)

## Motivation

linkspector was producing flaky tests and false positives in CI. lychee is a more robust alternative with better error reporting and retry handling.

## Test plan

- [x] Ran `lychee --config .lychee.toml .` locally — 0 errors, 141 links checked


Makefile
```diff
@@ -92,8 +92,7 @@ addlicense:	$(ADDLICENSE)
 .PHONY: checklinks
 checklinks:
 	@echo "Checking links..."
-	@if ! npm ls @umbrelladocs/linkspector; then npm install; fi
-	linkspector check
+	lychee --config .lychee.toml .
```
Kielek, line 95: I think that it will be non work locally by default.
Historically, it was installing some npm packages when needed, now you need to install it manually probably.

Kielek (approved):
CI side LGTM
Some note about local experience.

Makefile
puckpuck: yeah for locally you will need to install lychee manually.

`brew install lychee` is what I used. You can also install it via cargo.

Kielek: Consider extending https://github.com/puckpuck/opentelemetry-demo/blob/a4a515bc68bc8467b86310de2c4c9288df2b591c/Makefile#L107

PR #3276, opened 2026-04-17
Fix charge span leak in the payment service

### LINKED ISSUE #3275: Payment service: charge span leaks and misattributes exceptions on validation errors
# Bug Report

Which version of the demo you are using? https://github.com/open-telemetry/opentelemetry-demo/commit/0982a0268aec2f010c812fcb87ca4e6ff81be643

## Symptom

The manually-created `charge` span in the payment service (`charge.js)` is not ended when some validation errors occur, which means the span is never exported. Exceptions are also recorded on the parent gRPC span.

Image

The `charge` span should always be ended, regardless of whether the charge succeeds or fails. Also, when an error occurs inside `charge()`, the exception and `ERROR` status should be recorded on the `charge` span where the failure originates. The parent span should reflect the failure via
  `setStatus()` only, without duplicating the exception.

Image

The `charge` span is called at the top of the function, but three validation error paths throw without ever calling `span.end()`.

## Reproduce

Attempt to place an order with expired card. Change the **Month** and **Year** fields under **Payment Method** to **January** and **2026** respectively.

Observe in Jaeger that the `charge` span is not present for the corresponding trace and that the auto-instrumented parent gRPC span incorrectly owns the exception.

ayoisaiah:
# Changes

Fix span lifecycle and exception attribution in the payment service charge span.

  - Wrap the body of `charge()` in a `try/catch/finally` so `span.end()` is always called, including on validation error paths where the span previously leaked.
  - Record the exception and set `SpanStatusCode.ERROR` on the charge span in the catch block, where the failure actually originates.
  - Remove `span.recordException(err)` from `index.js`. The gRPC span now only has its status set to `ERROR` with the error message, avoiding duplicate exception attribution.

Fixes #3275 

## Merge Requirements

essential items:


src/payment/charge.js
```diff
@@ -24,65 +24,72 @@ function random(arr) {
...
-  } else {
-    span.setAttribute('app.payment.charged', true);
-  }
+    // Check baggage for synthetic_request=true, and add charged attribute accordingly
+    const baggage = propagation.getBaggage(context.active());
+    if (baggage && baggage.getEntry('synthetic_request') && baggage.getEntry('synthetic_request').value === 'true') {
+      span.setAttribute('app.payment.charged', false);
+    } else {
+      span.setAttribute('app.payment.charged', true);
+    }
```
osullivandonal, line 80: Could we revert the formatting here? It seems all these lines have been tabbed over?

osullivandonal (commented):
Nice work on this, thanks for the contribution! Could we add a changelog entry to [CHANGELOG.md](https://github.com/open-telemetry/opentelemetry-demo/blob/0982a0268aec2f010c812fcb87ca4e6ff81be643/CHANGELOG.md?plain=1#L1).

src/payment/charge.js
ayoisaiah: That happened because everything was wrapped in a try/catch block, which altered the indentation

osullivandonal: Ah ok, go it.

puckpuck:
@ayoisaiah Can you fix the yamllint check errors? Don't worry about the failing link checks, that's a known issue in this repo.

ayoisaiah:
@puckpuck Just pushed the fix, thanks!

puckpuck (approved):
Everything works great. Thanks for this.

PR #2994, opened 2026-02-13
feat: add `main` tagged images, drop date suffix for `nightly`

svrnm:
# Changes

fixes https://github.com/open-telemetry/opentelemetry-demo/issues/2971

- Add `main` tagged images on push to `main` branch
- Drop suffix for `nightly` builds, so they always publish to `nightly` (BREAKING)

## Merge Requirements

For new features contributions, please make sure you have completed the following
essential items:

* [x] `CHANGELOG.md` updated to document new feature additions
* [ ] Appropriate documentation updates in the [docs][]
* [ ] Appropriate Helm chart updates in the [helm-charts][]

PR #3302, opened 2026-04-23
fix: implement rolling weekday tags for nightly images

AnjaliMishra1st:
## Summary
Adds rolling weekday-based tags (e.g., nightly-Monday) for nightly Docker images.

## Why
Avoids overwriting the same nightly tag and improves traceability of builds.

## Impact
No breaking changes, only affects CI/CD tagging.

### DIFF
--- a/.github/workflows/component-build-images.yml
+++ b/.github/workflows/component-build-images.yml
@@ -11,17 +11,17 @@ on:
         type: boolean
       version:
         description: The version used when tagging the image
-        default: 'dev'
+        default: "dev"
         required: false
         type: string
       dockerhub_repo:
         description: Docker Hub repository
-        default: 'otel/demo'
+        default: "otel/demo"
         required: false
         type: string
       ghcr_repo:
         description: GHCR repository
-        default: 'ghcr.io/open-telemetry/demo'
+        default: "ghcr.io/open-telemetry/demo"
         required: false
         type: string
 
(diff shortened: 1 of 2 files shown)


.github/workflows/component-build-images.yml
```diff
@@ -33,7 +33,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
```
Kielek, line 36: Please keep tags, not only hashes

AnjaliMishra1st:
Thanks for the suggestion! I’ve replaced all commit hashes with version tags as requested. The workflow is now cleaner and consistent. Happy to make further changes if needed.

rogercoll (commented):
Oh, I forgot that we merged this PR https://github.com/open-telemetry/opentelemetry-demo/pull/2994 a couple of months ago that overrides the nightly tag. If you check the current publish images you will see the time difference: https://github.com/open-telemetry/opentelemetry-demo/pkgs/container/demo/versions?filters%5Bversion_type%5D=tagged

I would say that the only missing bit would be removing the old GitHub images tagged with `nightly-${date}...`. @open-telemetry/demo-maintainers Do you know if we have the privileges to do that?

AnjaliMishra1st:
Thanks for the clarification!
I understand that older nightly-${date} images still exist. Since removing those might require special permissions, I’ll leave that to maintainers.

Please let me know if any changes are needed from my side 🙂

AnjaliMishra1st:
I've cleaned up the branch and squashed changes into a single commit for clarity.
Please let me know if anything else is needed 🙂

rogercoll:
@AnjaliMishra1st Could you clarify why we still need this PR? Nightly images are already being overridden, what would be the benefit of adding the "github runner number" as part of the nightly tag?

AnjaliMishra1st:
Thanks for pointing that out. My intention was to improve traceability by keeping a rolling 7-day set of nightly tags (for example, `nightly-Monday` to `nightly-Sunday`) instead of overriding a single `nightly` tag.

This would make it easier to compare recent builds and roll back if needed. If the current behavior from #2994 already covers this use case, I’m happy to close or rework the PR based on your suggestion.

PR #3305, opened 2026-04-23
refactor: add OpenFeature flag generation support with checkout and product-catalog services

erka:
# Changes

- update checkout and product-catalog services to use OpenFeature CLI flag generation
- move `productCatalogFailure` evaluation logic into the flagd configuration
- perform minor cleanup in open-feature/go-sdk


julianocosta89 (commented):
Hey @erka I have a couple of questions on this PR.

I've tested it locally and when navigating to <localhost:8080/feature> all flags are disabled, but when looking at my traces I could see so spans failing and the reason is that the flag is enabled 🤔 

Image

This change breaks the way we control flags of the Demo, is that expected?
Is that the recommended approach for FF?

erka:
> This change breaks the way we control flags of the Demo, is that expected? Is that the recommended approach for FF?

Hey @julianocosta89 

`http://localhost:8080/feature` is not that great for `productCatalogFailure` flag as it only shows the default value and hides the evaluation rules which that flag has. You have to open `Advanced` mode to see the real flag configuration for it.

Please check my changes to `src/flagd/demo.flagd.json` and `src/product-catalog/main.go` in this PR. I believe people will use feature flags in such a way. Using hard-coded conditions in the sources is a bad showcase for feature flags.

erka:
@julianocosta89 

I updated the flagd-ui to handle this special flag, but I’m seeing this language for the first time.

julianocosta89:
@erka
> `http://localhost:8080/feature` is not that great for productCatalogFailure flag as it only shows the default value and hides the evaluation rules which that flag has. You have to open Advanced mode to see the real flag configuration for it.

that is definitely not the expected behaviour. I'll take a second look on Monday and I'll let you know.

julianocosta89 (commented):
Thx @erka, this looks great!

Before getting this one merged, could you add a description of how to generate the `flags_gen` files on the READMEs of each service?

erka:
Hey @julianocosta89

I've updated the READMEs but it looks like there is an issue with runners and I'm facing the error `remote: Your account is suspended`. Please ping me if I need to force-push it.

Kielek:
@erka, please check https://www.githubstatus.com/

image

EDIT: globabl problem, not related to your changes.

erka:
@Kielek thanks for the heads-up!

julianocosta89:
@erka the markdown error is real, could you take a look?

erka:
> @erka the markdown error is real, could you take a look?

This is fixed now.  

Could you please review the labels for this PR? I think they’re a bit off after all my changes. @julianocosta89

julianocosta89 (approved):

PR #3306, opened 2026-04-24
chore: add retry configuration parameters to lychee

julianocosta89:
This PR tries to fix flakynes of lychee.

### DIFF
--- a/.lychee.toml
+++ b/.lychee.toml
@@ -7,7 +7,7 @@ user_agent = "Mozilla/5.0 (compatible; lychee link checker)"
 max_retries = 3
 retry_wait_time = 10
 timeout = 30
-max_concurrency = 16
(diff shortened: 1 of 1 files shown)


.lychee.toml
```diff
@@ -4,6 +4,11 @@
 # Need this to avoid getting "bot denied"
 user_agent = "Mozilla/5.0 (compatible; lychee link checker)"
 
+max_retries = 3
+retry_wait_time = 10
+timeout = 30
+max_concurrency = 16
```
Kielek: With some other tools, I was albe to reduce flakines with setting similar configuration to 1

It might be important, if we hit same domain more than one.
```suggestion
max_concurrency = 1
```

Kielek (approved):
LGTM with some minor adjustment

PR #3314, opened 2026-04-24
ci: cache lychee link-check results and use during checking

chalin:
In `.github/workflows/checks.yml`:

- Restores the lychee cache before link checks to reuse recent results across CI runs.
- Saves the cache after each run so future checks avoid revalidating unchanged links.
- Moves accepted status handling into cached lychee args and removes the static TOML override.

This is based on the go repo setups (they also use Lychee caching).

/cc @puckpuck @julianocosta89

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -62,11 +62,25 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
+      - name: Restore lychee cache
+        uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae  # v5.0.5
+        id: cache-restore
+        with:
+          path: .lycheecache
+          key: ${{ runner.os }}-lychee-${{ hashFiles('.lychee.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-lychee-
       - name: Run lychee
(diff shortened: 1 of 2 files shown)


.lychee.toml
```diff
@@ -19,5 +19,3 @@ exclude = [
 exclude_path = [
   "CHANGELOG\\.md",
 ]
-
-accept = ["200", "429"]
```
chalin, line 23: With caching in place, we generally _want_ to know about 429, which may hide potential 404s. The caching will help reduce the likelihood of getting 429 ( 🤞 😄 )

.github/workflows/checks.yml
```diff
@@ -62,11 +62,25 @@ jobs:
...
+        with:
+          path: .lycheecache
+          key: ${{ runner.os }}-lychee-${{ hashFiles('.lychee.toml') }}
+          restore-keys: |
+            ${{ runner.os }}-lychee-
       - name: Run lychee
         uses: lycheeverse/lychee-action@8646ba30535128ac92d33dfc9133794bfdd9b411  # v2.8.0
         with:
-          args: --config .lychee.toml .
+          args: --config .lychee.toml --cache --max-cache-age 3d --cache-exclude-status 300..=599 .
```
chalin, line 76: 3 day cache freshness configured here. We can do more (7d feels reasonable too).

puckpuck (approved):

.github/workflows/checks.yml
puckpuck: I'm good with 3 days here

julianocosta89 (approved):

PR #3388, opened 2026-05-20
feat(ad): expose Prometheus /metrics endpoint scraped by the OTel Collector

cyrille-leclerc:
## Summary

- Add the Prometheus Java client library (`io.prometheus:prometheus-metrics` 1.3.6) to the **ad** service and expose a `/metrics` endpoint on `AD_PROMETHEUS_PORT` (default `9465`).
- Register a custom counter `demo_ad_served_total{category}` incremented every time ads are returned (with the real category, or `"random"` for the fallback branch). The metric name follows the new `demo.ad.*` / `demo_ad_*` convention introduced in #3387.
- Wire a `prometheus/ad` scrape job in `src/otel-collector/otelcol-config.yml` and include it in the `metrics` pipeline, so the new metric flows into Prometheus via the OTel Collector.

This is intended as a small, focused demo of the **OTel Collector's `prometheus` receiver scraping a Prometheus-exposed `/metrics` endpoint**, alongside the existing OTLP-push instrumentation that already lives in the ad service.

## Test plan

- [ ] `make redeploy service=ad` and `make redeploy service=otel-collector`
- [ ] `docker exec ad wget -qO- localhost:9465/metrics | grep demo_ad_served_total` — counter present and increases under load
- [ ] In the OTel Collector logs, confirm the `prometheus/ad` receiver is scraping `ad:9465` without errors
- [ ] In Prometheus (`http://localhost:8080/prometheus`), query `demo_ad_served_total` and confirm samples are present and increasing
- [ ] `make check` passes


cyrille-leclerc:
This Prometheus integration would help inform:
- https://github.com/open-telemetry/opentelemetry-specification/pull/4956#discussion_r3274268118

Kielek:
TBH I do not like an idea of mixing Prometheus with OpenTelemetry instrumentation. Even if it is possible I would not document it as a best practice. 

Can it be done differently? Eg. Possibility to configure demo with Prometheus exporter/reader with some switch?
I am not very familiar with Java stuff, but this one should be used IMO https://central.sonatype.com/artifact/io.opentelemetry/opentelemetry-exporter-prometheus

alternatively it should be built in Java Agent/JAR

cyrille-leclerc:
I do agree, I don't recommend keeping Prometheus instrumentation when adopting OpenTelemetry but I think it's a common use case when adopting OTel to have existing custom metrics with usually the plan to migrate and the otel-demo can show how to integrate them.

We could add a comment on the metric to explain this rationale and recommend migration to otel custom metrics.

BTW I'm considering writing an opentelemetry.io blog post on this migration topic.

WDYT?

Kielek:
I think that user tend to copy code as is. If you provide here an example of coexisting technologies, you will eventually find it in the wild.

Separate post, based on your fork/branch  seems fine, but including it here for a long term, no.

I think that we could consider merging this one when you prepare just a follow up with migration to the OTel stuff.
In this way, you have your story, Prometheus metrics migrated to the OTel. The bad part is that you include Prometheus to the package already instrumented by OTel.

cyrille-leclerc:
Thanks @Kielek , please look at the comments I added and check if the recommendation is clear enough 🙏

Kielek:
Message is clear, can we remove this transitional pattern in follow up PR and fully migrate to the OTel? :)

puckpuck (approved):
I'm good with this with all the comments indicating this is a migration pattern. 

Can we also make sure to update [docs](https://opentelemetry.io/docs/demo/services/ad/) about this so all places that people check for guidance are clear on it?

puckpuck:
@cyrille-leclerc this PR comes before we move to the layered compose.*.yaml files. You will need to implement changes from the docker-compose files to the new compose.* files

cyrille-leclerc:
@puckpuck conflict fixed + opentelemetry.io documentation PR created: https://github.com/open-telemetry/opentelemetry.io/pull/10033

puckpuck (approved):

Kielek (approved):

PR #3394, opened 2026-05-22
fix(ad): switch from grpc-netty to grpc-netty-shaded to fix startup crash

puckpuck:
## Summary

when we upgraded to gRPC 1.81.0 in #3346 the ad service experienced issues. This fixes that by using the `-shaded` plugin dependency to isolate the common dependency

## Root cause

```
Exception in thread "main" java.lang.AbstractMethodError: Receiver class
io.grpc.netty.shaded.io.grpc.netty.NettyServerBuilder$NettyClientTransportServersBuilder
does not define or inherit an implementation of the resolved method
'abstract io.grpc.internal.InternalServer buildClientTransportServers(java.util.List, io.grpc.MetricRecorder)'
```


julianocosta89:
I'll close this one as duplicate of your own work 😅 

https://github.com/open-telemetry/opentelemetry-demo/pull/3229/changes#diff-5db2129ea9c6b0166fcb137e85a2d795dd2aa3e5146193a5d8cb4b2343492171

Let's get #3229 merged

PR #3403, opened 2026-05-23
feat(make): add optional single-service support to build target

puckpuck:
Add an ability to use `make build service=<service_name>` to build an individual service. With the move to layered compose files, this will also ensure the full compose set is loaded when doing a build.

The pattern to specify the service is similar to what we do with the `restart` and `redeploy` targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,9 +128,20 @@ install-tools: $(MISSPELL) $(ADDLICENSE)
 	npm install
 	@echo "All tools installed"
 
+# Use to build all services, or a single service component
+# Example: make build service=frontend
 .PHONY: build
 build:
+# work with `service` or `SERVICE` as input
+ifdef SERVICE
+	service := $(SERVICE)
+endif
+
+ifdef service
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -128,9 +128,20 @@ install-tools: $(MISSPELL) $(ADDLICENSE)
...
+# Example: make build service=frontend
 .PHONY: build
 build:
+# work with `service` or `SERVICE` as input
+ifdef SERVICE
+	service := $(SERVICE)
+endif
+
+ifdef service
+	$(DOCKER_COMPOSE_CMD) $(DOCKER_COMPOSE_ENV) $(DOCKER_COMPOSE_FILES) build $(service)
```
ps48, line 141: The new `service= arg` works. Only the `SERVICE=` alias is broken and it's broken everywhere else in the file already for `restart` and `redeploy`. We can get this fix in now and do a follow up on the next PRs for ALL CAPS `SERVICE` fix.

ps48 (approved):

Issue #3267, opened 2026-04-16
Rename app.* attributes and metrics to demo.* with domain semantics


martinjt:
> **Note:** This proposal was developed using Claude Code for analysis of the current schema, attribute usage, and metric instrumentation across all services. All findings were manually verified.

## Summary

Follow-up from #3248 — `app.*` is now a reserved attribute namespace for client-side instrumentation. All custom attributes and metrics in the demo need to move to a `demo.*` prefix.

Beyond the prefix rename, this is an opportunity to clean up naming to use **domain semantics** (what the data represents) rather than architecture semantics (which service produces it), and to adopt semconv attributes where they already exist.

## Proposed Domain Taxonomy

| Prefix | Covers |
|---|---|
| `demo.product.*` | Catalog, search, reviews, filtering |
| `demo.ad.*` | Advertising, targeting |
| `demo.order.*` | Orders |
| `demo.cart.*` | Shopping cart |
| `demo.payment.*` | Transactions, card details |
| `demo.shipping.*` | Shipping, quotes, tracking |
| `demo.exchange.*` | Currency conversion operations |
| `demo.recommendation.*` | Recommendations, cache |
| `demo.notification.*` | Email confirmations |
| `demo.user_context.*` | User context (selected currency, loyalty) |
| `demo.feature_flag.*` | Feature flag state |
| `demo.synthetic_request` | Test traffic flag |

`user.id` should use the existing semconv attribute and be removed from the demo schema entirely.

### Product (`demo.product.*`)

| Current | Proposed |
|---|---|
| `app.product.id` | `demo.product.id` |
| `app.product.name` | `demo.product.name` |
| `app.product.quantity` | `demo.product.quantity` |
| `app.product.question` | `demo.product.review.question` |
| `app.products.count` | `demo.product.count` |
| `app.products_recommended.count` | `demo.product.recommended.count` |
| `app.filtered_products.count` | `demo.product.filtered.count` |
| `app.filtered_products.list` | `demo.product.filtered.list` |
| `app.products_search.count` | `demo.product.search.count` |
| `app.product_reviews.count` | `demo.product.review.count` |
| `app.product_reviews.average_score` | `demo.product.review.average_score` |

### Advertising (`demo.ad.*`)

| Current | Proposed |
|---|---|
| `app.ads.category` | `demo.ad.category` |
| `app.ads.count` | `demo.ad.count` |
| `app.ads.contextKeys` | `demo.ad.context_keys` |
[...]

PR #3435, opened 2026-05-28
fix: clean up some remaining demo telemetry naming leftovers

fbourgey:
Towards https://github.com/open-telemetry/opentelemetry-demo/issues/3267

# Changes

update Ad service documentation for `demo.ad.requests`, rename the shipping
  items metric to `demo.shipping.items_shipped`, remove duplicate local
  currency metric attributes.

## Merge Requirements

essential items:


src/currency/src/server.cpp
```diff
@@ -205,7 +205,7 @@ class CurrencyService final : public oteldemo::CurrencyService::Service
       span->SetAttribute("demo.exchange.from", from_code);
       span->SetAttribute("demo.exchange.to", to_code);
 
-      CurrencyCounter(to_code);
+      CurrencyCounter();
```
julianocosta89: If the currency code is not passed here, we do not need this counter anymore.

The whole currencyCounter() logic is to add the currency code as an attribute and have a counter for it.
I believe we have migrated to exchange `from` and `to`, so we are safe to remove that and the whole function on lines 231 to 237.

julianocosta89:
Sorry @fbourgey I think I'm misguiding you here 😞 

I like that this is aligned with #3267, but I don't like the idea of removing the custom metric from currency service.
Ideally we showcase how to create custom metrics in all different programming languages, and if we drop it here, we will miss C++ on the list.

Maybe we need to take a step back and define the currency metric we want to have here, and then implement it.

WDYT?!

fbourgey:
> Sorry @fbourgey I think I'm misguiding you here 😞
> 
> I like that this is aligned with #3267, but I don't like the idea of removing the custom metric from currency service. Ideally we showcase how to create custom metrics in all different programming languages, and if we drop it here, we will miss C++ on the list.
> 
> Maybe we need to take a step back and define the currency metric we want to have here, and then implement it.
> 
> WDYT?!

@julianocosta89 I've tried to follow your suggestion. Let me know what you think.

julianocosta89:
@fbourgey not really, we should keep the counting just on `to_code`.
With that we can have a metric like how many users are using currency XYZ.

If we add `from` and `to` we will have a high number of `USD`, as all currencies are converted from it.

julianocosta89 (approved):

julianocosta89:
Thanks for your patience on this one @fbourgey!

PR #3486, opened 2026-06-06
feat(payment): use zero-code JS instrumentation via NODE_OPTIONS

### LINKED ISSUE #1605: Use zero-code JS instrumentation for Payment service.
# Feature Request

Currently the payment service has an `opentelemetry.js` code file which is required as part of the node app startup. This file is the standard boilerplate code except for the config on a single auto-instrumentation library. 

All of that same functionality can be achieved by leveraging environment variables and the zero-code approach, which requires a file part of the `opentelemetry/auto-instrumentations-node` package. We can even showcase how to do this using environment variables only with `NODE_OPTIONS`. Here are the docs on how to get zero-code instrumentation for Node https://opentelemetry.io/docs/zero-code/js/#configuring-the-module

tlee768:
## Summary

- Replaces the manual `opentelemetry.js` SDK initialization file with zero-code instrumentation using `NODE_OPTIONS=--require @opentelemetry/auto-instrumentations-node/register`
- Removes 11 direct OTel SDK dependencies from `package.json` that are no longer needed (available as transitive deps of `auto-instrumentations-node`)
- Adds `OTEL_EXPORTER_OTLP_PROTOCOL=grpc` to compose.yaml since the default endpoint uses port 4317 (gRPC) but the SDK defaults to `http/protobuf`

Closes open-telemetry/opentelemetry-demo#1605

## Behavioral notes

- **FS instrumentation**: The old code explicitly enabled `@opentelemetry/instrumentation-fs` with `requireParentSpan: true`. The `auto-instrumentations-node` register entry point disables FS instrumentation by default. This reduces trace noise with no impact on payment service observability.
- **Instrumentation injection**: `NODE_OPTIONS` is set in `compose.yaml` (not baked into the Dockerfile) following the zero-code pattern — instrumentation is configured by the deployment environment, allowing operators to enable/disable it without rebuilding the image. The `package.json` start script includes the `--require` flag for local development convenience. Helm chart will need a corresponding update (labeled `helm-update-required`).

## Test plan

- [x] Docker image builds successfully
- [x] Payment service starts and connects to collector
- [x] Traces appear in Jaeger with custom span attributes (`app.payment.card_type`, `app.loyalty.level`, etc.)
- [x] Metrics (`app.payment.transactions` counter) exported
- [x] Logs forwarded via pino-opentelemetry-transport
- [x] Run full checkout flow end-to-end

Assisted-by: This contribution was prepared with the assistance of an AI development tool.


.gitignore
```diff
@@ -59,3 +59,4 @@ src/shipping/target/
 test/tracetesting/tracetesting-vars.yaml
 
 !src/currency/build
+CLAUDE.local.md
```
ps48: Nit: We should remove this. For any AI generated meta-data docs we already have `.claude` directory ignored in root-folder `.gitignore` https://github.com/open-telemetry/opentelemetry-demo/blob/4baa77b5e0bcb911ed6af53e9a61fc7adadc6140/.gitignore#L17-L21

ps48 (commented):
Tested this setup locally, the `Js` auto-instrumentation is a good addition to the Demo. Thanks @tlee768 for this contribution. Please revert the `.gitignore` change we should be good to merge this right after.

.gitignore
tlee768: Done

Kielek:
Shouldn't it update also opentelemetry.io?
I think that it is swith manual->auto, there should be a checkbox for the PR draft message.

cyrille-leclerc:
Thanks for this great improvement.
As @Kielek rightfully said, can you please create a PR in parallel to update the docs page?

* https://opentelemetry.io/docs/demo/services/payment/
* https://github.com/open-telemetry/opentelemetry.io/blob/main/content/en/docs/demo/services/payment.md

tlee768:
> Thanks for this great improvement. As @Kielek rightfully said, can you please create a PR in parallel to update the docs page?
> 
> * https://opentelemetry.io/docs/demo/services/payment/
> * https://github.com/open-telemetry/opentelemetry.io/blob/main/content/en/docs/demo/services/payment.md

@cyrille-leclerc @ps48 @Kielek 
Opened https://github.com/open-telemetry/opentelemetry.io/pull/10511

ps48 (approved):

julianocosta89 (approved):

PR #3496, opened 2026-06-09
feat(demo-data): add geo.location.lat/lon to web events

duart38:
Adds the OTel geo semconv `geo.location.lat` / `geo.location.lon` attributes to frontend web events so demo traffic can be plotted on a map. Coordinates correlate to the seeded user's city — jittered ~10km via `faker.location.nearbyGPSCoordinate` so users from one city spread out instead of stacking on a single point — and fall back to a random global point for non-seeded (real human) visitors. Wired end to end: city-center lat/lon added to all 9 seed users in `people.json`, threaded through `window.seed.location` by the loadgenerator, typed in `_app.tsx`, and emitted in `createRandomLocation.ts`.


duart38:
Closing — opened against the wrong repository by mistake. (Automated note via Claude Code.)

PR #3487, opened 2026-06-08
[chore] Add health check to services

julianocosta89:
# Changes

This PR adds `healthcheck` blocks to: 
- `checkout`
- `product-catalog`
- `shipping`
- `frontend-proxy`
- `load-generator`
- `flagd-ui`
- `image-provider`
- `shipping`

Some services required specific changes:

- On `src/image-provider/nginx.conf.template` added location `/health` returning 200
- On `src/checkout/Dockerfile` added `grpc_health_probe`
- On `src/product-catalog/Dockerfile` added `grpc_health_probe`
- On `src/shipping`
  - Added `healthcheck` binary to Dockerfile alongside the main binary
  - `Cargo.toml` added `[[bin]]` `healthcheck` entry
  - `src/healthcheck.rs` new binary: TCP connect to `SHIPPING_PORT`, exits `0/1`
  - `src/main.rs` added `GET /health` route (required for the probe target)

## Merge Requirements

For new features contributions, please make sure you have completed the following
essential items:

* [x] `CHANGELOG.md` updated to document new feature additions
* [ ] Appropriate documentation updates in the [docs][]
* [ ] Appropriate Helm chart updates in the [helm-charts][]

Maintainers will not merge until the above have been completed. If you're unsure
which docs need to be changed ping the
[@open-telemetry/demo-approvers](https://github.com/orgs/open-telemetry/teams/demo-approvers).

[docs]: https://opentelemetry.io/docs/demo/
[helm-charts]: https://github.com/open-telemetry/opentelemetry-helm-charts

PR #3356, opened 2026-05-10
[testing] Add telemetry sanity tests to validate end-to-end observability pipeline

ps48:
##  Changes

Adds a Dockerized pytest-based telemetry sanity test suite that validates all instrumented services are sending traces, metrics, and logs to the correct observability backends (Jaeger, Prometheus, OpenSearch). This replaces the defunct Tracetest-based integration tests with a lightweight, holistic approach covering all three telemetry pillars.

  - **Test suite** (`test/telemetry/`): Parametrized pytest tests that query backend APIs to verify each service's telemetry is flowing. Uses PPL for OpenSearch, PromQL for Prometheus, and Jaeger REST API for traces.
  - **Service matrix** (`services.py`): Single source of truth for which services emit which signals. Adding a new service = one line change.
  - **Two scopes**: `make run-telemetry-tests` (full, 54 tests) and `make run-telemetry-tests-minimal` (minimal compose, 40 tests)
  - **CI workflow**: Runs on PRs touching `src/`, `test/telemetry/`, or compose files
  - **Envoy DNS fix**: Switches frontend-proxy to `getaddrinfo` resolver to prevent c-ares DNS resolution failures in environments with custom DNS search domains
  - **Frontend healthcheck**: Adds healthcheck to frontend service so frontend-proxy only starts after Next.js is serving, preventing 503s on cold start

## Merge Requirements

For new features contributions, please make sure you have completed the following
essential items:

* [x] `CHANGELOG.md` updated to document new feature additions
* [ ] Appropriate documentation updates in the [docs][]
* [ ] Appropriate Helm chart updates in the [helm-charts][]

Maintainers will not merge until the above have been completed. If you're unsure
which docs need to be changed ping the
[@open-telemetry/demo-approvers](https://github.com/orgs/open-telemetry/teams/demo-approvers).

[docs]: https://opentelemetry.io/docs/demo/
[helm-charts]: https://github.com/open-telemetry/opentelemetry-helm-charts

PR #3498, opened 2026-06-10
test(telemetry): build PR images, harden warmup, cap test step at 15m

ps48:
# Changes

Hardens the Telemetry Tests CI workflow to fix three issues seen since #3356 merged:

## 1. Tests now exercise the PR's own images

Previously the workflow ran `make run-telemetry-tests` → `docker compose up`, which pulls the released `ghcr.io/open-telemetry/demo:latest-*` images from the registry. A PR's own service changes were never actually built or tested (raised by @julianocosta89 on #3487).

Now a dedicated `build-images` job runs `make build` on the PR head, `docker save | zstd`s the demo images, and uploads them as a run-scoped artifact. Both `telemetry-tests-full` and `telemetry-tests-minimal` `needs: build-images`, download the artifact, and `docker load` before running — so they test the PR's code. Third-party images (Jaeger, Prometheus, Grafana, ...) are still pulled from their own registries, so the artifact stays limited to the images we build (~3 GB zstd).

## 2. Per-test timeouts from a backend warm-up race

On a cold start the three backends become queryable at different times. Observed locally: Jaeger had 12 services at ~11s while Prometheus still had only 3 (it caught up to ~20 by ~39s). The warm-up gate only checked Jaeger, so the metrics tests started against a near-empty Prometheus and burned one 180s per-test timeout per service — that's how a run ballooned to ~45 minutes.

The warm-up fixture now waits for Jaeger, Prometheus, **and** OpenSearch to all report telemetry before the per-test polls begin, so the data is already there when tests run.

## 3. Runaway runs

The test step is now capped at `timeout-minutes: 15` (job timeout lowered to 30m), so a hung run can no longer consume 45 minutes. `WARMUP_SECONDS`/`POLL_TIMEOUT` are passed from the workflow with extra headroom for slow shared runners, and `POLL_TIMEOUT` is now env-overridable.

## Verification

Verified locally from a cold start, both scopes:

- `make run-telemetry-tests` → **67 passed**, demo torn down, no timeouts
- `make run-telemetry-tests-minimal` → **54 passed**, demo torn down, no timeouts
- End-to-end signals confirmed flowing: traces (18 services in Jaeger incl. multi-service edges), metrics (21 services in Prometheus), logs (~8k records in OpenSearch), load generator producing traffic
- `make build` from source produces fresh `demo:latest-*` images; `docker save | gzip | docker load` round-trip confirmed

## Merge Requirements

* [x] `CHANGELOG.md` updated
* [ ] Appropriate documentation updates in the docs

### DIFF
--- a/.github/workflows/run-telemetry-tests.yml
+++ b/.github/workflows/run-telemetry-tests.yml
@@ -37,7 +37,7 @@ permissions:
   contents: read
 
 # Coalesce concurrent runs on the same PR or branch so an approval right
-# after a push doesn't fire two 45-minute builds in parallel.
+# after a push doesn't fire two builds in parallel.
 concurrency:
   group: telemetry-tests-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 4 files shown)


Kielek:
@ps48 I know that includign tests directly in the build pipeline was the long term plan, but if you want to rebuild it anyway, we could do it now?

With helthckecks merge it should be pretty stable.

julianocosta89:
We already have a `Build images` workflow, I wonder if we could use it instead of having a full build on every PR approved.
That will increase the execution time significantly, no?!

julianocosta89:
I don't think that's actually an option given the multi services and PRs that can happen at the same time.
Let's move on with that and see where we can improve from there

julianocosta89 (approved):

PR #3520, opened 2026-06-18
fix(recommendation): use SDK-detected resource for LoggerProvider

mwimpelberg28:
## Summary

The `LoggerProvider` in the recommendation service is constructed with a manually built `Resource` containing only `service.name`:

```python
logger_provider = LoggerProvider(
    resource=Resource.create(
        {
            'service.name': service_name,
        }
    ),
)
```

The trace and metric providers use the SDK-configured global providers, which build their resource via `Resource.create({})` and therefore pick up all attributes from `OTEL_SERVICE_NAME`, `OTEL_RESOURCE_ATTRIBUTES`, and the SDK's built-in detectors (`telemetry.sdk.language`, `telemetry.sdk.name`, `telemetry.sdk.version`).

Because the log provider's resource only contains `service.name`, log records arrive at the backend with fewer resource attributes than spans and metrics. This causes resource-based log-to-trace correlation to fail in backends like Grafana/Tempo, which match signals by comparing resource attributes.

## Fix

Replace the manual `Resource` dict with `Resource.create({})`, which delegates to the same SDK detection path used by the trace and metric providers:

```python
logger_provider = LoggerProvider(
    resource=Resource.create({}),
)
```

## Test plan

- [ ] Restart the recommendation service and confirm log records in the observability backend carry `telemetry.sdk.language`, `telemetry.sdk.name`, `telemetry.sdk.version` alongside `service.name`
- [ ] Confirm log-to-trace correlation works in Grafana (log records link to spans with matching resource)


src/recommendation/recommendation_server.py
```diff
@@ -138,11 +138,7 @@ def check_feature_flag(flag_name: str):
 
     # Initialize Logs
     logger_provider = LoggerProvider(
-        resource=Resource.create(
-            {
-                'service.name': service_name,
-            }
-        ),
+        resource=Resource.create({}),
```
cyrille-leclerc:
```suggestion
        resource = Resource.create({}),
```

cyrille-leclerc (approved):

Kielek (approved):

PR #3521, opened 2026-06-18
fix(frontend): show quantity-adjusted price in cart

### LINKED ISSUE #1526: [bug] Product quantity not being considered during checkout
# Bug Report

742594fc2339653d4b38aa67ddf9a017561c919a

## Symptom

In the web UI, when selecting more than 1 product, the total value is not updated, and also not considered in the final price of the shopping cart.

image

When you buy more than 1 product you are expected to be charged per unity 😅 

## Reproduce

Run the demo, navigate to a product page, chose more than 1 product and add to the cart.
Neither the product page nor the shopping cart will take in consideration the amount of products.

mwimpelberg28:
On the cart page, each line item now shows the unit price and a line total (unit price × quantity). Users can change the quantity per item directly in the cart via a dropdown selector, and the line total updates immediately. The order total already reflected quantity correctly; this change makes the per-item breakdown visible and editable.

The product detail page continues to show the unit price unchanged.

Fixes #1526

## Before

Before

## After

After


Kielek:
Could you please add CHANGELOG entry?
Screens before/after will be also great.

Kielek:
> Could you please add CHANGELOG entry? Screens before/after will be also great.

I think that I was not clear enough... at least for Claude Sonnet.
* please add note to the CHANGELOG.md
* screens should be included only in the PR description, it should not be committed.

mwimpelberg28:
> > Could you please add CHANGELOG entry? Screens before/after will be also great.
> 
> 
> 
> I think that I was not clear enough... at least for Claude Sonnet.
> 
> * please add note to the CHANGELOG.md
> 
> * screens should be included only in the PR description, it should not be committed.

Sorry about that!  Would it be better to start a new branch and new PR then?

Kielek:
It is up to you. I would just rewrite git history and push new changes here.

julianocosta89 (approved):

Issue #2431, opened 2025-08-07
CartIcon.svg initial ERR_CONNECTION_REFUSED


rcastley:
# Bug Report

Which version of the demo you are using? Kubernetes Manifest 2.0.2.

## Symptom

When the site loads, it attempts to initially fetch `CartIcon.svg` from `localhost:8080` which generates an error in the console. However, a few objects later the icon is fetched successfully using the correct domain name.

<img width="970" height="181" alt="Image" src="https://github.com/user-attachments/assets/a2b2f7de-d883-4cd3-be01-33791176beb9" />

This also generates errors in the Locust Load Generator

**What is the expected behavior?**

No error in the console or the Locust Load Generator

What do you expect to see?

No errors

**What is the actual behavior?**

Attempting to fetch icon from - `Request URL
http://localhost:8080//icons/CartIcon.svg?w=48&q=75`

Please describe the actual behavior experienced.

## Reproduce

Could you provide the minimum required steps to resolve the issue you're seeing?

Just load the homepage in the browser. 

We will close this issue if:

* The steps you provided are complex.
* If we can not reproduce the behavior you're reporting.

## Additional Context

Please feel free to add any other context about the problem here.

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

bhuvan-somisetty:
Hi, I'd like to fix this. I'll trace where the CartIcon.svg URL is being constructed in the frontend code and fix the incorrect localhost:8080 fallback. Will submit a PR soon.

PR #3582, opened 2026-06-27
fix(frontend): fix CartIcon ERR_CONNECTION_REFUSED on initial page load

bhuvan-somisetty:
# Changes

Fixes the `CartIcon.svg` `ERR_CONNECTION_REFUSED` error on initial page load, as reported in [#2431](https://github.com/open-telemetry/opentelemetry-demo/issues/2431).

## Root Cause

The custom Next.js image loader (`utils/imageLoader.js`) used module-level variables that defaulted to `localhost:8080`. During SSR `window` is undefined, so these defaults were embedded into the initial server-rendered HTML:

```
http://localhost:8080//icons/CartIcon.svg?w=48&q=75
```

The browser fetched these before hydration could correct them, producing `ERR_CONNECTION_REFUSED`. There was also a double-slash because `src` already starts with `/` and the template literal prepended another `/`.

## Fix

Replace the loader with a pure function that always returns a root-relative URL, identical during SSR and client rendering:

```js
export default function imageLoader({ src, width, quality }) {
  const normalizedSrc = `/${src.replace(/^\/+/, '')}`;
  return `${normalizedSrc}?w=${width}&q=${quality || 75}`;
}
```

- No `localhost` fallback — eliminates `ERR_CONNECTION_REFUSED` on initial load
- SSR and browser produce identical output — no React hydration mismatch
- Strips one or more leading slashes before restoring exactly one — fixes the double-slash
- The frontend-proxy already routes `/icons/` and `/images/` from the same origin, so an absolute URL assembled from `window.location` adds no value

## Merge Requirements

* [x] `CHANGELOG.md` updated
* [ ] Appropriate documentation updates in the [docs][] — _not required; bug fix with no user-visible behaviour change_
* [ ] Appropriate Helm chart updates in the [helm-charts][] — _not required; no compose/otelcol/Grafana files changed_


bhuvan-somisetty:
Hi @julianocosta89 and @Kielek — could you take a look at this when you get a chance? Happy to make any changes needed.

Kielek:
I asked Codex to verify changes, as I am not js expert...

Long feedback, I hope you will find it valuable. It will be great if you can manually double check the proposed fix and ensure that it is working.

---------------

## Finding

### [P2] Make the image loader deterministic across SSR and hydration

Location: [`src/frontend/utils/imageLoader.js`, lines 13-24](https://github.com/open-telemetry/opentelemetry-demo/blob/ecbc6f0d778b118b370a6987ff22247aff91c930/src/frontend/utils/imageLoader.js#L13-L24)

The proposed loader has two output paths:

```js
if (typeof window === "undefined") {
  return `/${normalizedSrc}?w=${width}&q=${quality || 75}`;
}

// ...read window.location...
return `${protocol}://${hostname}:${port}/${normalizedSrc}?w=${width}&q=${quality || 75}`;
```

For the cart icon, an exact smoke test against the PR head produced:

| Render environment | Loader result |
| --- | --- |
| SSR, where `window` is undefined | `/icons/CartIcon.svg?w=48&q=75` |
| Browser at `https://shop.example` | `https://shop.example:443/icons/CartIcon.svg?w=48&q=75` |

Both URLs identify the same resource, but they are not the same attribute value.
That distinction matters during hydration.

Next.js uses the custom loader from its shared image-property code to generate
both `srcSet` entries and the final `src`. The relevant code is in Next.js
[`generateImgAttrs`](https://github.com/vercel/next.js/blob/v16.2.9/packages/next/src/shared/lib/get-img-props.ts#L224-L282),
which is explicitly shared by server and client rendering. The loader will thus
be rerun in the browser for the same `Image` component and input props.

React requires the initial client render to produce the same output as the
server render. Its [`hydrateRoot` documentation](https://react.dev/reference/react-dom/client/hydrateRoot)
specifically identifies `typeof window !== 'undefined'` branches in rendering
logic as a common cause of hydration errors. React warns about these mismatches
in development, and it does not guarantee that attribute differences will be
patched.

#### Impact

- The failed `localhost:8080` request is fixed, but a hydration warning can take
  its place in development.
- The server and client calculate different `src` and every generated `srcSet`
  candidate, not just one value.
- React may retain the server-rendered attributes rather than applying the
  client values. The relative server values happen to be correct today, but the
  component is relying on undocumented mismatch recovery.
- The loader remains dependent on execution environment, so future changes to
  routing or image behavior can turn this into a visible correctness problem.

This is P2 rather than P1 because the server-rendered relative URL resolves to
the intended same-origin resource and should eliminate the reported failed
network request in the current deployment. The problem is nevertheless
blocking for approval because a deterministic one-line URL construction solves
both the original bug and the hydration issue without a meaningful tradeoff.

## Why a relative URL is sufficient

Every current `next/image` usage passes a same-origin path. The relevant paths
are under `/icons/` or `/images/`; no current call site requires a remote image
origin.

The frontend proxy already routes:

- `/images/` to the `image-provider` cluster; and
- the `/` catch-all, including `/icons/`, to the frontend cluster.

An absolute URL assembled from `window.location` therefore adds no routing
information. A leading-slash relative URL reaches exactly the same origin and
proxy routes, works behind non-localhost ingress hostnames, avoids hardcoded
ports, and is identical during SSR and hydration.

## Potential fix

The smallest robust implementation is a pure function of the loader arguments:

```js
// Copyright The OpenTelemetry Authors
// SPDX-License-Identifier: Apache-2.0

/*
 * Images are served through the same frontend-proxy origin as the application.
 * Return the same relative URL during SSR and client rendering so the generated
 * image attributes remain stable during hydration.
 */
export default function imageLoader({ src, width, quality }) {
  const normalizedSrc = `/${src.replace(/^\/+/, '')}`;
  return `${normalizedSrc}?w=${width}&q=${quality || 75}`;
}
```

This version:

- strips one or more leading slashes and restores exactly one;
- never emits `localhost`;
- never emits a protocol, hostname, or port;
- returns identical values with and without `window`;
- preserves the current `w` and `q` query contract;
- does not introduce a new dependency or configuration option.

### Candidate patch

The following is a proposal for the author to adapt. It has not been applied to
the working tree or PR branch.

```diff
diff --git a/src/frontend/utils/imageLoader.js b/src/frontend/utils/imageLoader.js
index 804e163a..00000000 100644
--- a/src/frontend/utils/imageLoader.js
+++ b/src/frontend/utils/imageLoader.js
@@ -1,25 +1,13 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
+
 /*
-  * We connect to image-provider through the envoy proxy, straight from the browser, for this we need to know the current hostname and port.
-  * During server-side rendering, window is undefined, so we return a relative URL to avoid
-  * hardcoded localhost references being embedded in the initial HTML.
-  */
+ * Images are served through the same frontend-proxy origin as the application.
+ * Return the same relative URL during SSR and client rendering so the generated
+ * image attributes remain stable during hydration.
+ */
 
 export default function imageLoader({ src, width, quality }) {
-  // Normalize src to avoid double slashes when src already starts with '/'
-  const normalizedSrc = src.startsWith('/') ? src.slice(1) : src;
-
-  if (typeof window === "undefined") {
-    // SSR: return a relative URL so no localhost reference ends up in the initial HTML
-    return `/${normalizedSrc}?w=${width}&q=${quality || 75}`;
-  }
-
-  const { hostname, protocol: rawProtocol, port: rawPort } = window.location;
-  const protocol = rawProtocol.slice(0, -1); // Remove trailing ':'
-  const port = rawPort ? parseInt(rawPort, 10) : (rawProtocol === "https:" ? 443 : 80);
-
-  // We pass down the optimisation request to the image-provider service here, without this,
-  // Next.js would try to use its internal optimiser which does not work with the external image-provider.
-  return `${protocol}://${hostname}:${port}/${normalizedSrc}?w=${width}&q=${quality || 75}`;
+  const normalizedSrc = `/${src.replace(/^\/+/, '')}`;
+  return `${normalizedSrc}?w=${width}&q=${quality || 75}`;
 }
```

### Suggested changelog shape

The PR's changelog entry is a single long line and does not follow the wrapping
and link style used by nearby entries. After rebasing, a more consistent entry
would be:

```markdown
* [frontend] Avoid hardcoded `localhost:8080` image URLs during SSR and
  normalize leading slashes in the custom image loader
  ([#3582](https://github.com/open-telemetry/opentelemetry-demo/pull/3582))
```

This is a non-functional cleanup, but it should be handled while resolving the
existing changelog conflict.

bhuvan-somisetty:
Thanks @Kielek for the thorough review and the clear explanation.

You're right — the SSR/hydration split was unnecessary and introduced a mismatch risk. I've simplified the loader to a pure function that always returns a root-relative URL, identical across SSR and client renders. Also updated the changelog entry to match the wrapping and link style used by nearby entries.

Latest commit: 70cada98

Kielek:
please resolve changelog conflicts

bhuvan-somisetty:
@Kielek Updated as suggested — simplified the image loader to a pure function that always returns a root-relative URL, identical during SSR and client rendering. No more `window` check, no hydration mismatch. Changelog entry also reformatted to match the surrounding style. All checks pass. Would appreciate another look when you have a moment.

Kielek:
Please check 
image

bhuvan-somisetty:
@Kielek Rebased onto the latest main and resolved the CHANGELOG conflict - our entry is at the top of the Unreleased section, all main entries preserved below. Single clean commit, no conflicts remaining. All checks pass.

Kielek (approved):
LGTM, but it needs review from someone with netter js undertanding.

bhuvan-somisetty:
Hi @julianocosta89 @Kielek - EasyCLA is now passing and the branch is up to date. 
There are 2 workflows awaiting maintainer approval before build-test and CodeQL can 
run. Could one of you approve the workflows when you get a chance?
 Happy to address any further feedback!

julianocosta89 (approved):
Thank you @bhuvan-somisetty!

PR #3598, opened 2026-06-29
[chore] Skip telemetry re-run on dependabot PR approval

julianocosta89:
# Changes

skip telemetry re-run on dependabot PR approval
Dependabot PRs are already tested when opened/synced via the
pull_request trigger. Approving them was firing a second redundant
run through the pull_request_review path.


ps48 (approved):
(not blocking) There might be a theoretical edge case: dependabot PRs that don't match `pull_request.paths`

> If dependabot ever opens a PR touching only files outside the paths: list (e.g., root-level `.github/` config changes not listed), the pull_request trigger wouldn't fire, and now the approval-path won't either. In practice this doesn't happen since dep bumps always touch `src/**`, but worth a mental note if the paths filter is ever narrowed.

PR #3611, opened 2026-06-30
[chore] Add tests to agentic services

julianocosta89:
# Changes

## Add telemetry tests for agentic services

Adds trace coverage for the `agent`, `mcp`, and `chatbot` services introduced by `compose.agent.yaml`, along with a dedicated CI workflow that only runs when those services are modified.

### What changed

**Test infrastructure**
- `test/telemetry/services.py` — extended `SIGNAL_MATRIX` with the three agentic services (traces only; no OTLP metrics or log exporters configured). Added `AGENTIC_ONLY_SERVICES`, `AGENTIC_SERVICES`, and `AGENTIC_EDGES` (`chatbot→agent`). New `agentic` scope is handled by `services_for_scope` and `edges_for_scope`.
- `test/telemetry/test_agentic.py` — new test that asserts the Traceloop `@workflow` decorator emits an `astronomy_shop_agent_workflow` span in Jaeger for the `agent` service.
- `test/telemetry/conftest.py` — added agentic warmup probe: hits `agent /prompt` (any HTTP response confirms spans are emitted), then calls `list_products` via the MCP Streamable HTTP transport (preserving `Mcp-Session-Id` from the `initialize` response) to trigger an outbound httpx call that `HTTPXClientInstrumentor` captures as a span. Chatbot probe uses Gradio 6.x's `/gradio_api/call/respond` endpoint.
- `test/telemetry/README.md` — updated service matrix, `TEST_SCOPE` values, Makefile targets, env vars, and CI section.

**Makefile**
- `run-telemetry-tests-agentic` — starts the full stack with `compose.agent.yaml`, runs the test container with `TEST_SCOPE=agentic`.
- `build-agentic` — builds all services including `compose.agent.yaml` so `latest-agent`, `latest-mcp`, and `latest-chatbot` images are present in the CI artifact.

**CI workflow** (`.github/workflows/run-agentic-telemetry-tests.yml`)
- Triggered on `push` to `main` and `workflow_dispatch` when agentic paths change.
- On `pull_request`: runs automatically for dependabot PRs only.
- On `pull_request_review`: runs for human PRs after reviewer approval, with path filtering in the gate job. Dependabot PRs are skipped here — they already ran on `pull_request`.

### Notes

- `MCP_ENABLED` is intentionally left `false`. The VCR cassette was recorded with local `@tool` functions; enabling MCP changes the tool schemas and breaks cassette body matching. The `agent→mcp` edge is therefore not asserted; a comment in `services.py` documents this.
- The agentic warmup probe does not drive checkout traffic (`WARMUP_PROBE_ENABLED=false`) since those services have no checkout path.
## Merge Requirements

essential items:

### DIFF
--- a/.github/workflows/run-agentic-telemetry-tests.yml
+++ b/.github/workflows/run-agentic-telemetry-tests.yml
@@ -1,0 +1,189 @@
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+name: Agentic Telemetry Tests
+
+on:
+  # Dependabot PRs only. GitHub Actions does not support actor filters in the
+  # on: block, so the gate job below enforces the dependabot-only rule.
+  # Human PRs that touch these paths must be approved first — see
+  # pull_request_review below.
+  pull_request:
+    branches: [main]
+    paths:
+      - 'src/agent/**'
+      - 'src/mcp/**'
+      - 'src/chatbot/**'
+      - 'compose.agent.yaml'
+      - 'test/telemetry/**'
+      - '.env'
+      - '.github/workflows/run-agentic-telemetry-tests.yml'
+  # Human PRs: run after a reviewer approves a PR that touches agentic paths.
+  # Path filtering is done inside the gate job because GitHub does not apply
+  # paths filters to the pull_request_review event.
+  pull_request_review:
+    types: [submitted]
+  # Post-merge validation on main.
+  push:
+    branches: [main]
+    paths:
+      - 'src/agent/**'
+      - 'src/mcp/**'
+      - 'src/chatbot/**'
+      - 'compose.agent.yaml'
+      - 'test/telemetry/**'
+      - '.env'
+      - '.github/workflows/run-agentic-telemetry-tests.yml'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: agentic-telemetry-tests-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  # Decide whether the test jobs should run.
+  # For push/dispatch the paths filter above already ensures relevance.
+  # For pull_request_review we must check changed files ourselves because
+  # GitHub does not apply paths filters to that event.
+  gate:
+    name: "Agentic Test Gate"
+    runs-on: ubuntu-latest
+    outputs:
+      should_run: ${{ steps.check.outputs.should_run }}
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
+          fetch-depth: 0
+          persist-credentials: false
+
+      - id: check
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+          REVIEW_STATE: ${{ github.event.review.state }}
+          PR_STATE: ${{ github.event.pull_request.state }}
+          PR_AUTHOR: ${{ github.event.pull_request.user.login }}
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          # push / workflow_dispatch: paths filter already ensured relevance.
+          if [[ "$EVENT_NAME" == "workflow_dispatch" || "$EVENT_NAME" == "push" ]]; then
+            echo "should_run=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          # pull_request: only dependabot PRs run without approval.
+          # Human PRs must go through pull_request_review instead.
+          if [[ "$EVENT_NAME" == "pull_request" ]]; then
+            if [[ "$PR_AUTHOR" == "dependabot[bot]" ]]; then
+              echo "should_run=true" >> "$GITHUB_OUTPUT"
+            else
+              echo "should_run=false" >> "$GITHUB_OUTPUT"
+            fi
+            exit 0
+          fi
+
+          # pull_request_review: run when a human approves a PR that touches
+          # agentic paths. This is the only path for non-dependabot contributors.
+          # Dependabot PRs are skipped here — they already ran on pull_request.
+          if [[ "$EVENT_NAME" == "pull_request_review" \
+             && "$REVIEW_STATE" == "approved" \
+             && "$PR_STATE" == "open" \
+             && "$PR_AUTHOR" != "dependabot[bot]" ]]; then
+            git fetch --quiet origin "$BASE_SHA"
+            AGENTIC_PATTERN='^(src/(agent|mcp|chatbot)/|compose\.agent\.yaml|test/telemetry/|\.env$|\.github/workflows/run-agentic-telemetry-tests\.yml)'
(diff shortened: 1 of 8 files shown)


.github/workflows/run-agentic-telemetry-tests.yml
```diff
@@ -0,0 +1,189 @@
...
+
+on:
+  # Dependabot PRs only. GitHub Actions does not support actor filters in the
+  # on: block, so the gate job below enforces the dependabot-only rule.
+  # Human PRs that touch these paths must be approved first — see
+  # pull_request_review below.
+  pull_request:
+    branches: [main]
+    paths:
+      - 'src/agent/**'
```
Kielek, line 13: I think that you should add here also
`.github` or at least `.github/workflows`

.github/workflows/run-agentic-telemetry-tests.yml
```diff
@@ -0,0 +1,189 @@
...
+  # Human PRs: run after a reviewer approves a PR that touches agentic paths.
+  # Path filtering is done inside the gate job because GitHub does not apply
+  # paths filters to the pull_request_review event.
+  pull_request_review:
+    types: [submitted]
+  # Post-merge validation on main.
+  push:
+    branches: [main]
+    paths:
+      - 'src/agent/**'
```
Kielek, line 31: same here

.github/workflows/run-agentic-telemetry-tests.yml
```diff
@@ -0,0 +1,189 @@
...
+  # For push/dispatch the paths filter above already ensures relevance.
+  # For pull_request_review we must check changed files ourselves because
+  # GitHub does not apply paths filters to that event.
+  gate:
+    name: "Agentic Test Gate"
+    runs-on: ubuntu-latest
+    outputs:
+      should_run: ${{ steps.check.outputs.should_run }}
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
```
Kielek: I think that zizmor can be unhappy without specific names for each step.
It is pedantic persona ;)

.github/workflows/run-agentic-telemetry-tests.yml
```diff
@@ -0,0 +1,189 @@
...
+
+          # pull_request_review: run when a human approves a PR that touches
+          # agentic paths. This is the only path for non-dependabot contributors.
+          # Dependabot PRs are skipped here — they already ran on pull_request.
+          if [[ "$EVENT_NAME" == "pull_request_review" \
+             && "$REVIEW_STATE" == "approved" \
+             && "$PR_STATE" == "open" \
+             && "$PR_AUTHOR" != "dependabot[bot]" ]]; then
+            git fetch --quiet origin "$BASE_SHA"
+            AGENTIC_PATTERN='^(src/(agent|mcp|chatbot)/|compose\.agent\.yaml|test/telemetry/|\.env$|\.github/workflows/run-agentic-telemetry-tests\.yml)'
```
Kielek: and also here, paths needs to be adjusted.

julianocosta89:
@Kielek addressed! 😊

CHANGELOG.md
```diff
@@ -292,6 +292,8 @@ the release.
 * [cleanup] Remove tracetest
 ([#3602](https://github.com/open-telemetry/opentelemetry-demo/pull/3602),
   [#3603](https://github.com/open-telemetry/opentelemetry-demo/pull/3603))
+* [chore] Add tests to agentic services
```
cijothomas, line 295: nit: do we add such entries to changelog? they are not end user facing right?

julianocosta89: We never limited changelog to end-user facing changes. But that's a good point to consider

cijothomas: IMHO, we should only do it for end user facing changes. For developers/contributors, git blame exists :)

cijothomas: (Most OTel repos follow that. In face, most of the CI checks that enforce changelog entires skip it if the PR title has "chore" in it :))

Kielek:
One more, from Codex:
[P2] Expand the agentic path filters. The [trigger paths](https://github.com/open-telemetry/opentelemetry-demo/blob/d5b87d10fda4755ddfc9ba7b6ebd44b0dfec6c22/.github/workflows/run-agentic-telemetry-tests.yml#L12-L19) and [review-event regex](https://github.com/open-telemetry/opentelemetry-demo/blob/d5b87d10fda4755ddfc9ba7b6ebd44b0dfec6c22/.github/workflows/run-agentic-telemetry-tests.yml#L96-L98) omit inputs used by these tests: src/shared/**, Makefile, and the core compose layers. Changes there can break agent/MCP behavior without running agentic telemetry tests.

Other parts LGTM. but I am not feel fluent in python

julianocosta89:
> One more, from Codex: [P2] Expand the agentic path filters. The [trigger paths](https://github.com/open-telemetry/opentelemetry-demo/blob/d5b87d10fda4755ddfc9ba7b6ebd44b0dfec6c22/.github/workflows/run-agentic-telemetry-tests.yml#L12-L19) and [review-event regex](https://github.com/open-telemetry/opentelemetry-demo/blob/d5b87d10fda4755ddfc9ba7b6ebd44b0dfec6c22/.github/workflows/run-agentic-telemetry-tests.yml#L96-L98) omit inputs used by these tests: src/shared/**, Makefile, and the core compose layers. Changes there can break agent/MCP behavior without running agentic telemetry tests.
> 
> Other parts LGTM. but I am not feel fluent in python

@Kielek if we expand that, every change on `compose*` will build `agentic` as well.
 Do we really need that or is this just Codex being picky?!

Kielek:
IMO it is cheaper to execute to prevent issues on main. Especially if you consider non-returning contributors.
I am no telling that it is a must, but we SHOULD execute it.

julianocosta89:
> IMO it is cheaper to execute to prevent issues on main. Especially if you consider non-returning contributors. I am no telling that it is a must, but we SHOULD execute it.

@Kielek makes sense 🙏🏽 
Pushed

Kielek (approved):

test/telemetry/README.md
```diff
@@ -127,18 +132,28 @@ make run-telemetry-tests-minimal   # Minimal scope
...
-| `TEST_SCOPE`             | `minimal`    | `minimal` or `full`                             |
+| `TEST_SCOPE`             | `minimal`    | `minimal`, `full`, or `agentic`                 |
 | `WARMUP_SECONDS`         | `240`        | Max seconds to wait for backends before testing |
 | `POLL_TIMEOUT`           | `180`        | Per-test seconds to poll a backend for data     |
 | `WARMUP_PROBE_ENABLED`   | `true`       | Drive checkouts during warmup (see Approach)    |
 | `WARMUP_PROBE_CHECKOUTS` | `5`          | Number of checkouts the warmup probe drives     |
 | `WARMUP_PROBE_TIMEOUT`   | `120`        | Max seconds to complete warmup probe checkouts  |
+| `AGENT_URL`              | (derived)    | Override agent service URL (agentic scope)      |
+| `MCP_URL`                | (derived)    | Override MCP service URL (agentic scope)        |
+| `CHATBOT_URL`            | (derived)    | Override chatbot service URL (agentic scope)    |
```
ps48: Nit: The `AGENT_URL/MCP_URL/CHATBOT_URL` override env vars add config surface that has no exercised path the test container always runs on the compose network where the derived defaults (`http://agent:8010` etc.) work. Should we consider dropping the overrides and just using the derivation directly; fewer knobs = less to document and fewer ways to misconfigure.

Kielek (approved):

PR #3632, opened 2026-07-06
Bound concurrent browser sessions in load generator

### LINKED ISSUE #2678: [load-generator] High Memory Consumption
# Bug Report

Which version of the demo you are using?:
- https://github.com/open-telemetry/opentelemetry-demo/releases/tag/2.1.3

## Symptom

Load generator (locust) eats all the available RAM that i set in the memory limit in k8s.

I increased from the default value used in the chart (1500Mi) to 2000Mi and, eventually, it will use all of it.

Image

## Reproduce
I deployed the opentelemetry-helm chart with default values.

Maybe we should modify `locustfile.py` to trigger GCs more often?

bhuvan-somisetty:
# Changes

Closes #2678.

`WebsiteBrowserUser` (the Playwright-based user used when
`LOCUST_BROWSER_TRAFFIC_ENABLED=true`) keeps its own headless Chromium
process alive for its whole lifetime once locust spawns it —
`locust-plugins`'s `@pw` decorator recycles the page/context between
tasks, but never touches the top-level browser process. Neither
`WebsiteUser` nor `WebsiteBrowserUser` set an explicit `weight`, so
Locust defaulted both to `1`, meaning roughly half of `LOCUST_USERS`
became persistent Chromium processes. That scales memory linearly with
user count and is exactly what's reported in the issue (OOM once
`LOCUST_USERS` grows past a couple of instances).

This gives `WebsiteBrowserUser` a low `weight` relative to
`WebsiteUser` (1:9), so only a small, bounded share of virtual users
spawn a browser process no matter how `LOCUST_USERS` is scaled up.
Browser-based traffic (and the frontend telemetry it generates) is
unchanged in kind, just proportionally rarer.

- `src/load-generator/locustfile.py`: added `weight = 9` to
  `WebsiteUser` and `weight = 1` to `WebsiteBrowserUser`.
- `CHANGELOG.md`: added an entry under Unreleased.

## Merge Requirements

* [x] Appropriate documentation updates in the [docs][] — no docs
  reference specific user weights today, so none needed
* [ ] Appropriate Helm chart updates in the [helm-charts][] — not
  needed, this only changes in-process Locust user weighting, not any
  deployment configuration


bhuvan-somisetty:
@julianocosta89 @Kielek could either of you take a look when you have a moment? This closes #2678 -WebsiteBrowserUser had no explicit weight, so it defaulted to a 1:1 ratio against WebsiteUser and spawned a persistent Chromium process for roughly half of LOCUST_USERS, scaling memory linearly with user count. Lowering its weight bounds the number of concurrent browser processes regardless of scale. All checks are green.

julianocosta89:
@bhuvan-somisetty please update the repo and point your coding agent to the new AGENTS.md.

Your PR adds 2 lines, but the agent is adding 7 lines of comments for something that doesn't need a comment.
I hope the new AGENTS.md solves that.

Another thing that I'm a bit concerned regarding your change, does that mean that we will have way less traffic on the `frontend-web`?

bhuvan-somisetty:
Thanks for the feedback! Pushed a follow-up:

- Trimmed the comment down to one line and made the weight configurable via a new `LOCUST_BROWSER_USER_WEIGHT` env var (default 1) instead of hardcoding the ratio.
- On the frontend-web traffic question: yes, by default this does reduce how many virtual users hit real page loads/RUM instrumentation through the browser, since previously roughly half of `LOCUST_USERS` were spawning a Chromium session. That's what was driving the memory usage, so some reduction is the actual fix. Making the weight configurable means anyone who wants more frontend/browser coverage than the default can just raise `LOCUST_BROWSER_USER_WEIGHT`, without reverting to the memory-scaling behavior for everyone else.

@julianocosta89 let me know if you'd like a different default ratio, happy to adjust.

.env
```diff
@@ -102,6 +102,10 @@ LOCUST_USERS=5
 LOCUST_HOST=http://${FRONTEND_PROXY_ADDR}
 LOCUST_WEB_HOST=load-generator
 LOCUST_AUTOSTART=true
+# Locust weight for the Playwright-based browser user relative to the plain
+# HTTP user (default weight 9). Raise this to send more real browser/frontend
+# traffic at the cost of more concurrent Chromium processes.
```
julianocosta89:
```suggestion
```

bhuvan-somisetty:
Dropped the comment, thanks. And it should already be passed through - compose.yaml has `LOCUST_BROWSER_USER_WEIGHT` in the load-generator environment list (line 427). Let me know if you're still seeing it missing anywhere.

bhuvan-somisetty:
@julianocosta89 pushed the fix, all checks are green again. Ready for another look whenever you're free.

bhuvan-somisetty:
@Kielek would you also have a moment to review this one? All checks are green.

src/load-generator/locustfile.py
```diff
@@ -111,6 +111,7 @@ def get_flagd_value(FlagName):
 people = json.load(people_file)
 
 class WebsiteUser(HttpUser):
+    weight = 9
```
ps48: If someone sets `LOCUST_BROWSER_USER_WEIGHT=9`, the ratio becomes 1:1 again (same as before the fix). But if they want more browser traffic without reverting to the OOM behavior, they can't increase browser weight beyond 9 without it becoming the majority. 

Should we consider making WebsiteUser weight also configurable and document the configuration behavior in https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/load-generator/README.md?

bhuvan-somisetty:
Good catch, thanks @ps48. Made LOCUST_HTTP_USER_WEIGHT configurable too and added a Configuration section to the load generator README covering both vars. All checks are green.

@julianocosta89 @ps48 ready for another look whenever you're free.

src/load-generator/locustfile.py
```diff
@@ -224,6 +225,9 @@ def on_start(self):
 
 if browser_traffic_enabled:
     class WebsiteBrowserUser(PlaywrightUser):
+        # Bounds how many concurrent Chromium processes run, since each one
+        # persists for the user's lifetime instead of being recycled per task.
```
julianocosta89:
```suggestion
```

Please avoid adding comments where they are not needed.

bhuvan-somisetty:
Removed the comment, branch is up to date with main, all checks green.

@julianocosta89 ready for another look.

julianocosta89 (approved):

julianocosta89 (approved):

Issue #1731, opened 2024-09-30
Supporting read-only filesystems


mhausenblas:
# Feature Request

**Is your feature request related to a problem?**

In environments with heightened security runtimes for container platforms, running the demo applications from containers with writable root filesystems on the container presents a problem due to security concerns which in turn makes it more difficult to use the demo application as a showcase for OpenTelemetry.  

**Describe the solution you'd like:**

We would like to have exposed either configuration options and/or instructions on best practices to be able to run the demo app on a container platform (such as Kubernetes) with the option to enable a read-only root filesystem for the demo app containers. This would make the demo application much easier to adopt in environments with restricted security runtimes where the use of the root filesystem on containers is prohibited and instead temporary filesystems (e.g., ephemeral volumes) need to be used.

**Describe alternatives you've considered.**

N/A

## Additional Context

N/A

PR #3658, opened 2026-07-09
Add opt-in read-only root filesystem support for checkout, product-catalog, shipping

bhuvan-somisetty:
# Changes

Towards #1731.

Environments with hardened container security runtimes (e.g. Kubernetes Pod
Security Standards "restricted") often prohibit writable root filesystems,
which the demo doesn't currently support for any service.

This adds `compose.readonly.yaml` as an additional compose layer, following
the same pattern already used by `compose.full.yaml` and
`compose.observability.yaml`. It covers `checkout`, `product-catalog`, and
`shipping` — the services I could confirm have no runtime file writes in
their own source (all distroless Go/Rust binaries doing network I/O only:
gRPC, Kafka, a DB connection). Each gets `read_only: true` with a `/tmp`
tmpfs mount as a safety margin.

This is intentionally a starting point, not full coverage. Several other
services write files at startup as part of normal operation and would need
dedicated tmpfs mounts of their own before they could be added safely:
nginx (`image-provider`, `telemetry-docs`) and Envoy (`frontend-proxy`) all
render their runtime config from a template into their own filesystem on
every start, and the load generator's headless Chromium session needs
writable scratch space. I didn't want to guess at those without being able
to verify them running.

Usage:

```sh
docker compose -f compose.yaml -f compose.readonly.yaml up
```

- `compose.readonly.yaml`: new opt-in layer.
- `compose.yaml`: documented the new layer alongside the existing ones in
  the top-of-file usage comment.
- `CHANGELOG.md`: added an entry under Unreleased.

## Merge Requirements

* [x] Appropriate documentation updates in the [docs][] — usage documented
  in `compose.yaml`'s own header, matching how the other compose layers are
  documented; no external docs reference this yet
* [ ] Appropriate Helm chart updates in the [helm-charts][] — not needed,
  this only adds a new opt-in Docker Compose layer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ the release.
 
 ## Unreleased
 
+* [compose] Run `checkout`, `product-catalog`, and `shipping` with a
+  read-only root filesystem (`read_only: true` plus a `/tmp` tmpfs mount),
+  for container platforms that prohibit writable root filesystems. Limited
+  to these three services for now since they're the ones verified to have
+  no runtime file writes of their own; other services write files at
+  startup and need dedicated tmpfs mounts before they can be switched over
+  ([#1731](https://github.com/open-telemetry/opentelemetry-demo/issues/1731))
 * [llm] Increase `llm` service memory limit from 50M to 100M to prevent a
   startup restart loop caused by the container exceeding its memory limit
   ([#2944](https://github.com/open-telemetry/opentelemetry-demo/issues/2944))
--- a/compose.yaml
+++ b/compose.yaml
@@ -165,6 +165,9 @@ services:
         condition: service_started
       flagd:
         condition: service_started
+    read_only: true
+    tmpfs:
+      - /tmp
     logging: *logging
 
   # Currency service
@@ -545,6 +548,9 @@ services:
         condition: service_healthy
     volumes:
       - ./otel-config.yml:/otel-config.yml
+    read_only: true
+    tmpfs:
+      - /tmp
     logging: *logging
 
   # Quote service
@@ -663,6 +669,9 @@ services:
         condition: service_started
       flagd:
         condition: service_started
+    read_only: true
+    tmpfs:
+      - /tmp
     logging: *logging
 
   # ******************


bhuvan-somisetty:
@julianocosta89 @Kielek could either of you take a look when you have a moment? Scoped to three services I could verify have no runtime file writes, all checks are green.

julianocosta89:
Hello @bhuvan-somisetty, I don't understand this change. Could you please elaborate?
1. What happened on `CHANGELOG`?
2. Why do we need an extra `compose.readonly` file?
3. Does changing the services to read-only has any impact on the service itself? If not, I'd suggest just changing the behaviour of the services as a whole, so then we do not need to have 2 versions of each.

bhuvan-somisetty:
Thanks for the questions, all fair. I pushed an update:

1. That was a mistake on my end — an earlier "update branch" merge pulled in a bunch of already-merged Unreleased entries as duplicates instead of just my one line. I rebased onto current main so the branch is now a single clean commit and the CHANGELOG diff is just the one entry.

2. It was meant to be opt-in since I'd only verified 3 of the ~15 services, and didn't want to change default behavior for the others without checking them too. But that reasoning doesn't hold up against your point below.

3. No impact — checkout, product-catalog, and shipping are distroless Go/Rust binaries doing gRPC/Kafka/DB I/O only, nothing writes to the filesystem outside of what the /tmp tmpfs mount already covers. So you're right, no reason to keep two versions of these services around. I dropped compose.readonly.yaml and just set `read_only: true` + the tmpfs mount directly on those three services in compose.yaml.

The other services (image-provider, telemetry-docs, frontend-proxy, load-generator) do write files at startup (nginx/Envoy config templating, Chromium scratch space), so I left those out until I can verify them properly — happy to look into those as a follow-up if that's useful.

Diff is down to 16 lines now, all checks are green.

julianocosta89 (approved):
Awesome!

bhuvan-somisetty:
@julianocosta89 Thanks for taking another look, glad it's a cleaner change now. All checks are green.

julianocosta89:
@bhuvan-somisetty just one last thing just FYI, I've updated  the PR description to`Towards #1731`, as the case is not fully closed yet.

Issue #2144, opened 2025-04-08
internal server errors found on remote web ui for otel-demo app


sophiaxu0424:
# Bug Report

Which version of the demo you are using? (please provide either a specific
[commit hash](75b4e53f0bb906a5e17e0a5afb2d57c212eb41fe)

## Symptom
opentelemetry-demo is seeing internal server errors in in a remote machine (otel-app was deployed on a redhat VM) web browser:

<img width="1702" alt="Image" src="https://github.com/user-attachments/assets/887d8b27-2a2f-45ac-98f8-e177e1c16128" />

And clicking the buttons for go shopping or adding product into cart doesn't work normally for now.

**What is the expected behavior?**
Web UI should be working normal as a shopping app, and no internal server error (500) in the console.

What do you expect to see?
Web UI should be working normal as a shopping app, and no internal server error (500) in the console.

**What is the actual behavior?**
shopping button doesn't work, console errors

Please describe the actual behavior experienced.

## Reproduce

Could you provide the minimum required steps to resolve the issue you're seeing?

follow the steps showing on docker deployment for otel-demo app

We will close this issue if:

* The steps you provided are complex.
* If we can not reproduce the behavior you're reporting.

## Additional Context

Please feel free to add any other context about the problem here.

morningspace:
Thanks @julianocosta89 for the quick reply.

Just to add, we are also seeing that the frontend service keeps reporting errors as below.

```console
PageNotFoundError: Cannot find module for page: /500
    at getPagePath (/app/node_modules/next/dist/server/require.js:94:15)
    at requirePage (/app/node_modules/next/dist/server/require.js:99:22)
    at /app/node_modules/next/dist/server/load-components.js:98:84
    at ZoneDelegate.invoke (/app/node_modules/zone.js/bundles/zone.umd.js:400:30)
    at Zone.run (/app/node_modules/zone.js/bundles/zone.umd.js:160:47)
    at /app/node_modules/zone.js/bundles/zone.umd.js:1318:38
    at ZoneDelegate.invokeTask (/app/node_modules/zone.js/bundles/zone.umd.js:434:35)
    at Zone.runTask (/app/node_modules/zone.js/bundles/zone.umd.js:205:51)
    at drainMicroTaskQueue (/app/node_modules/zone.js/bundles/zone.umd.js:620:39)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
 ⨯ Error: Could not parse target name ""
[...]

PR #3664, opened 2026-07-10
Add custom 404 and 500 error pages to frontend

bhuvan-somisetty:
# Changes

`_app.tsx` defines a custom `getInitialProps`, which opts every page out of
Next.js's automatic static optimization -- including the framework's
built-in 404/500 fallback pages. With no `pages/404.tsx` or `pages/500.tsx`
of our own, that leaves nothing for Next.js to render when a server-side
error happens, and it crashes instead with:

```
PageNotFoundError: Cannot find module for page: /500
```

This surfaced in #2144, where a server-side error during a remote
deployment triggered exactly this crash in the frontend logs. That issue
is mostly a remote-deployment/config thread, but the missing error pages
are a real, separate gap worth fixing on their own -- this PR just
addresses that piece, not the full thread.

Both new pages are intentionally self-contained (no `Layout`/`Header`, no
data fetching), since Next.js may render them on either the client or the
server and they shouldn't depend on anything that could itself fail.

Verified with a production build (`/404` and `/500` now show up as
compiled routes, where before they didn't exist at all) and by hitting
both routes directly against a `next start` server -- they return the
correct status code and render their own content instead of crashing.

## Merge Requirements

* [ ] Appropriate documentation updates in the [docs][] -- not needed, this
  is an internal frontend fix with no user-facing config or usage change
* [ ] Appropriate Helm chart updates in the [helm-charts][] -- not needed,
  no deployment/service changes


bhuvan-somisetty:
@julianocosta89 @Kielek could either of you take a look when you have a moment? Small, self-contained fix (two new pages + shared styles), no other behavior changed. All checks are green.

bhuvan-somisetty:
@julianocosta89 @Kielek friendly bump on this one whenever you get a chance, still all green.

julianocosta89 (approved):
Thank you @bhuvan-somisetty

PR #3690, opened 2026-07-14
fix(docker): Propagate default histogram aggregation setting

ywwg:
# Changes

Propogate OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION to containers so the value can be overridden in .env.override

## Merge Requirements

essential items:


julianocosta89:
hey @ywwg thanks for contributing to the Demo!

I think we need `OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION= explicit_bucket_histogram` to be set on the `.env` file.
I know that this is the default, but as we are adding the env var we should add the value as well.

CHANGELOG.md
```diff
@@ -7,10 +7,9 @@ the release.
 
 ## Unreleased
 
-* [opamp-server] Bump `OPAMP_GO_REF` to pick up example-server UI improvements
-  from `opamp-go` (improved agent list, agent uptime on the agent page, and
-  startup logging of the OpAMP/admin UI addresses).
-  ([#3685](https://github.com/open-telemetry/opentelemetry-demo/pull/3685))
+* [docker] fix `compose.yaml` to allow
```
cijothomas, line 10: not all services support this env variable, so this may give the wrong impression to an user that they can set this and all services would honor it.
(this is also an existing problem for other env variables too).

perhaps call out that not all services support this?

julianocosta89: @cijothomas is there an easy way to find which SDKs are already supporting this?
In the docs we do not say anything :/

cijothomas: https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#environment-variables this is the best we have. No guarantee it is 100% accurate. (I can confirm Rust does not implement it, but unsure of others)

My comment was more generic - ENV variable are optional per-spec. and support varies. One day, we'll be able to use declarative config, but until then, nothing much to be done about ENV variables, other than documented/warning users that support varies by language, and hence, setting it may not have desired effect.

Even website is just showing a warning for this:
https://github.com/open-telemetry/opentelemetry.io/pull/9275/changes

PR #3696, opened 2026-07-16
[flagd-ui] Use LiveView navigation between the Basic and Advanced tabs

### LINKED ISSUE #2588: [flagd-ui] Spans with error
# Bug Report

[fe0458c](https://github.com/open-telemetry/opentelemetry-demo/commit/fe0458cd35100d0efa9394e6421b7ee6a39956e9)

## Symptom

No errors on spans.

Every time we open the `/feature` page we get a trace containing 4 spans that are properly working.
If we navigate between `Basic` and `Advanced` tab we get a different trace that contains 3 spans, and this one contains an error: `connection termination`.

Image

## Reproduce

Clone the repo, start the demo and navigate to `/feature`.

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

bhuvan-somisetty:
# Changes

Fixes #2588.

The `Basic` and `Advanced` tabs in the flagd-ui navbar were plain `<a href>` links, so switching tabs triggered a full browser page load. That tears down the existing LiveView WebSocket, and the aborted connection is recorded as a span with a `connection termination` error — the 3-span trace described in the issue.

Switching the links to `<.link navigate={...}>` keeps navigation on the existing socket, so no connection is torn down and no error span is produced.

Two related cleanups fell out of this:

- The links hardcoded the `/feature` prefix, which is only correct in prod — the endpoint sets `path: "/feature"` in `config.exs`, but dev and test override it to `/`. Using the `~p` sigil derives the prefix from the endpoint config instead, so the navbar works in every environment.
- The module did both `use Phoenix.Component` and `use FlagdUiWeb, :live_view`, which defines `__phoenix_component_verify__/1` twice and produces a redundant-clause warning as soon as a function component is used. It's a component module, so `use FlagdUiWeb, :html` is the right entrypoint.

Added `test/flagd_ui_web/live/navbar_test.exs` covering both directions. The tests assert the click produces a `live_redirect` rather than a `redirect` — against the old navbar they fail with `{:redirect, %{to: "/feature/advanced"}}`, which is the full page load at the root of this bug. LiveView 1.2's test helpers need `lazy_html`, so that's added as a test dependency.

## Merge Requirements


bhuvan-somisetty:
@julianocosta89 @Kielek @cijothomas @ps48 this is ready for review when you have a chance — all checks are green.

It fixes the `connection termination` error span from #2588. The `Basic`/`Advanced` tabs were plain `<a href>` links, so switching tabs did a full page load, which tore down the LiveView WebSocket and produced that error span. Using `<.link navigate={...}>` keeps navigation on the existing socket, so nothing gets torn down. Added tests that fail on the old navbar with `{:redirect, %{to: "/feature/advanced"}}` — the full page load itself.

julianocosta89 (approved):

julianocosta89:
@bhuvan-somisetty LGTM!
I've tested and the error is gone!
Thx for that

bhuvan-somisetty:
Thanks for testing it @julianocosta89! Glad the error is gone on your side too.

bhuvan-somisetty:
@julianocosta89 
 anything else needed here before this can go in?

julianocosta89:
> @julianocosta89 anything else needed here before this can go in?

@bhuvan-somisetty me hitting the merge button 😅 

I was waiting the tests to run

PR #3721, opened 2026-07-20
fix(flagd-ui): remove unreachable clause in Storage.update_flag/3

### LINKED ISSUE #3718: [flagd-ui] Dead/unreachable clause in Storage.update_flag/3
# Bug Report

Which version of the demo you are using? `main` at commit
[a140185](https://github.com/open-telemetry/opentelemetry-demo/commit/a14018590bb21fbb04276899c4f4f5c847746661)

## Symptom

In [`src/flagd-ui/lib/flagd_ui/storage.ex`](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/flagd-ui/lib/flagd_ui/storage.ex#L57-L78),
the private `update_flag/3` function passes a 3-clause anonymous function to
`Enum.map/2`:

```elixir
defp update_flag(flags, flag_name, value) do
  flags
  |> Enum.map(fn
    {flag, data} when flag == flag_name ->
      updated_data =
        case get_in(data, ["targeting", "if"]) do
          [_cond, _on, _off] ->
            put_in(data, ["targeting", "if", Access.at(1)], value)
          _ ->
            Map.replace(data, "defaultVariant", value)
        end
      {flag, updated_data}
    {flag, data} when flag == flag_name -> {flag, Map.replace(data, "defaultVariant", value)}
    {flag, data} -> {flag, data}
  end)
  |> Map.new()
end
```

The first clause (lines 60-73) and the second clause (line 74) have the
**exact same pattern and guard**: `{flag, data} when flag == flag_name`.
Clauses are matched top to bottom, so whenever an element matches that
guard, the first clause always fires — the second clause can never run.

```mermaid
flowchart TD
    Start["Enum.map element: {flag, data}"] --> Check1{"flag == flag_name?"}
    Check1 -- yes --> C1["Clause 1 (lines 60-73)\ntargeting-aware update\nALWAYS taken when it matches"]
    Check1 -- no --> C3["Clause 3 (line 75)\n{flag, data} -> {flag, data}"]
    C2["Clause 2 (line 74)\nsame pattern + guard as Clause 1"]
    C1 -. "shadows" .-> C2
    classDef dead fill:#5b1a1a,stroke:#d73a4a,color:#fff;
    class C2 dead;
```

Every clause in the function should be reachable. Clause 1 already covers
everything clause 2 does — including the exact same fallback,
`Map.replace(data, "defaultVariant", value)`, on line 70 — so clause 2 looks
like a leftover from an earlier version of the function that was never
removed once clause 1 grew to handle the `targeting.if` case too.

`storage.ex:74` is dead code. It has the identical pattern and guard as the
clause immediately above it, so it's unreachable.

## Reproduce

1. Open `src/flagd-ui/lib/flagd_ui/storage.ex` and look at `update_flag/3`,
   lines 57-78.
2. Compare the guard on line 60 with the guard on line 74 — both are
   `{flag, data} when flag == flag_name ->`.
3. Since clause 1 already matches (and returns from) every case where
   `flag == flag_name`, clause 2 never gets a chance to execute.

## Additional Context

Runtime behavior isn't affected today, since clause 1's fallback branch
(line 70) already does what clause 2 was doing — this is a dead-code /
code-cleanliness issue rather than a functional regression. Suggested fix
is to delete the redundant clause on line 74:

```elixir
defp update_flag(flags, flag_name, value) do
  flags
  |> Enum.map(fn
    {flag, data} when flag == flag_name ->
      updated_data =
        case get_in(data, ["targeting", "if"]) do
          [_cond, _on, _off] ->
            put_in(data, ["targeting", "if", Access.at(1)], value)
          _ ->
            Map.replace(data, "defaultVariant", value)
        end
      {flag, updated_data}
    {flag, data} -> {flag, data}
  end)
  |> Map.new()
end
```

Scope is a single line removed in `storage.ex`.

<sub>**Tip**: [React](https://github.blog/news-insights/product-news/add-reactions-to-pull-requests-issues-and-comments/) with 👍 to help prioritize this issue. Please use comments to provide useful context, avoiding `+1` or `me too`, to help us triage it. Learn more [here](https://opentelemetry.io/community/end-user/issue-participation/).</sub>

bhuvan-somisetty:
Fixes #3718

In src/flagd-ui/lib/flagd_ui/storage.ex, update_flag/3 matches over a list of {flag, data} tuples with three clauses:

    {flag, data} when flag == flag_name -> ...
    {flag, data} when flag == flag_name -> {flag, Map.replace(data, "defaultVariant", value)}
    {flag, data} -> {flag, data}

The first two clauses have the exact same pattern and guard. Elixir matches clauses top to bottom, so whenever flag == flag_name the first clause always wins and the second one can never run. It's dead code.

It also turns out the first clause's fallback branch (the `_ -> Map.replace(data, "defaultVariant", value)` inside its case) does the same thing the second clause was doing, so nothing changes at runtime. Deleted the second clause.

## Merge Requirements

* [ ] CHANGELOG.md updated (not applicable, no user-facing change)
* [ ] Appropriate documentation updates in the docs (not applicable)
* [ ] Appropriate Helm chart updates (not applicable)


bhuvan-somisetty:
All checks are green here, small one-line dead code cleanup. @julianocosta89 @cijothomas @Kielek could one of you take a look when you get a chance?

julianocosta89:
hey @bhuvan-somisetty thank you for all the contributions, please keep them coming!!!

I noticed you always tag us on the PR, but there is no need for that.
All Demo approvers and maintainers get a notification when a new PR is opened, no need to tag us directly on the PR, unless you need an input, or if the PR is opened for a long time without any comments/reviews.

bhuvan-somisetty:
Ah got it, thanks for the heads up! Didn't realize approvers get notified automatically on new PRs. Will hold off on tagging going forward and only do it if a PR's been sitting quiet for a while.

julianocosta89 (approved):

bhuvan-somisetty:
Looked into the two merge queue failures, both are infra flakiness unrelated to this change:

- First one: the flagd-ui image build hit `Connection reset by peer` talking to the Debian mirror during `apt-get install`.
- Second one: an arm64 cross-build segfault (`qemu: uncaught target signal 11`) while configuring `libc-bin` under QEMU emulation.

Both happened in the same job (flagd-ui image build), which is the heaviest multi-arch build in the matrix, so it's the one most exposed to this kind of transient runner flakiness. Nothing in `storage.ex` is involved in either failure. Should be safe to add back to the merge queue whenever convenient.

osullivandonal (approved):

julianocosta89:
@bhuvan-somisetty the issue on the build seems to be related to your change, would you be able to take a look?

bhuvan-somisetty:
Took a look. Ran `mix test` locally 8 times in a row (13 tests each run) and got 0 failures.

The CI failure trace shows `Storage.init/1` (storage.ex line 23) getting a `Jason.DecodeError` on an empty read of the shared `demo.flagd.json` test fixture. That line, and the whole read path, is untouched by this PR — the only change here is removing a duplicate, unreachable pattern-match clause inside `update_flag/3`, which by definition never executes (the first clause with the identical guard always matches first), so it can't affect timing or file I/O anywhere else in the app.

This is also the third different failure signature across three merge queue attempts on this PR (a Debian mirror reset, then a QEMU segfault in `libc-bin`, now this), and all three happened in the same job: the flagd-ui image build, which is the one Dockerfile in the repo that actually boots the BEAM VM and runs `mix test` under QEMU arm64 emulation as part of the build. That points to flakiness in that cross-build step rather than something in this diff. Happy to dig further if it keeps happening on a clean run.

julianocosta89 (approved):

julianocosta89:
Let's try again!

PR #3736, opened 2026-07-21
fix(react-native-app): use getUniqueIdSync() for device.id instead of getDeviceId()

### LINKED ISSUE #3722: [react-native-app] device.id resource attribute uses getDeviceId() (hardware model) instead of a unique per-device identifier
## Describe the bug

In `src/react-native-app/hooks/useTracer.ts`, the `device.id` resource attribute is populated using `getDeviceId()` from `react-native-device-info`:

https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/react-native-app/hooks/useTracer.ts#L58

```ts
const resource = resourceFromAttributes({
  [ATTR_SERVICE_NAME]: "react-native-app",
  [ATTR_OS_NAME]: Platform.OS,
  [ATTR_OS_VERSION]: getSystemVersion(),
  [ATTR_SERVICE_VERSION]: getVersion(),
  [ATTR_DEVICE_ID]: getDeviceId(), // <-- not a unique identifier
});
```

The problem is that `getDeviceId()` does **not** return a unique-per-device value. Per the `react-native-device-info` docs, it returns the hardware/model identifier (e.g. `iPhone13,4`, `SM-G973F`), which is identical for every physical unit of that model. The method that actually returns a persistent, per-install/per-device identifier is `getUniqueId()`.

This conflicts with the OpenTelemetry semantic convention for [`device.id`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/device/), which defines it as:

> A unique identifier representing the device... SHOULD be identical for all apps on a device and SHOULD NOT change if an app is uninstalled and re-installed.

Because the demo currently sets `device.id` to a model string instead of a unique identifier, every trace/span emitted by two different phones of the same model (e.g. two testers both on an `iPhone13,4`) will carry the **same** `device.id` value. Any dashboard, exemplar, or debugging session that groups or filters telemetry by `device.id` — which is one of the more instructive things this attribute is meant to demonstrate in the demo — will silently merge unrelated devices/users together.

## Where this shows up

```mermaid
flowchart LR
    A["react-native-app starts"] --> B["useTracer() hook runs"]
    B --> C["setupTracerProvider()"]
    C --> D["getDeviceId() called"]
    D --> E["ATTR_DEVICE_ID = 'iPhone13,4'"]
    E --> F["Span/Resource exported via OTLP"]
    F --> G{"Two different physical\ndevices, same model"}
    G -->|"Device A"| H["device.id = iPhone13,4"]
    G -->|"Device B"| I["device.id = iPhone13,4"]
    H --> J["Backend cannot\ndistinguish A from B"]
    I --> J
```

## Current vs. expected value

| API call | What it actually returns | Example value | Unique per physical device? |
|---|---|---|---|
| `getDeviceId()` (currently used) | Hardware/model identifier | `iPhone13,4` | No — same for every unit of that model |
| `getUniqueId()` (expected) | Persistent per-install/device identifier (IDFV on iOS, Android ID / generated UUID on Android) | `E621E1F8-C36C-495A-93FC-0C247A3E6E5F` | Yes |

## Steps to reproduce

1. Run the `react-native-app` demo component on two different physical devices (or simulators) that report the same hardware model.
2. Trigger any traced action (e.g. browsing products) on both devices.
3. Inspect the exported resource attributes (e.g. in the Jaeger/OTel Collector output) for `device.id` on the spans from each device.
4. Observe that `device.id` is identical across both devices.

## Expected behavior

`device.id` should be a value that uniquely identifies the individual device/install, matching the OpenTelemetry semantic convention definition, so telemetry from different physical devices can be told apart.

## Suggested fix

Swap `getDeviceId()` for `getUniqueId()` (both exported by `react-native-device-info`) when populating `ATTR_DEVICE_ID` in `setupTracerProvider()`. If the hardware/model string is still useful to keep around for other purposes, it should be reported separately (e.g. under `device.model.identifier`), not conflated with `device.id`.

## Component(s)

react-native-app

## Additional context

Relevant `react-native-device-info` API reference: https://github.com/react-native-device-info/react-native-device-info#readme

bhuvan-somisetty:
# Changes

Fixes #3722

`src/react-native-app/hooks/useTracer.ts` was populating the `device.id` resource attribute with `getDeviceId()` from `react-native-device-info`. That method returns the hardware/model identifier (e.g. `iPhone13,4`), which is the same for every physical unit of that model — it's not a per-device value.

That conflicts with the OTel semantic convention for `device.id`, which requires a unique identifier per device. As a result, two different physical devices of the same model would show up with an identical `device.id` on their spans.

Swapped it for `getUniqueIdSync()`, which returns a persistent per-install/device identifier (IDFV on iOS, a generated/Android ID on Android), and kept the change synchronous so it drops straight into the existing `resourceFromAttributes()` call alongside `getSystemVersion()`/`getVersion()`.

## Merge Requirements

* [x] `CHANGELOG.md` updated
* [ ] Appropriate documentation updates in the docs (not applicable)
* [ ] Appropriate Helm chart updates (not applicable)


julianocosta89:
@bhuvan-somisetty can you add a changelog for this?
I know it is a small change, but I think it is worth it

bhuvan-somisetty:
Good call, added a changelog entry.
all checks are green, pls have a look

julianocosta89 (approved):

PR #3741, opened 2026-07-21
fix(currency): correct IPV6_ENABLED string comparison

### LINKED ISSUE #3735: [currency] IPV6_ENABLED check uses pointer comparison instead of string comparison, so the toggle never works
## Describe the bug

`IPV6_ENABLED` is supposed to make a service bind to `[::]` instead of `0.0.0.0` (this was introduced across services to resolve #2482). In the currency service, this switch is silently broken because of a C-string pointer comparison instead of a content comparison.

https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/currency/src/server.cpp#L256-L263

```cpp
void RunServer(uint16_t port)
{
  std::string ip("0.0.0.0");

  const char* ipv6_enabled = std::getenv("IPV6_ENABLED");

  if (ipv6_enabled == "true") {
    ip = "[::]";
    logger->Info(eventName("currency.server.ip_overwrite"),
                 "Overwriting Localhost IP",
                 opentelemetry::common::MakeAttributes({{"server.address", ip.c_str()}}));
  }
  ...
```

`std::getenv` returns a `const char*`. `ipv6_enabled == "true"` compares that pointer against the address of the string literal `"true"` in the binary — i.e. it's a **pointer comparison, not a string-value comparison**. Those two addresses can never be equal at runtime, so `ipv6_enabled == "true"` evaluates to `false` no matter what the `IPV6_ENABLED` environment variable is actually set to (`"true"`, `"false"`, unset, anything). The branch that switches to `[::]` and logs `currency.server.ip_overwrite` is dead code — it can never execute.

## Why this matters

This exact `IPV6_ENABLED` toggle was added to every relevant service specifically to fix #2482 ("IPv6 Compatibility" — the demo was inaccessible in IPv6-only environments because of hardcoded IPv4 binds). Every other service that implements the identical feature compares the *value*, not the pointer, so the toggle works there — the currency service is the one outlier where the fix never actually took effect:

| Service | File | Comparison used | Works? |
|---|---|---|---|
| currency (C++) | `src/currency/src/server.cpp:258` | `ipv6_enabled == "true"` (raw `const char*` pointer compare) | ❌ always false |
| payment (JS) | `src/payment/index.js:51` | `ipv6_enabled == "true"` (JS string value compare) | ✅ |
| shipping (Rust) | `src/shipping/src/main.rs:37` | `ipv6_enabled == "true"` (`String`/`&str` value compare) | ✅ |
| quote (PHP) | `src/quote/public/index.php:95` | `$ipv6_enabled == "true"` (PHP string value compare) | ✅ |

`compose.yaml` passes `IPV6_ENABLED` through to the currency container (it's forwarded at lines 193, 470, 567, 645), and `.env` defaults it to `IPV6_ENABLED=false` — so the variable is genuinely wired up end-to-end, it's just that the C++ side can never honor a `true` value.

## Impact

```mermaid
flowchart TD
    A["IPV6_ENABLED=true set in .env / compose"] --> B["Container env var passed to currency service"]
    B --> C["std::getenv(&quot;IPV6_ENABLED&quot;) returns const char*"]
    C --> D{"ipv6_enabled == &quot;true&quot;\n(pointer compare)"}
    D -->|"expected: true, value matches"| E["ip = '[::]'\n(bind dual-stack / IPv6)"]
    D -->|"actual: always false\n(different memory addresses)"| F["ip stays '0.0.0.0'\n(IPv4-only bind)"]
    F --> G["Service unreachable on\nIPv6-only clusters/networks\n(same symptom as #2482)"]
```

In an IPv6-only Kubernetes cluster or network (the exact scenario #2482 was filed for), setting `IPV6_ENABLED=true` will not fix the currency service — it will keep binding to `0.0.0.0` and remain unreachable, while payment, shipping, and quote correctly switch over. This is a partial regression of #2482 that's easy to miss because the env var is present and "looks" wired up correctly; it just never has any effect in this one service.

## Steps to reproduce

1. Set `IPV6_ENABLED=true` in `.env` (or the container environment for the `currency` service).
2. Start the currency service (`docker compose up currency` or run it locally).
3. Check what address it's actually listening on (e.g. `netstat`/`ss`, or attempt to reach it only over IPv6).
4. Observe it is still bound to `0.0.0.0`, and the `currency.server.ip_overwrite` log line is never emitted — compare against payment/shipping/quote with the same env var set, which do switch to `[::]` and do log the equivalent line.

## Expected behavior

When `IPV6_ENABLED=true`, the currency service should bind to `[::]`, matching the behavior of every other service that implements this same environment variable.

## Suggested fix

Compare the string's contents instead of the pointer, e.g.:

```cpp
if (ipv6_enabled != nullptr && std::strcmp(ipv6_enabled, "true") == 0) {
```

or convert to `std::string` first:

```cpp
if (ipv6_enabled != nullptr && std::string(ipv6_enabled) == "true") {
```

(the `nullptr` check also guards against `getenv` returning `nullptr` when the variable isn't set at all, which `ipv6_enabled == "true"` currently "handles" only by accident).

## Component(s)

currency

## Additional context

Related to / partial regression of #2482, which introduced this same `IPV6_ENABLED` pattern across services but didn't catch that the C++ implementation uses pointer comparison instead of string comparison.

bhuvan-somisetty:
# Changes

Fixes #3735

IPV6_ENABLED is supposed to switch the currency service from binding 0.0.0.0 to binding [::], same as payment, shipping and quote already do. The check here compared the const char* returned by std::getenv directly against a string literal, which compares pointers, not the actual string contents. That can never be true, so the branch was dead code no matter what the env var was set to.

Fixed by wrapping the value in std::string before comparing, and added a null check since getenv returns nullptr when the variable isn't set at all (the old code only "handled" that case by accident).

Checked the equivalent logic in payment/index.js, shipping/src/main.rs and quote/public/index.php to confirm the comparison now matches how the other three services do it.

## Merge Requirements

* [x] Documentation updates in the [docs][]: none needed, IPV6_ENABLED isn't called out anywhere in the docs today, this just makes the currency service match the behavior payment/shipping/quote already have
* [x] Helm chart updates in the [helm-charts][]: none needed, currency already has IPV6_ENABLED wired through in values.yaml, this only changes how the currency binary interprets the value it's given

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ the release.
 
 ## Unreleased
 
+* [currency] Fix `IPV6_ENABLED` check comparing a `const char*` pointer
+  against a string literal instead of the string's contents, so the check
+  was always false and the service could never bind to `[::]`
+  ([#3735](https://github.com/open-telemetry/opentelemetry-demo/issues/3735))
 * [kafka] Add `KAFKA_TOPIC` environment variable to configure the Kafka topic
   name used by `checkout`, `accounting`, and `fraud-detection`, defaulting to
   `orders` to preserve existing behavior
--- a/src/currency/src/server.cpp
+++ b/src/currency/src/server.cpp
@@ -254,8 +254,8 @@ void RunServer(uint16_t port)
   std::string ip("0.0.0.0");
 
   const char* ipv6_enabled = std::getenv("IPV6_ENABLED");
-  
-  if (ipv6_enabled == "true") {
+
+  if (ipv6_enabled != nullptr && std::string(ipv6_enabled) == "true") {
     ip = "[::]";
     logger->Info(eventName("currency.server.ip_overwrite"),
                  "Overwriting Localhost IP",


ps48:
Nice catch verified locally: g++ itself warns comparison with string literal results in unspecified behavior `[-Waddress]`, and with `IPV6_ENABLED=true` the old pointer check returns false while your `std::string` version correctly returns true (and the `nullptr` guard handles the unset case without crashing). Brings currency in line with how payment/shipping/quote already compare it.

ps48 (approved):

PR #3750, opened 2026-07-22
fix(payment): correct nanos handling for demo.payment.amount

### LINKED ISSUE #3747: [payment] demo.payment.amount telemetry attribute is wrong whenever nanos has fewer than 9 digits
# Bug Report

Commit 29d4806cd635f527b2db5855647b623f5ee31ced (main).

## Symptom

`chargeServiceHandler` in `src/payment/index.js` builds the `demo.payment.amount` span attribute by string-concatenating `units` and `nanos` from the `Money` proto directly, with no zero-padding:

```js
const amount = call.request.amount
span?.setAttributes({
  'demo.payment.amount': parseFloat(`${amount.units}.${amount.nanos}`).toFixed(2)
})
```

Per `pb/demo.proto`, `nanos` is the fractional part of the amount in billionths of a unit (e.g. `$-1.75` is represented as `units=-1, nanos=-750,000,000`), so it has to be treated as a 9-digit, zero-padded decimal fraction. This code instead just drops the raw integer straight after the decimal point, so whenever the printed form of `nanos` has fewer than 9 digits, the decimal point ends up in the wrong place.

`demo.payment.amount` should match the actual charge amount, the way the telemetry schema documents it (`telemetry-schema/attributes/payment.yaml` gives examples like `"29.99"`, `"99.95"`).

The attribute is inflated whenever the fractional part is smaller than 0.1 units, i.e. for most prices with a leading zero after the decimal point.

Concrete example: a charge of `units=19, nanos=5000000` represents `$19.005`. `nanos` printed as a plain integer is `"5000000"` (7 digits, not zero-padded to 9), so the code computes:

```
parseFloat("19.5000000").toFixed(2) === "19.50"
```

instead of the correct `"19.01"` (rounding `$19.005`). That's roughly a 50x error in the reported amount. The same thing happens for any `nanos` value under `100,000,000` (i.e. any price where the cents/sub-cent portion starts with one or more leading zeros), which is common for ordinary two-decimal prices like `$x.0y`.

For comparison, `src/payment/charge.js:96` logs `units`/`nanos` as separate fields and is not affected, this is specific to the span attribute built in `index.js`.

## Reproduce

1. Send a `Charge` request to the payment service with an amount whose `nanos` value has fewer than 9 digits when printed, e.g. `units: 19, nanos: 5000000` (`$19.005`).
2. Look at the `demo.payment.amount` attribute on the resulting span.
3. Compare it to the actual amount charged, it will be off by a power of 10 (`19.50` instead of `~19.01` in this example).

## Additional Context

Same class of bug as #3742 (mishandling the `Money.nanos` field when building telemetry attributes), but in a different service and a different call site, `nanos` needs to be zero-padded to 9 digits before being appended after the decimal point, not concatenated raw.

## Component(s)

payment, telemetry

bhuvan-somisetty:
# Changes

Fixes #3747

chargeServiceHandler in src/payment/index.js built the demo.payment.amount span attribute by string concatenating units and the raw nanos integer with a decimal point in between, then parsing that back to a float. Money.nanos is billionths of a unit and needs to be treated as a 9 digit fractional value, so concatenating it raw only works by coincidence when nanos happens to have 9 digits already. Whenever nanos prints with fewer digits, the decimal point ends up in the wrong place.

For example units=19, nanos=5000000 represents $19.005, but the old code produced parseFloat("19.5000000").toFixed(2), which is "19.50", off by close to 50x.

Replaced the string concatenation with units + nanos / 1000000000, the same approach already used for Money to decimal conversion in src/frontend/components/ProductPrice/ProductPrice.tsx, so both places handle nanos the same way. Added a changelog entry under Unreleased.

Tested by checking node -c on the file and running the numbers by hand for a few Money values (19/5000000, 29/990000000, 99/950000000, 5/0), all now produce the correct amount instead of the inflated one.

## Merge Requirements

* [ ] Appropriate documentation updates in the docs (not applicable, this only fixes the value the telemetry emits, the schema already documents the correct examples)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,13 @@ the release.
   ([#3751](https://github.com/open-telemetry/opentelemetry-demo/issues/3751))
 * [agentic] Move agent, chatbot and mcp to OTLP http exporter
   ([#3745](https://github.com/open-telemetry/opentelemetry-demo/pull/3745))
+* [payment] Fix `demo.payment.amount` telemetry attribute throwing a TypeError
+  at runtime. `Money.units` is an `int64` proto field decoded by
+  `@grpc/proto-loader` as a `Long` object, not a JS number. Adding a `Long` to
+  a float still returns a `Long`, which has no `.toFixed()` method. Wrap
+  `amount.units` in `Number()` before the arithmetic so `.toFixed(2)` works
+  correctly
+  ([#3747](https://github.com/open-telemetry/opentelemetry-demo/issues/3747))
 * [currency] Guard the `VERSION` environment variable lookup against `nullptr`:
   constructing a `std::string` directly from `std::getenv("VERSION")` crashes
   when the variable is unset, so fall back to `"unknown"` instead
--- a/src/payment/index.js
+++ b/src/payment/index.js
@@ -14,7 +14,7 @@ async function chargeServiceHandler(call, callback) {
   try {
     const amount = call.request.amount
     span?.setAttributes({
-      'demo.payment.amount': parseFloat(`${amount.units}.${amount.nanos}`).toFixed(2)
+      'demo.payment.amount': (Number(amount.units) + amount.nanos / 1000000000).toFixed(2)
     })
     logger.info("Charge request received.")
 


julianocosta89 (changes requested):
@bhuvan-somisetty this is not working.

I tested it locally and got this in Jaeger

Image

```
(amount.units + (amount.nanos / 1000000000)).toFixed is not a function
--
```

bhuvan-somisetty:
Fixed — amount.units is decoded as a Long object by @grpc/proto-loader (proto int64 field), not a plain JS number. Adding a Long to a float returns a Long, which has no .toFixed(). Wrapped amount.units in Number() so the arithmetic and .toFixed(2) call work correctly.

julianocosta89 (approved):

bhuvan-somisetty:
All checks are green, pls take a look whenever ur free

PR #3752, opened 2026-07-22
fix(react-native-app): correct nanos divisor in ProductCard price

### LINKED ISSUE #3751: [react-native-app] ProductCard shows wrong price, nanos divided by 1e8 instead of 1e9
# Bug Report

Commit 92df8a2783543d18e2446a924f2578c83d8a506c (main).

## Symptom

`ProductCard` in `src/react-native-app/components/ProductCard/ProductCard.tsx` converts a product's `Money` price to a display string like this:

```ts
const price = (priceUsd?.units + priceUsd?.nanos / 100000000).toFixed(2);
```

Per `src/react-native-app/protos/demo.ts`, `nanos` is documented as nano (10^-9) units of the amount, e.g. `$-1.75` is represented as `units=-1, nanos=-750,000,000`. Converting `nanos` back to a fractional dollar amount requires dividing by `1,000,000,000` (1e9), not `100,000,000` (1e8). The divisor here is off by a factor of 10.

The frontend web app does this correctly in `src/frontend/components/ProductPrice/ProductPrice.tsx` (`units + nanos / 1000000000`), the react-native-app's `ProductCard.tsx` even has a comment at the top saying it was "Copied with modification from src/frontend/components/ProductCard/ProductCard.tsx", the modification appears to be where the wrong divisor was introduced.

The price shown on the product card should match the actual `Money` value, e.g. a product priced at `units=24, nanos=990000000` ($24.99) should render as `$ 24.99`.

The fractional part of the price is inflated by 10x. For `units=24, nanos=990000000`:

```
24 + 990000000 / 100000000 = 24 + 9.9 = 33.90
```

so the card shows `$ 33.90` instead of `$ 24.99`.

## Reproduce

1. Run the react-native-app against the demo backend (or just read `ProductCard.tsx` line 49 alongside any `Money` value where `nanos` is non-zero).
2. Look at any product card's displayed price.
3. Compare to the same product's price shown in the web frontend, the react-native price will be inflated whenever the product's cents aren't `.00`.

## Additional Context

Same class of bug as #3742 and #3747 (mishandling the `Money.nanos` field), this time in a price shown directly to the user rather than in telemetry. `nanos` needs to be divided by `1_000_000_000`, not `100_000_000`.

## Component(s)

react-native-app

bhuvan-somisetty:
# Changes

Fixes #3751

ProductCard in `src/react-native-app/components/ProductCard/ProductCard.tsx` converted a product's `Money` price to a display string by dividing `priceUsd.nanos` by `100000000` (1e8) instead of `1000000000` (1e9). Per `demo.proto`, `Money.nanos` represents nano (10^-9) units of the amount, so dividing by 1e8 inflated the fractional portion of the price by 10x.

For example, a product priced at `units=24, nanos=990000000` ($24.99) was displayed as `$ 33.90` (24 + 9.90).

Corrected the divisor to `1000000000`, aligning it with how `ProductPrice.tsx` in `src/frontend/components/ProductPrice/ProductPrice.tsx` and `chargeServiceHandler` in `src/payment/index.js` convert `Money` values. Added a changelog entry under Unreleased.

Tested by checking the calculation against multiple `Money` values (`units=24, nanos=990000000` -> `$ 24.99`, `units=1, nanos=50000000` -> `$ 1.05`), ensuring the price matches what the frontend web app shows.

## Merge Requirements

* [ ] Appropriate documentation updates in the docs (not applicable, this only fixes the price display string in the mobile app)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ the release.
 
 ## Unreleased
 
+* [react-native-app] Fix `ProductCard` price calculation where `nanos` was
+  divided by `100_000_000` (1e8) instead of `1_000_000_000` (1e9), inflating
+  the displayed price
+  ([#3751](https://github.com/open-telemetry/opentelemetry-demo/issues/3751))
 * [agentic] Move agent, chatbot and mcp to OTLP http exporter
   ([#3745](https://github.com/open-telemetry/opentelemetry-demo/pull/3745))
 * [currency] Guard the `VERSION` environment variable lookup against `nullptr`:
--- a/src/flagd-ui/lib/flagd_ui/storage.ex
+++ b/src/flagd-ui/lib/flagd_ui/storage.ex
@@ -46,16 +46,16 @@ defmodule FlagdUi.Storage do
   end
 
   @impl true
-  def handle_cast({:replace, json_string}, _) do
-    new_state =
-      case Jason.decode(json_string) do
-        {:ok, parsed} -> parsed
-        {:error, _} -> %{}
-      end
+  def handle_cast({:replace, json_string}, state) do
+    case Jason.decode(json_string) do
+      {:ok, new_state} ->
+        write_state(json_string)
+        {:noreply, new_state}
 
-    write_state(json_string)
-
-    {:noreply, new_state}
+      {:error, _} ->
+        Logger.warning("Ignoring replace with invalid JSON")
+        {:noreply, state}
+    end
   end
 
   @impl true
--- a/src/flagd-ui/lib/flagd_ui/storage.ex
+++ b/src/flagd-ui/lib/flagd_ui/storage.ex
@@ -77,7 +77,12 @@ defmodule FlagdUi.Storage do
   end
 
   defp write_state(json_string) do
-    File.write!(@file_path, json_string)
+    # Write-then-rename so concurrent readers (e.g. flagd, or another Storage
+    # process in tests) never observe a truncated/empty file mid-write:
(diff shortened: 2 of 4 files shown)


julianocosta89 (approved):

julianocosta89 (approved):

julianocosta89 (approved):

julianocosta89:
@bhuvan-somisetty commit(2d525d78) changed `Storage.handle_cast({:replace, json_string}, _)` to gracefully handle invalid JSON instead of crashing — but it called `write_state(json_string)` unconditionally, even when `Jason.decode/1` failed.

The practical effect: if `Storage.replace` ever receives malformed JSON, the in-memory flag state silently resets to `%{}`, but the invalid JSON still gets written to `demo.flagd.json` — the same file flagd itself watches for flag definitions. So a bad `:replace` call could corrupt the live flag config on disk while flagd-ui shows nothing wrong.

Commit [7211548](https://github.com/open-telemetry/opentelemetry-demo/pull/3752/commits/7211548736d1544704553d3785e390bb99ff6731) reorders the logic so `write_state/1` only runs after a successful decode. On failure, it logs a warning and keeps the previous in-memory state, so invalid input is rejected rather than silently persisted. A regression test was added to `storage_test.exs` asserting that both the in-memory state and the on-disk file are unchanged after a replace with invalid JSON — verified it fails against the old code and passes against the fix.

bhuvan-somisetty:
Thanks for catching that! That makes total sense — validating Jason.decode before calling write_state/1 ensures invalid payloads are rejected rather than persisted to disk.

julianocosta89 (approved):

PR #3759, opened 2026-07-24
fix(payment): add error.type attribute to error spans

### LINKED ISSUE #3758: [paymentservice] error.type attribute missing on server spans during payment failures
## Description

When an error occurs during payment processing in the Payment Service (for example, when the `paymentFailure` feature flag is enabled or when invalid payment credentials are provided), the Payment Service correctly sets the span status to `ERROR`. However, it does not set the `error.type` attribute on the span.

According to the OpenTelemetry Semantic Conventions for Exceptions and Errors, whenever an operation fails or span status is `ERROR`, an `error.type` attribute SHOULD be recorded on the span to allow backend metrics exporters and observability platforms to aggregate and categorize errors cleanly.

---

## Service Interaction & Data Flow

```
┌─────────────────┐       gRPC ChargeRequest       ┌─────────────────┐
│                 │ ─────────────────────────────> │                 │
│ CheckoutService │                                │ PaymentService  │
│                 │ <───────────────────────────── │    (Node.js)    │
└─────────────────┘      gRPC Error / Exception    └─────────────────┘
                                                            │
                                                            ▼
                                                   Span Status: ERROR
                                                  [Missing error.type!]
```

---

## Current vs Expected Telemetry

| Attribute Name | Current Behavior | Expected Behavior |
| -------------- | ---------------- | ----------------- |
| `span.status_code` | `ERROR` | `ERROR` |
| `status.message` | `"Payment failed"` | `"Payment failed"` |
| `error.type` | *(Not Present)* | `"PaymentFailureError"` / `"InvalidArgument"` |
| `rpc.service` | `"oteldemo.PaymentService"` | `"oteldemo.PaymentService"` |
| `rpc.method` | `"Charge"` | `"Charge"` |

---

## Impact

Without the `error.type` attribute:
1. **Metrics Aggregation:** Error rate metrics (such as span metrics processors or OTel collector connector metrics) cannot group error counts by error classification.
2. **Dashboard Visualizations:** Grafana and Jaeger trace charts cannot filter traces by specific failure categories.
3. **Spec Alignment:** The service falls out of alignment with current OpenTelemetry Semantic Conventions.

---

## Steps to Reproduce

1. Start the OpenTelemetry Demo application stack.
2. Enable the `paymentFailure` feature flag in `flagd` (or set it to 100%).
3. Place an order through the store UI or send a `ChargeRequest` to the Payment Service.
4. Inspect the generated span for `oteldemo.PaymentService/Charge` in Jaeger / Grafana / OTLP exporter logs.
5. Observe that the span status is set to `ERROR`, but no `error.type` attribute is attached.

---

## Proposed Fix Plan

In the Payment Service charge handler (`src/paymentservice/charge.js`), when an error occurs during payment processing:
1. Set `error.type` on the active span using the error class name or code (e.g., `span.setAttribute('error.type', err.name || 'PaymentFailureError')`).
2. Ensure exception recording is properly associated with the span.

bhuvan-somisetty:
# Changes

Fixes #3758

The Payment Service correctly sets span status to `ERROR` when payment processing fails — whether from the `paymentFailure` feature flag, an invalid card number, an unsupported card type, or an expired card — but neither `charge.js` nor `index.js` set the `error.type` attribute on those error spans.

`error.type` is a [required attribute in the OpenTelemetry Semantic Conventions](https://opentelemetry.io/docs/specs/semconv/general/recording-errors/#recording-errors-on-spans) whenever `span.status_code` is `ERROR`. Without it, span metrics processors and collector connectors that derive error rate metrics from spans cannot group or filter by error category, and observability platforms like Grafana and Jaeger cannot differentiate between a payment declined due to an expired card, an unsupported card type, or the `paymentFailure` flag.

Added `span.setAttribute('error.type', err.name || 'Error')` in the catch block of `charge.js` (which runs with the `charge` child span) and the catch block of `index.js` (which runs with the gRPC server span). Using `err.name` gives the standard JavaScript error class name (`'Error'` for generic `new Error(...)` throws) and falls back to `'Error'` if not set, keeping the value consistent and non-null.

Tested by running `node --check` on both files and manually verifying that the attribute value matches the thrown error's `name` property for each of the four failure paths in `charge.js`.

## Merge Requirements

* [ ] Appropriate documentation updates in the docs (not applicable, this only adds a missing telemetry attribute on existing error spans)

### DIFF
--- a/src/payment/charge.js
+++ b/src/payment/charge.js
@@ -100,6 +100,7 @@ module.exports.charge = async request => {
   } catch (err) {
     span.recordException(err);
     span.setStatus({ code: SpanStatusCode.ERROR, message: err.message });
+    span.setAttribute('error.type', err.name || 'Error');
 
     throw err;
   } finally {
--- a/src/payment/charge.js
+++ b/src/payment/charge.js
@@ -1,6 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 const { context, propagation, trace, metrics, SpanStatusCode } = require('@opentelemetry/api');
+const { ATTR_ERROR_TYPE } = require('@opentelemetry/semantic-conventions');
 const cardValidator = require('simple-card-validator');
 const { v4: uuidv4 } = require('uuid');
 
@@ -100,7 +101,7 @@ module.exports.charge = async request => {
   } catch (err) {
     span.recordException(err);
     span.setStatus({ code: SpanStatusCode.ERROR, message: err.message });
-    span.setAttribute('error.type', err.name || 'Error');
+    span.setAttribute(ATTR_ERROR_TYPE, err.name || 'Error');
 
     throw err;
   } finally {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ the release.
 * [cart] Report health to the demo's OpAMP server when running with the
   observability stack.
   ([#3656](https://github.com/open-telemetry/opentelemetry-demo/pull/3656))
+* [payment] Add `error.type` attribute to error spans in `charge.js` and
+  `index.js` to align with OpenTelemetry Semantic Conventions for errors
+  ([#3758](https://github.com/open-telemetry/opentelemetry-demo/issues/3758))
 * [react-native-app] Render missing `City` and `State` input fields in `CheckoutForm`
   ([#3754](https://github.com/open-telemetry/opentelemetry-demo/issues/3754))
 * [react-native-app] Fix `ProductCard` price calculation where `nanos` was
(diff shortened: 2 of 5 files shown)


bhuvan-somisetty:
@julianocosta89 PTAL whenever u get chance.

src/payment/charge.js
```diff
@@ -100,6 +100,7 @@ module.exports.charge = async request => {
   } catch (err) {
     span.recordException(err);
     span.setStatus({ code: SpanStatusCode.ERROR, message: err.message });
+    span.setAttribute('error.type', err.name || 'Error');
```
julianocosta89: @bhuvan-somisetty would it make sense to use the `ATTR_ERROR_TYPE` from `@opentelemetry/semantic-conventions` here?

bhuvan-somisetty:
Good call, updated to use ATTR_ERROR_TYPE from @opentelemetry/semantic-conventions. Thanks for the suggestion!

osullivandonal (approved):

PR #3763, opened 2026-07-24
Add agentic loadgen and updated caching

fali007:
# Changes

1. Agentic loadgen: Added natural language requests to enable load generation for agent to existing k6 load generator.
2. LLM Request response caching : Existing vcr based approach stores a file for all requests. It had two issues.
- During load generation multiple requests get processed concurrently which causes the cache file to rewritten (in record mode) and eventually it will contain only last written request and response. (This is not an issue if recording is off)
- While searching for a similar request, sequence matcher will have to match all requests which is O(n**2).

These issues are resolved by a request - hash based filename for each request-responses. To check if a requests is available in the cache is a O(1) lookup. 
To match the similar requests, an index is created with number of messages in the requests to prune non matching requests faster.

Tries to resolve #3762 

## Merge Requirements

For new features contributions, please make sure you have completed the following
essential items:

* [ ] `CHANGELOG.md` updated to document new feature additions
* [ ] Appropriate documentation updates in the [docs][]
* [ ] Appropriate Helm chart updates in the [helm-charts][]

Maintainers will not merge until the above have been completed. If you're unsure
which docs need to be changed ping the
[@open-telemetry/demo-approvers](https://github.com/orgs/open-telemetry/teams/demo-approvers).

[docs]: https://opentelemetry.io/docs/demo/
[helm-charts]: https://github.com/open-telemetry/opentelemetry-helm-charts

PR #3766, opened 2026-07-24
Felix 3763 vcr fix

mwimpelberg28:
# Approach: per-request cache mode for synthetic traffic (PR #3763)

## Problem

The load generator fires a fixed, high-volume set of prompts at the agent
concurrently. Under `LLM_CACHE_MODE=hybrid` (the default), any prompt that
misses the cache falls through to a live LLM call. That makes load-test runs
slow, non-deterministic, and costly  and reintroduces the class of bug this
PR is fixing in the first place, just at the LLM-call layer instead of the
cache-file layer.

Forcing the whole service into `LLM_CACHE_MODE=replay` avoids that, but
breaks the other legitimate use case: a human typing a custom, uncached
question into the chatbot needs that request to fall through to a live call
and get recorded, so it works next time. `LLM_CACHE_MODE` is a single
process-wide setting, so it can't satisfy both needs at once when both kinds
of traffic hit the same `agent` instance.

## Approach

Make the cache mode a per-request decision instead of a process-wide one:

- Synthetic traffic (the load generator) is always forced into `replay`
  semantics, regardless of the configured `LLM_CACHE_MODE`: a cache hit
  replays normally, a miss raises `LLMCacheMiss` immediately rather than
  calling the live LLM.
- Organic traffic (e.g. the chatbot UI) is unaffected and keeps the
  process-wide `LLM_CACHE_MODE` as configured today, so custom questions
  still get recorded on the fly.

The two are distinguished using the `synthetic_request=true` baggage
attribute the load generator already sends on every request (`script.js`'s
`otelHeaders` helper) — the same attribute `checkout` and `payment` already
use to separate synthetic from real traffic, so no new attribute or header
was introduced.

## Implementation steps

- `agents.py`: `handle_prompt` reads `synthetic_request` off the OTel
  baggage of the incoming request (already populated by `FastAPIInstrumentor`
  from the `baggage` header) and threads it through `run_agent` into
  `ChatLLM(is_synthetic=...)`.
- `llm.py`: `ChatLLM` takes an `is_synthetic` flag. `_effective_mode()`
  returns `"replay"` for synthetic calls and the cache's configured mode
  otherwise; `_generate`/`_agenerate` use that effective mode instead of
  reading `cache.mode` directly.

Two files changed, no changes to `llm_cache.py`'s storage/matching logic.

## Verification

Built the agent image from this branch and ran it in Docker against the
existing `fixtures/llm_cache` fixtures:

- Synthetic request, cached prompt → `LLM cache exact hit`, replays.
- Synthetic request, novel prompt → blocked immediately (`LLMCacheMiss`), no
  live LLM call in the logs, cache directory unchanged (530 files before and
  after).
- Organic request (no baggage header), same novel prompt → reached the real
  LLM endpoint (401 with a dummy key), confirming record/replay is unchanged
  for custom questions.


fali007:
We are trying to simulate an agentic system as much as possible. Right now the LLM response from cache is really quick which is wrong imo (This would a really good feature to have). It should have some minimum latency proportional to the number of tokens because agentic workflows are expected to be slow.
I am okay with this approach of being in replay (read only) mode for loadgen requests. But my concern is shouldn't be optional (like replay, hybrid, record modes) because for someone who is trying to measure the latency of the workflow should be able to run it without any code changes. 

I am not an expert here.

I would like to hear what others has to say.
@julianocosta89 @ps48

julianocosta89:
Sorry @mwimpelberg28 but there is no way we can accept this PR.
This adds 937 files and 253k LoC.

PR #3783, opened 2026-07-28
Check response.ok in the shared request() fetch wrapper

### LINKED ISSUE #3782: Request() fetch wrapper doesn't check response.ok — HTTP errors with JSON bodies are treated as success
## Description

The shared `request()` fetch wrapper in `src/frontend/utils/Request.ts` (and the copy in `src/react-native-app/utils/Request.ts`) never checks `response.ok` before treating a response as successful. It just reads the response body and, if non-empty, `JSON.parse`s it and returns it as the result:

```ts
const responseText = await response.text();

if (!!responseText) return JSON.parse(responseText);

return undefined as unknown as T;
```

`fetch()` only rejects on network-level failures — it resolves normally for HTTP error statuses like 400/404/500. Because the status code is never inspected, any API route that responds with a non-2xx status and a JSON body (e.g. `{"message": "..."}`, `{"error": "..."}`, a NextAuth-style error payload, etc.) is parsed and handed back to the caller exactly as if the call had succeeded. The promise only rejects by accident, if the error body happens not to be valid JSON.

This affects every caller that goes through this wrapper — cart, checkout, products, recommendations, shipping — in both `src/frontend` and `src/react-native-app`.

## Current vs Expected Behavior

| Scenario | Current Behavior | Expected Behavior |
| :--- | :--- | :--- |
| API route returns 4xx/5xx with a JSON error body | Promise **resolves** with the error JSON treated as valid data | Promise **rejects** with a meaningful error (status + message) |
| API route returns 4xx/5xx with a non-JSON body | Promise rejects, but with an opaque `SyntaxError: Unexpected token ...` instead of the actual HTTP error | Promise rejects with a clear error identifying the failed request/status |
| API route returns 2xx | Works as expected | Works as expected |

## Why this matters

Any caller that adds error handling around one of these API calls (e.g. a `try/catch` around `placeOrder`) can only catch failures that happen to produce invalid JSON. A backend that fails "cleanly" with a JSON error body will be silently accepted as a success, so the UI proceeds down the success path (success toast, cart cleared, navigation away, etc.) even though the request failed server-side.

## Steps to Reproduce

1. Point any endpoint behind `request()` (e.g. `/api/cart`) at a handler that returns a non-2xx status with a JSON body, for example `res.status(500).json({ message: 'boom' })`.
2. Call the corresponding gateway method (e.g. `ApiGateway.getCart` / `addCartItem`) from the frontend or react-native-app.
3. Observe that the call resolves successfully with `{ message: 'boom' }` as the "data" instead of rejecting.

## Proposed Fix

Check `response.ok` in `request()` and throw before attempting to use the body as a success payload, e.g.:

```ts
if (!response.ok) {
  const text = await response.text();
  throw new Error(`Request to ${url} failed with status ${response.status}: ${text}`);
}
```

This should be applied to both `src/frontend/utils/Request.ts` and `src/react-native-app/utils/Request.ts` since they're kept in sync.

## Component(s)

frontend, react-native-app

bhuvan-somisetty:
# Changes

The shared `request()` fetch wrapper in `src/frontend/utils/Request.ts` (and
the copy in `src/react-native-app/utils/Request.ts`) never checked
`response.ok` before treating a response as successful. `fetch()` only
rejects on network-level failures, it resolves normally for HTTP error
statuses like 400/500. Since the status code was never inspected, any API
route that responded with a non-2xx status and a JSON body (e.g.
`{"message": "..."}`) was parsed and handed back to the caller exactly as if
the call had succeeded.

This meant callers such as `placeOrder` could only catch failures that
happened to produce invalid JSON. A backend that failed "cleanly" with a
JSON error body was silently accepted as a success, so the UI would proceed
down the success path (success toast, cart cleared, navigation away, etc.)
even though the request failed server-side.

This PR checks `response.ok` in `request()` in both `src/frontend` and
`src/react-native-app` and throws a descriptive error (including the URL,
status, and response body) before the body is parsed as a success payload.

Fixes #3782

## Merge Requirements

essential items:


osullivandonal:
I'm not sure this change adds value to the demo, the demo is used to demonstrate how to instrument with OTel.

A failing backend call is already captured:
- Client-side fetch is auto-instrumented by FrontendTracer, so the span already records http.status_code and the error regardless of this check.
- The server-side InstrumentationMiddleware.ts already records exceptions and sets span status to ERROR.

The demo deliberately injects failures via feature flags (adFailure, cartFailure, productCatalogFailure, etc.). Can you show that the behaviour of these feature flags has not changed and that the telemetry gathered is still the same? Please see [CONTRIBUTING.md](https://github.com/open-telemetry/opentelemetry-demo/blob/514e37e18976db525ab7f7a47ca9e92d67860a93/CONTRIBUTING.md?plain=1#L295)

bhuvan-somisetty:
Good point, and thanks for digging into the instrumentation side of it. You're right that the fetch auto-instrumentation records http.status_code on the span regardless of what the app code does with the response afterward, so this change shouldn't affect what gets exported it's purely about the app layer not silently treating a failed request as a success.

I don't have Docker available in my current setup, so I can't do the before/after trace comparison with the feature flags right this second, but I'll spin it up separately and test with adFailure, cartFailure, and productCatalogFailure and report back what I see in Jaeger.

One thing I noticed while looking into this again: CartDetail.tsx calls await placeOrder(...) with no try/catch, so with this fix a failed checkout now throws an unhandled rejection instead of silently continuing to a broken order page there's no user-facing error message either way. Happy to scope that as a follow-up if you think it's worth fixing, or fold it into this PR if you'd rather keep it together.

cijothomas (approved):

cijothomas:
Oops, wrong PR I approved. :( Plz ignore.

PR #3788, opened 2026-08-03
[Draft] Demo llm for astronomy shop application.

fali007:
# Changes

Add demo LLM, a tiny 5.2M parameter decoder only model which can be retrained and inference within a laptop. `demo-llm` uses 1500MB of RAM while serving. It is built using a custom tokenizer and only understands Otel-demo terminology. 
Now `agent` will call the `demo-LLM` for instead of relying cached responses or on a paid LLM api-keys. Users have the option to configure a premium model by configuring the credentials in `.env`.
This completely removes the requirement of cached files will running the natural language load generation. That PR was rejected because of large number of cache files.
Another benefit of the LLM is we will get metrics like token-throughput, latency, inter-token latency, pre-fill and decode time etc. which are important metrics which llm inference engines emit. Alerts on top of those can be a good addition to the gen-ai dashboard.

## Merge Requirements

essential items:


fali007:
After receiving feedback from @julianocosta89 and @ps48 on the resource usage I performed a load test to understand the resources.

```
profile  : 15 -> 60 RPM over 360s
prompts  : 143 from artifacts/raw_data/prompts.json
sampling : agent, demo-llm, mcp, chatbot

  t=   74s  rate= 24.3/min  sent=  25  done=  23
  t=  129s  rate= 31.1/min  sent=  50  done=  43
  t=  173s  rate= 36.7/min  sent=  75  done=  67
  t=  212s  rate= 41.5/min  sent= 100  done=  94
  t=  246s  rate= 45.8/min  sent= 125  done= 120
  t=  278s  rate= 49.7/min  sent= 150  done= 142
  t=  307s  rate= 53.4/min  sent= 175  done= 167
  t=  334s  rate= 56.8/min  sent= 200  done= 192
  sent 224, draining (up to 120s)

RESOURCE USAGE
service      baseline    mean    peak  peak%  cpu mean  cpu peak
----------------------------------------------------------------
agent            178M    209M    235M    78%        2%       18%
demo-llm         293M    449M    504M    49%      663%      843%
mcp              130M    130M    130M    65%        0%        0%
chatbot          140M    140M    140M    70%        3%        4%

REQUESTS  224 completed | ok 148 (66%) | failed 76
latency (ok only)  p50 8.1s  p90 34.6s  p95 49.6s  max 89.0s

rate rpm        n   ok  fail   fail%   p50 lat   p95 lat
--------------------------------------------------------
15-19         12   12     0      0%      8.9s     42.4s
20-24         15   15     0      0%      4.5s     11.3s
25-29         18   15     3     17%     11.2s     68.6s
30-34         21   12     9     43%     17.7s     81.8s
35-39         25   16     9     36%      8.5s     34.6s
40-44         28   21     7     25%      7.1s     14.5s
45-49         32   24     8     25%      8.1s     35.4s
50-54         35   23    12     34%      7.3s     57.4s
55-59         38   10    28     74%     12.6s     48.3s
```

From previous experiments I've found that the agent spawns more than 50+ requests. Ive added a `limit_concurrency=8` to `agent`. Now if more than 8 are active, `agent` will send a `503` error instead of sending the request to `demo-llm`.

This ties the resource usage of `demo-llm` below 600MB where it can serve upto 20 RPM. Ive also reduced the resources for `agent`, `mcp` and `chatbot` to 300, 200 and 200 respectively. They are also holding up okay in this experiment.

There is a computational bug in the `demo-llm`. It was written as a minimal server and doesn't have KV cache to store the intermediate attention scores. I think it will be minimal changes and can reduce the compute time a lot. That should improve the latency and hence throughput.

fali007:
After adding KV cache to the model there is an improvement in throughput.

```
profile  : 15 -> 60 RPM over 360s
prompts  : 143 from artifacts/raw_data/prompts.json
sampling : agent, demo-llm, mcp, chatbot

  t=   74s  rate= 24.3/min  sent=  25  done=  23
  t=  129s  rate= 31.1/min  sent=  50  done=  48
  t=  173s  rate= 36.7/min  sent=  75  done=  72
  t=  212s  rate= 41.5/min  sent= 100  done=  94
  t=  246s  rate= 45.8/min  sent= 125  done= 120
  t=  278s  rate= 49.7/min  sent= 150  done= 146
  t=  307s  rate= 53.4/min  sent= 175  done= 171
  t=  334s  rate= 56.8/min  sent= 200  done= 192
  sent 224, draining (up to 120s)

RESOURCE USAGE
service      baseline    mean    peak  peak%  cpu mean  cpu peak
----------------------------------------------------------------
agent            176M    202M    234M    78%        3%       23%
demo-llm         265M    438M    710M    69%      362%      619%
mcp              130M    130M    130M    65%        0%        1%
chatbot          140M    140M    140M    70%        3%        4%

REQUESTS  224 completed | ok 214 (96%) | failed 10
latency (ok only)  p50 3.5s  p90 8.9s  p95 12.9s  max 33.1s

rate rpm        n   ok  fail   fail%   p50 lat   p95 lat
--------------------------------------------------------
15-19         12   12     0      0%      2.5s      4.5s
20-24         15   15     0      0%      1.5s      3.6s
25-29         18   18     0      0%      2.4s      9.6s
30-34         21   21     0      0%      2.5s     12.9s
35-39         25   25     0      0%      2.2s      6.0s
40-44         28   27     1      4%      5.3s     14.3s
45-49         32   29     3      9%      5.5s     15.3s
50-54         35   34     1      3%      4.3s     12.8s
55-59         38   33     5     13%      5.0s     15.2s
```

CPU usage came down as expected and throughput doubled. I will run more tests to with less CPU's to get the numbers in a less powerful machine. I will also run experiment with a 500MB `demo-llm` to check its throughput.

fali007:
# Building demo LLM

## Index

1. Data
2. Tokenisation
3. Model
4. Training
5. Testing
6. Serving

### Data

Data used to build this is from previous LLM cache.
I am using data from the caching which contains sequence of LLM interactions of astronomy shop agent (LLMs are gpt5.5 and opus4.7).
Using this data to generate a training and validation split.

**Cleaning and pre-processing of data**

**Structual tokens** : `["<|user|>", "<|assistant|>", "<|call|>", "<|result|>", "<|end|>", "<|uid|>", "<|tools|>"]` and `<|sys:{}|>` to indicate the unique system token.

The unique ids in the tool calls are replaced with `uid` since those informations doesnt carry any useful information in learning.
`user`, `assistant` indicates the the incoming text from interacting user and LLM generated response in a previous interaction respectively.
`tools` tag indicates the list of tools at our disposal for LLM to choose from.
`call` indicates a tool call requested by LLM and `result` indicates the response generated by the tool for a `call`.
`end` tags the generation is stopped.

`SYSTEM` - is deliberately dropped as it was invariant in the dataset. A better model can be trained with varying system prompt as well. That requires a much larger dataset and is out of scope for this project. Replaced with `<|sys:{}|>` -> `<|sys:0|>`

Each turn of agent process the same prefix text repeatedlty. We want to avoid the keep the frequency of the data to 1. So sorting and taking the longest sequence is what i do to get the longest sequence of a flow.

Anything not a prefix of something longer is kept. If two runs of same prompt diverges at kth run, neither is a prefix of other.

Hallucination check: removes the trajectories which tries to call an unavailable tool.

Train and validation split is done on the unique user prompts rather than selecting random trajectories. This is done to prevent model from seeing even one trajectory in validation that is also in train. (Our data contains trajectories of same user prompt for both `opus4.7` and `gpt5.5`)

**Data Augmentation**
`shipping address, currency, quantity` are some free variables. These free variables are copied and ive created many copies of similar data points. This is done consistently. For example a shopping that starts in India will remain Indian throughout the trajectory. If it was not consistent, inconsistent swap would teach the model that these fields are noise. Product identity is not permuted by default because it might change the context of the question. (Telescope to book: context chages with product ID)

### Tokenisation

I chose a domian specific tokeniser here because astronomy shop vocabulary is minimal compared to an english vocabulary of a general open source tokeniser vocabulary. That low-entropy domain is the precondition that makes a custom tokenizer pay off. A general tokenizer makes a tiny LLM (1-10M) model arithmetically impossible.

GPT 2 tokenizer is 50k token vocabulary. At `d_model = 256`, embedding table itself would be `50k * 256` ~13M which is not feasible for me.

An embedding row is updated only when its token is seen in training data. A token which is randomly initialised stays as it is. In a generation step they add noise which causes hallucination. With the custom vocabulary we are training 82% of the embedding table.

1. Emebedding table is a dictionary: It says what token `i` means in isolation.
2. Layers of transformer is computation: Attention lets the tokens read and MLP layers transform it into next token where it predicts next token conditioned on this context which attention gives us.
3. Increasing the vocabulary makes the embedding table taller.
Depth is what lets each layer build on the previous layer's abstractions.
This is why `3 layers × 128` **beats** `10 layers × 19` and why it beats `1 layer × 220`.

An embedding row receives a gradient only on steps where its token actually appears. If a token never occurs in training, its row's gradient is exactly zero, forever — it stays at random initialisation.
Counting updates per row per epoch:

If embeddings are not well trained

1. Reading: the vector is near-random, so the model can't tell what it just read.
2. Writing: the logit for token `i` is hidden · `row_i`. Dot with a random vector and you get an arbitrary logit. Might lead to highly hallucinated output.

So the custom tokeniser boils down to better compression means fewer tokens (`291K vs 374K`), which means each remaining token carries more meaning and gets more updates.

```bash
text ──▶ ① atoms ──▶ ② regex pretokenize ──▶ ③ BPE ──▶ ids
```

**Atoms** : 28 strings that are not supposed to split. Structural markers, System prompt, tool names (`conditioning.json`) and product Ids which are seen more than 5x. Advantage is `66VCHSJNUP` will have just 1 token now vs 10 tokens else. We omit splitting of these strings in `pretokenize` step.

`'(?:s|t|re|ve|m|ll|d)| ?[A-Za-z]+| ?[0-9]{1,3}| ?[^\sA-Za-z0-9]+|\s+` : `Byte pair encoding - BPE`

Digits are capped at 3 so prices and quantities share token instead of every number becoming unique. We only see english literature so `[A-Za-z]`. A leading space will be present in words which are in the inside of a sentance. A starting word and a in-sentance word have different tokens.

Bytes are repeatedly merged if they are very frequent. Json specific syntaxes etc are merged here. `NationalParkFoundationExplorascope`, `EclipsmartTravelRefractorTelescope` end up being single token. Training picks the most frequent pairs to merge. Rank of BPE is the priority score on the stage the merging of tokens happened. It creates an ordered list of merges. Lower rank -> high frequency -> earlier it was picked for merging -> high priority. This gives determinism while encoding.

`<|call|> | add_to_cart | {" | productId | ":" | 66VCHSJNUP | "," | quantity | ": | 2 | } | <|result|>` These is an example from our dataset. GPT's tokeniser would have produced more than 30 tokens. This helps to achieve the full trajectory part of the single context (2048 *might increase in the future).

**Think** : How SVD works here in formulating the manifold dimension and how a smaller tokeniser vocabulary can improve the participation ratio/effective rank.

1. How many directions dominate?" - This is the participation ratio. ~22 for our data.
2. How many directions are in use at all? - This is the effective rank and its 153 for our data.

So `256d` is a defensive pick.

### Model

**Model size calculation**

```bash
idx ──▶ wte ──▶ [Block × 6] ──▶ ln_f ──▶ lm_head ──▶ logits
                   │
                   └── x + attn(ln_1(x))  then  x + mlp(ln_2(x))
```

1. `idx` - Input token Indices
2. `wte` - Word Token Embeddings (256)
3. `Block` - The 6 Attention layers.
4. `ln_1` and `ln_2` are layer norm. We also have residual connection which is the branch.
5. `mlp` - Multi-layer perceptron.
7. `ln_f` - Final layer norm which gives logits. A softmax gives final token probs.

Most of the params are part of `blocks`. Each block contains below number of params. We have 6 such blocks.

```bash
(Q, K, V, O) => 4 * (256 * 256)
MLP = 256 * 1024
```

Vocabulary Size - 2048
Embedding Dim - 256
Context Length - 2048 (This is fixed because this helps in packing 90% of the trajectories we have within the context)
Layers = 6
Attention Heads = 8

Token Embeddings = vocabulary size * Embedding Dim = 2048 * 256
Positional Embedding = 0 (RoPE instead of learned positions)
MLP = Layers * 12 * (Embedding Dim ** 2)

Total ~ 5M params

RoPE - Rotary Positional Embeddings. Instead of adding the sinuusiodal PE for each tokens, we relapce the positional embeddings altogether by rotating each token correspoding to its token. This preserves the relative token information as well.

`d_model % n_head == 0` and `hs` head size = (`Embedding Dim // Attention Heads`) should be even (Cannot do a single vector... it should be pairs) so RoPE has whole pairs to rotate.

The division in attention matters because the variance will be high which can saturate softmax and vanish gradients.

How much should query token $i$ pay attention to key token $j$ in head $h$ for batch item $b$? The following softmax and multiplication with $V$ gives a single $32$-dimensional vector containing the blended information from all context tokens.

Following MLP layers have `256→1024 → GELU → 1024→256` params. They transform the attention values to aid next token prediction. These neurons contains the fatual informations at different levels.

How to we validate the network? Initial loss should be ln(2048) ~ 7.6, All logits are equally likely.

### Training

Next token prediction on the full dataset.
Masked dataset to compute loss only on predicted data can be done.
We already have golden outputs which omits the requirement of DPO or other RLHF.

`B` - Batch size for training. (8)

### Serving

VLLM serving is not required here. A pytorch based llm serving instance will be more than enough.
OpenAI compatible server is what would be best here. Added KV cache to improve performance (Reduced CPU usage a lot)

## Evaluation

Model is not emitting jargon which was surprising result as it is.

**Shortcomings**

1. Large contexts threw it off. (Whole catalogue always does)
2. Loses context for a follow up question.
3. There are few tools which has a higher error rate than some others.
4. Not able to comprehend numbers. (Addition and Multiplication is all chaos)

### Testing locally

Run the build using `make build-agentic`
Run the demo using `make start-agentic`


### Performance Evaluation

500 MB of RAM is enough for `demo-llm` when the request rate is really low (5 RPM). I ran the agent from 15 RPM to 60 RPM. Stable till 20 RPM. `agent`, `mcp` and `chatbot` ran with 300, 200, 200 MB of RAM respsecively and `demo-llm` ran with 1024MB. For more than 20 RPM, I could find requests failing. The memory usage of LLM grow in proportion to the active requests.

To improve the throughput I should add a KV cache (avoids repeated computation) in demo-llm.

julianocosta89:
hey @fali007, this looks really cool!
I played around a bit and didn't feel it was too much of a pain for my laptop.
I saw that we are reducing about 800Mi from the other services, so that in reality is just adding a rough 300Mi to the whole demo, not bad.

The fact that I can actually interact with the chat without paying for it is super cool!
I tested a bit and I was impressed at first, but then I think things got a bit messy 😅 

On the good things. Without looking at the suggested prompts I passed my session ID and asked chatbot to add a lens cleaning kit to my cart (no product ID passed).
It simply worked!
I was with frontend opened on a second tab, and I saw the product popping up!

Regarding instrumentation of `demo-llm`, I'm still not 100% sure, I'll ask around and try to get more opinions.

Now the things we need to fix.

Sometimes I asked questions and  it simply returned a blank text baloon.
Then, I tried using a predefined prompt and the bot got crazy, check this out:

Screenshot 2026-08-13 at 10 07 39

IDK if we can simply configure it to return an error or something, or if this would also be a good case to show misbehaviours (but for that I'd rather have a feature flag to control on/off)

fali007:
Thankyou @julianocosta89 for trying it out and sharing the feedback.

I am aware of some of the shortcomings (Especially when list products tool is invoked). I think we can resolve some of them by increasing the context length and training with more data. Follow up questions generate empty response (empty chat bubble) because model is not trained with such examples.

>Regarding instrumentation of demo-llm, I'm still not 100% sure, I'll ask around and try to get more opinions.

I will wait for the feedback.

julianocosta89:
> Follow up questions generate empty response (empty chat bubble) because model is not trained with such examples.

Ah ha, that makes sense then!
Maybe we could make that clear to the user

julianocosta89:
We have discussed that on slack: https://cloud-native.slack.com/archives/C03B4CWV4DA/p1786615281407359

I'd vote to not move on with this PR, given that this will add maintainability burden to the Demo.

Issue #2375, opened 2025-07-21
Scheduled Start&Stop of scenarios


svrnm:
# Feature Request

To run the demo environment for regular presentations I need to have some of the scenario feature flags to be turned on for a while and then turned of for a while again

**Is your feature request related to a problem?**

Imagine I want to have an ongoing issue within every hour of the day, such that my backend solution can show the issue appearing at a certain "recent" point in time. RIght now I have to use external tooling to regularly turn a flag on and off

**Describe the solution you'd like:**

I think it would be great to have a "scheduler" in the feature flag UI, that allows me to define

- an interval in which the flag should be toggled
- a min & max duration for how long the flag should be toggled, a random value in between will be picked
- optional: selecting which feature flags should be toggled/which shouldn't be toggled
- optional: a seed to allow reproducible patterns

**Describe alternatives you've considered.**

I created some python scripts that can do that for me automatically:

https://gist.github.com/svrnm/1ccec8dd126a0b804e421ee385f6ed58

They work as expected, but I was wondering if this is something the demo could provide ootb?

abhijeets25012-tech:
I’d like to work on this feature.

My current plan is to introduce a lightweight scheduler service that periodically toggles feature flags through flagd using a configurable YAML file.

Initial scope:

* configurable intervals
* random min/max durations
* docker compose integration
* documentation/examples

Future enhancements could include UI scheduling support and reproducible seeded schedules.

puckpuck:
My desire would be to bring a UI component into this a fast follow, or phase 2 for, and not be delegated as _future enhancement_

PR #3810, opened 2026-08-05
feat(flagd-ui): add a scheduler to turn failure scenarios on and off automatically

mwimpelberg28:
# Changes

Adds a Scheduler tab to flagd-ui that periodically activates randomly picked
feature flags for a random duration, then turns them back off, so failure
scenarios appear and resolve on their own while the demo is left running.

This implements #2375 inside the existing Elixir app rather than as a new
service: the scheduler is a GenServer alongside the existing Storage process,
plus a LiveView page for configuration.

Configurable:

* Interval between activations
* Minimum and maximum hold duration, with a random value in between picked per
  activation
* How many distinct flags run at once, each with its own duration and offset
* Which flag variants may be picked, so `cartFailure` can be scheduled at 10%
  and 25% while 100% is left out
* An optional seed for reproducible patterns, reported back when not supplied

A flag is offered for scheduling when it has an `off` variant to return to and
at least one other variant to switch to. That leaves out `loadGeneratorVUs`,
which is a tuning knob rather than a failure scenario, and
`loadGeneratorTraffic`, whose resting state is `on`.

`Storage` now broadcasts the configuration after every write, so the Basic tab
reflects the changes the scheduler makes.

## Merge Requirements


src/flagd-ui/lib/flagd_ui_web/live/scheduler.ex
```diff
@@ -0,0 +1,461 @@
...
+                    None
+                  </button>
+                </div>
+              </div>
+
+              <p :if={@available == []} class="text-sm text-warning">
+                No schedulable flags found in the current configuration.
+              </p>
+
+              <div class="max-h-96 overflow-y-auto">
```
ps48: The flag list is in a `max-h-96 overflow-y-auto` container. On a normal ~900px viewport that shows about 8 of the 13 flags, and the rest (`kafkaQueueProblems`, `paymentFailure`, `paymentUnreachable`, `productCatalogFailure`, `recommendationCacheFailure`) are clipped below the fold with no visible scrollbar cue so it reads as "only some flags are here." Could we drop or raise the max-height, or add a visible scroll affordance / "scroll for more" hint? Small thing but it caused a double-take.

src/flagd-ui/README.md
```diff
@@ -20,6 +20,41 @@ The application can be run with the rest of the demo using the documented
...
+  random value in between is picked for each activation. The maximum has to fit
+  within the interval.
+* **Flags at once**: how many distinct flags may be activated per interval. A
+  flag is never picked twice in the same interval, and asking for more than the
+  number of selected flags simply activates all of them.
+* **Flags**: which flag variants may be picked, chosen one variant at a time. A
+  flag with graded variants can take part with only some of them, so
+  `cartFailure` can be scheduled at `10%` and `25%` while `100%` is left out.
+  A flag is offered when it has an `off` variant to return to and at least one
+  other variant to switch to. That leaves out `loadGeneratorVUs`, which is a
```
ps48: Just double-checking the exclusions: `loadGeneratorTraffic` (rests at `on`) and `loadGeneratorVUs` (no `off` variant) are intentionally left out, which matches the code and tests. Should we also add these in the scheduler too and reset them to defaults after the timer goes off?

ps48 (commented):
Really nice feature ran it locally against the full stack and the scheduler works well. Toggle All/None, per-flag "toggle all", presets, and Start/Stop all function, and telemetry flows end-to-end (flagd-ui traces in Jaeger, metrics in Prometheus, logs in OpenSearch). Tests are thorough (48 passing). A couple of UX notes below, one blocking-ish. Also the usual: CHANGELOG is in, docs are maintainer calls.

mwimpelberg28:
@ps48 Ive addressed the comments.  Mind having another look please?

ps48 (approved):

ps48:
Thanks for the contribution! @mwimpelberg28

PR #3817, opened 2026-08-09
fix(flagd): restore productCatalogFailure so the flag can be enabled

### LINKED ISSUE #3816: productCatalogFailure feature flag has been a no-op since #3425
## What happens

`productCatalogFailure` cannot be turned on. Enabling it in flagd-ui, or setting `defaultVariant` to `on` in `src/flagd/demo.flagd.json`, produces no errors in product-catalog and no failing traces.

## Why

The flag's targeting rule returns `"off"` in both branches:

```json
"targeting": {
  "if": [
    { "==": [{ "var": "product_id" }, "OLJCESPC7Z"] },
    "off",
    "off"
  ]
}
```

flagd resolves `targeting` before falling back to `defaultVariant`, and this rule always resolves to a concrete variant. So `defaultVariant` is never consulted and the flag evaluates to `false` for every product, whatever the UI shows.

## Where it came from

#3305 added the rule correctly, with `"on"` in the true branch. #3425 ("fix/disable ff by default") changed that branch to `"off"`:

```diff
@@ -37,7 +37,7 @@
               "OLJCESPC7Z"
             ]
           },
-          "on",
+          "off",
           "off"
         ]
```

That PR's goal was to stop flags being enabled by default. For every other flag, flipping `defaultVariant` achieved it. `productCatalogFailure` is the only flag in the file with a targeting rule, and editing the rule instead of the default disabled it permanently rather than by default.

## Suggested fix

Return `null` from the true branch so the rule scopes the failure to the one product but leaves the on/off decision to `defaultVariant`:

```json
"targeting": {
  "if": [
    { "==": [{ "var": "product_id" }, "OLJCESPC7Z"] },
    null,
    "off"
  ]
}
```

flagd falls back to `defaultVariant` when a targeting rule resolves to null, so this keeps the flag off by default (#3425's intent) while making the flagd-ui toggle work again.

## Impact

This is one of the documented failure scenarios and it is exposed in flagd-ui, so anyone using the demo to produce a reproducible product-catalog incident — for testing, for demos, or as a fixture for trace-analysis work — currently gets a silently green system.

I ran into it while using the demo's feature flags as fault injection for trace-based root-cause analysis. Happy to open the PR.

Priyanshu-byte-coder:
Fixes #3816

## Problem

`productCatalogFailure` cannot be turned on. Its targeting rule returns `"off"` from both branches, and flagd resolves `targeting` before falling back to `defaultVariant`, so the flag evaluates to `false` for every product regardless of the flagd-ui toggle.

#3305 added the rule with `"on"` in the true branch. #3425, which disabled feature flags by default, edited that branch to `"off"` rather than the `defaultVariant`:

```diff
@@ -37,7 +37,7 @@
               "OLJCESPC7Z"
             ]
           },
-          "on",
+          "off",
           "off"
         ]
```

That disabled the flag permanently instead of by default. It is the only flag in `demo.flagd.json` with a targeting rule, so no other flag was affected by the same edit.

## Fix

Return `null` from the true branch. flagd falls back to `defaultVariant` when a targeting rule resolves to null, so the rule keeps scoping the failure to `OLJCESPC7Z` while the on/off decision returns to `defaultVariant` — the flag stays off by default, which was #3425's intent, and the toggle works again.

```diff
           },
-          "off",
+          null,
           "off"
         ]
```

## Verification

Against the running demo (`compose.yaml` + `compose.observability.yaml`), restarting `flagd` between each step:

| targeting | defaultVariant | request | result |
|---|---|---|---|
| fixed | `off` | `GET /api/products/OLJCESPC7Z` | `200` |
| fixed | `on` | `GET /api/products/OLJCESPC7Z` | `500` |
| fixed | `on` | `GET /api/products/66VCHSJNUP` | `200` |

Off by default, fails only for the targeted product when enabled, other products unaffected.

Confirmed the failing request produces an error span on `oteldemo.ProductCatalogService/GetProduct` in the trace, so the scenario is usable as fault injection again.


Priyanshu-byte-coder:
Disclosure per the [GenAI contribution policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md): I used an AI assistant while working on this, for the code and for drafting this PR and issue #3816. I should have said so when I opened them.

What that does and does not mean here. The behaviour is verified by hand against a running demo, not asserted: with the targeting fix and `defaultVariant: off`, `GET /api/products/OLJCESPC7Z` returns 200; flipping `defaultVariant` to `on` returns 500; and `66VCHSJNUP` stays 200 either way, so the rule still scopes the failure to the one product. The regression was traced to the specific diff in #3425 that changed the true branch from `on` to `off`.

I can answer questions about any of it without assistance — in particular why the fix returns `null` rather than `on`: flagd resolves `targeting` before `defaultVariant`, so a rule that always resolves to a concrete variant makes the flagd-ui toggle inert. Returning `null` from the matching branch falls through to `defaultVariant`, which keeps the flag off by default (#3425's intent) while making it switchable again.

Happy to rework or drop it if the disclosure changes how you'd like to handle the PR.

julianocosta89 (approved):

julianocosta89 (approved):

julianocosta89:
@Priyanshu-byte-coder and @jason59000 I believe both PRs are not fixing anything.
PR #3898 changed the flag to `on` and `off`, which was exactly the behaviour I've changed on #3495.
This PR changes it to `null` and `off`, and it seems to work at first, but as soon as you switch the FF `on`, and after some time turns it back off, the value goes to `off` and `off`.

So I'd say that the current state is correct.
To test my theory I ran the Demo with `main`.
`main` has currently `off` and `off`. When I start the Demo the FF is `off` (as expected):
image

When I turn the FF `on` via flagd-ui (http://localhost:8080/feature), the value on flagd.json switches to `on`.
image


And I start getting errors:
image


I don't think we have an issue, and the Demo is working as expected.
Issue https://github.com/open-telemetry/opentelemetry-demo/issues/3816 was wrongly created, and not tested.

Priyanshu-byte-coder:
You're right, and I can point at the exact mechanism now — it makes your test result and my claim consistent, with mine being the wrong one.

`src/flagd-ui/lib/flagd_ui/storage.ex` decides where a toggle is written:

```elixir
case get_in(data, ["targeting", "if"]) do
  [_cond, _on, _off] ->
    put_in(data, ["targeting", "if", Access.at(1)], value)
  _ ->
    Map.replace(data, "defaultVariant", value)
end
```

When a flag has a three-element `targeting.if`, flagd-ui writes the toggled variant into **index 1 — the true branch** — and never touches `defaultVariant`. `productCatalogFailure` matches that shape, so toggling it on rewrites `[cond, "off", "off"]` to `[cond, "on", "off"]`, and the failure fires for `OLJCESPC7Z`. That is exactly what your screenshots show. The flag is not a no-op through the UI, which is the demo's actual toggle path, so #3816 was wrong.

It also explains why my change was worse than no change. `[cond, null, "off"]` is still three elements, so the same branch matches and the UI overwrites index 1 on the first toggle. On → off then leaves `["off", "off"]` — the sequence you described. The `null` cannot survive a single use of the UI, so it fixes nothing even on its own terms.

What I actually exercised was setting `defaultVariant` by hand in the file, where the targeting rule does win and the flag stays off. I generalised from that to "the flag is broken" without testing the path users actually take. That was the error, and you were right to call it untested.

Sorry for the noise on this and on #3898. Nothing needed from you — leaving it closed is correct.

PR #3825, opened 2026-08-10
fix(cart): make cart mutations atomic and stop checkout from wiping concurrent adds

### LINKED ISSUE #3824: Cart service has no atomicity guarantees — concurrent AddItem calls lose updates, and checkout's EmptyCart can silently discard items added during checkout
**Version:** verified against commit `6c4b663a8a2c243f60f6b1a7d93e52467b3124f1` on `main` (2026-08-10).

## Summary

`ValkeyCartStore` stores each user's cart as a single serialized blob under one Redis/Valkey hash field and mutates it with an unprotected **read → modify → write** sequence (no `WATCH`/`MULTI`, no Lua script, no per-field atomic ops, no optimistic-concurrency/version check). Because of this, concurrent writes to the same cart are not serialized in any way, which produces two distinct, reproducible failure modes:

1. **Lost updates on concurrent `AddItem`** — two "add to cart" calls for the same user that overlap in time silently clobber each other; one increment is lost and the user is undercharged/under-fulfilled with no error surfaced anywhere (not in the response, not in a span, not in a log).
2. **Silent item loss during checkout** — `checkoutservice.PlaceOrder` calls `cartservice.GetCart` and then, after building the order, calls `cartservice.EmptyCart` (`src/checkout/main.go:520` and `:528`). If an `AddItem` for that same user lands in the window between those two calls (e.g. a second browser tab, a retried request, or the load generator), `EmptyCartAsync` unconditionally overwrites the hash field with an empty cart, silently discarding an item that was **never included in the placed order and is never billed** — a silent data/revenue-consistency bug with no telemetry signal indicating anything went wrong.

Both symptoms trace back to the same root cause in `src/cart/src/cartstore/ValkeyCartStore.cs`, so this is one subsystem-level defect, not two unrelated ones.

## Reproduction steps

**A. Lost update on concurrent AddItem**

1. `docker compose up` (or run `cartservice` against Valkey directly).
2. Fire two concurrent `AddItem` gRPC calls for the same `user_id`/`product_id` (e.g. two `grpcurl` invocations launched in parallel, or two goroutines calling the generated client):
   ```
   grpcurl -plaintext -d '{"user_id":"race-user","item":{"product_id":"OLJCESPC7Z","quantity":1}}' localhost:5070 oteldemo.CartService/AddItem &
   grpcurl -plaintext -d '{"user_id":"race-user","item":{"product_id":"OLJCESPC7Z","quantity":1}}' localhost:5070 oteldemo.CartService/AddItem &
   wait
   ```
3. Call `GetCart` for `race-user`.
4. **Expected:** quantity `2`. **Actual (intermittently, under real network/scheduling jitter):** quantity `1` — one of the two increments is lost because both requests read the same pre-update blob before either writes it back (`ValkeyCartStore.cs:140-165`).

**B. Silent item loss racing with checkout**

1. Add an item to a cart for `user_id=checkout-race-user`.
2. Call `checkoutservice.PlaceOrder` for that user, but pause/delay the request (or add a temporary sleep) between its `GetCart` call and its `EmptyCart` call (`src/checkout/main.go:520-528`) to simulate realistic latency under load.
3. During that window, call `cartservice.AddItem` again for the same user with a different product.
4. Inspect the placed order (it will not contain the item added in step 3 — it wasn't in the cart when `GetCart` ran) and then call `GetCart` again.
5. **Expected:** the item added in step 3 should either be in a fresh cart or surfaced as a conflict. **Actual:** the cart is empty — `EmptyCartAsync` (`ValkeyCartStore.cs:178-194`) does an unconditional `HashSetAsync` overwrite with no check against the state `PlaceOrder` originally read, so the item vanishes: not ordered, not billed, not in the cart, and nothing logs or traces this as an error.

## Root cause

`src/cart/src/cartstore/ValkeyCartStore.cs`:
- `AddItemAsync` (lines 127-176): `db.HashGetAsync` → deserialize → mutate in memory → `db.HashSetAsync`. Two concurrent calls both read the same starting state and the second write wins, discarding the first.
- `EmptyCartAsync` (lines 178-194): unconditional `HashSetAsync` with an empty cart blob — never conditioned on the version/state that the caller (e.g. checkout) originally observed.
- `GetCartAsync` (lines 196-227): plain read with no version/ETag returned to callers, so callers have no way to detect that the cart changed underneath them.

There is no `WATCH`/`MULTI-EXEC` transaction, no Lua script, and no per-item atomic field (e.g. `HINCRBY` on a `productId` field) anywhere in this file — the entire cart is treated as one opaque blob guarded only by a connection-level lock (`_locker`) that protects Redis *connection setup*, not cart *data* mutations.

Compounding this: cart mutation logic currently has **zero effective test coverage**. All three tests in `src/cart/tests/CartServiceTests.cs` (`GetItem_NoAddItemBefore_EmptyCartReturned`, `AddItem_ItemExists_Updated`, `AddItem_New_Inserted`) are marked `[Fact(Skip = ...)]` referencing PR #746 and have been disabled since that PR merged, so there is no regression safety net over `AddItemAsync`/`EmptyCartAsync`/`GetCartAsync` today, and no concurrency test exists for this path at all.

## Impact

- **Data/revenue consistency**: items can be silently dropped from an order during checkout under realistic concurrent load (double-tab, client retry, load generator) — the worst class of bug for a checkout flow, since it fails silently rather than loudly.
- **Correctness under concurrency**: `AddItem` is not idempotent/atomic, so quantities are simply wrong under contention, with no error returned to the caller and no exception/log/span marking the loss.
- **Observability credibility**: this is a project whose entire purpose is to demonstrate correct, observable microservice behavior. A silent, un-instrumented data-loss bug in the reference checkout flow undermines that goal — anyone using this demo to practice diagnosing production issues will find no trace, log, or metric pointing at the actual defect.
- **No regression protection**: because the existing cart tests are skipped, this class of bug can be reintroduced or worsened by future changes without CI ever catching it.

## Affected files/components

- `src/cart/src/cartstore/ValkeyCartStore.cs` — core fix: replace blob read-modify-write with atomic operations (e.g. per-product hash fields with `HashIncrementAsync`, or a Lua script / `WATCH`+`MULTI`+`EXEC` transaction with retry-on-conflict for the whole-cart representation).
- `src/cart/src/cartstore/ICartStore.cs` — may need a version/ETag-aware signature (e.g. optional expected-version parameter or a compare-and-swap style API) so callers like checkout can detect concurrent modification instead of blindly overwriting.
- `src/cart/src/services/CartService.cs` — surface conflict/retry behavior (e.g. return a specific gRPC status) if a CAS-based store reports contention.
- `src/checkout/main.go` (`PlaceOrder`, lines ~520-528) — must not unconditionally call `EmptyCart` after `GetCart` without accounting for concurrent modification; needs either a conditional/versioned empty-cart call or an explicit "remove exactly these items" operation instead of "clear everything."
- `src/cart/tests/CartServiceTests.cs` — re-enable the disabled tests (fixing whatever made them skip in #746) and add concurrency-focused tests reproducing both failure modes above.
- Documentation: any architecture/README notes describing the cart storage contract should note the atomicity guarantees once fixed.

## Acceptance criteria

1. Two concurrent `AddItem` calls for the same user/product each register their full increment — no lost updates under concurrent load (verified by an automated concurrency test, e.g. N parallel `AddItem` calls resulting in the correct summed quantity).
2. `EmptyCart` (or whatever replaces the checkout-side cart-clearing call) cannot silently discard an item that was added after the state `PlaceOrder`/the caller last observed; the system either preserves the new item or surfaces an explicit, observable conflict (span exception, error status, or documented compensating behavior) — no silent loss.
3. The three currently-skipped tests in `CartServiceTests.cs` are re-enabled and passing, plus new tests cover the concurrent-AddItem and checkout/AddItem-race scenarios described above.
4. No behavioral regression to the existing single-writer cart flow (add/get/empty) used throughout the rest of the demo (frontend, load generator, checkout, recommendation).

## High-level implementation plan

1. Redesign the Valkey cart schema in `ValkeyCartStore.cs` to make mutations atomic — most direct option: store each cart item as its own hash field (`HINCRBY` for quantity) instead of one serialized blob, eliminating the read-modify-write pattern for `AddItem` entirely; alternatively, keep the blob but wrap read+modify+write in a `WATCH`/`MULTI`/`EXEC` transaction with bounded retry on conflict, or a Lua script for atomicity.
2. Update `EmptyCartAsync`/checkout's clear-cart call so it can't blindly wipe state that has changed since it was last read — e.g. accept the expected item set/version and only clear what was actually ordered, or fail/report conflict rather than silently overwrite.
3. Update `ICartStore`/`CartService` gRPC surface as needed to propagate conflict signals, keeping backward-compatible behavior for the common non-concurrent path.
4. Update `checkoutservice.PlaceOrder` in `src/checkout/main.go` to use the new conflict-aware clear operation.
5. Re-enable and fix the three skipped tests in `CartServiceTests.cs`; add targeted concurrency tests for both reproduction scenarios above.
6. Verify manually with the reproduction steps in this issue, and note the fix in any relevant architecture docs describing cart storage guarantees.

bhuvan-somisetty:
# Changes

Fixes #3824

## Problem

`ValkeyCartStore` stored each cart as a single serialized blob and mutated it with an unprotected read -> modify -> write. Two concurrent `AddItem` calls for the same cart could silently lose one of the updates. Separately, `checkoutservice.PlaceOrder` called `EmptyCart` unconditionally after reading the cart, so an item added in the window between `GetCart` and `EmptyCart` (a second tab, a retried request, the load generator) was silently discarded — never ordered, never billed, and with no error or trace surfaced anywhere.

Repro steps for both failure modes are in the issue.

## Fix

- `AddItemAsync` now applies its mutation through an optimistic-concurrency retry loop: it reads the cart, applies the change, then writes it back in a Redis transaction conditioned on the hash field still matching what was read (`Condition.HashEqual`/`HashNotExists`). If another writer changed the cart in between, the condition fails, and we retry against the fresh state (bounded retries with jittered backoff) instead of overwriting the concurrent write.
- `AddItemAsync` also supports negative quantity deltas, removing the item once its quantity reaches zero or below.
- Checkout no longer calls `EmptyCart` after placing an order. It calls the existing `AddItem` RPC once per ordered item with a negative quantity, removing exactly what was ordered rather than clearing the whole cart. Anything added concurrently now survives instead of being silently wiped. This reuses the existing `AddItem` RPC/proto contract, so there's no wire-format change — `EmptyCart` is untouched and still used for the frontend's explicit "empty cart" action.

## Tests

- Re-enabled the three `CartServiceTests` that had been skipped since #746 (the test host wiring was incomplete — no `Configure`/`ConfigureServices` was ever registered, so nothing actually ran against a cart store). Fixed the host setup to stand up a real `CartService` backed by `ValkeyCartStore` against a Valkey instance.
- Added a concurrency test: 20 parallel `AddItem` calls for the same product sum to the correct total (previously would have lost updates).
- Added a test reproducing the checkout race: an item added concurrently between the "order read" and "order cleanup" steps is preserved, while the ordered item is removed.
- Added a `cart-unit-tests` CI job that dynamically uses the Valkey image from `.env` so these run on every PR going forward.

Verified locally: `dotnet test` (5/5 passing against a real Valkey container), `go build ./...` and `go vet ./...` for checkout.

## Merge Requirements

* [ ] Appropriate documentation updates in the docs (not applicable, internal state consistency bug fix)
* [ ] Appropriate Helm chart updates (not applicable, no configuration surface changed)

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -145,6 +145,34 @@ jobs:
             "${WEAVER_IMAGE}" \
             registry check -r source
 
+  cart-unit-tests:
+    name: Cart unit tests
+    runs-on: ubuntu-latest
+    services:
+      valkey:
+        image: ghcr.io/valkey-io/valkey:9.0.4-alpine3.23
+        ports:
+          - 6379:6379
+        options: >-
+          --health-cmd "valkey-cli ping"
+          --health-interval 5s
+          --health-timeout 3s
+          --health-retries 10
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Set up .NET
+        uses: actions/setup-dotnet@a98b56852c35b8e3190ac28c8c2271da59106c68  # v6.0.0
+        with:
+          dotnet-version: '10.0.x'
+      - name: Run cart unit tests
+        working-directory: src/cart
+        env:
+          VALKEY_ADDR: localhost:6379
+        run: dotnet test tests/cart.tests.csproj
+
   check-react-native-changes:
     name: Check React Native changes
     runs-on: ubuntu-latest
@@ -198,6 +226,7 @@ jobs:
       sanity,
       checklicense,
       weaver-check,
+      cart-unit-tests,
       react-native-build,
       codeql-analysis,
       zizmor
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -150,7 +150,7 @@ jobs:
     runs-on: ubuntu-latest
     services:
       valkey:
-        image: ghcr.io/valkey-io/valkey:9.0.4-alpine3.23
+        image: ghcr.io/valkey-io/valkey:9.0.4-alpine3.23@sha256:107b203169a20ef7c87548c157f4e4a833dbbbd20fefd3a26740b7f98ac80216
         ports:
           - 6379:6379
         options: >-
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -145,6 +145,34 @@ jobs:
             "${WEAVER_IMAGE}" \
             registry check -r source
 
+  cart-unit-tests:
+    name: Cart unit tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Start Valkey
+        run: |
+          VALKEY_IMAGE=$(grep -oP '(?<=^VALKEY_IMAGE=)\S*' .env)
+          docker run -d --rm --name valkey -p 6379:6379 "${VALKEY_IMAGE}"
+          for i in {1..30}; do
+            if docker exec valkey valkey-cli ping 2>/dev/null | grep -q PONG; then
+              break
+            fi
+            sleep 1
+          done
+      - name: Set up .NET
+        uses: actions/setup-dotnet@a98b56852c35b8e3190ac28c8c2271da59106c68  # v6.0.0
+        with:
+          dotnet-version: '10.0.x'
+      - name: Run cart unit tests
+        working-directory: src/cart
+        env:
+          VALKEY_ADDR: localhost:6379
+        run: dotnet test tests/cart.tests.csproj
+
   check-react-native-changes:
     name: Check React Native changes
     runs-on: ubuntu-latest
@@ -190,6 +218,7 @@ jobs:
       sanity,
       checklicense,
       weaver-check,
+      cart-unit-tests,
       react-native-build,
       codeql-analysis
     ]
--- a/src/cart/src/cartstore/ValkeyCartStore.cs
+++ b/src/cart/src/cartstore/ValkeyCartStore.cs
@@ -18,6 +18,7 @@ public class ValkeyCartStore : ICartStore
     private readonly ILogger _logger;
     private const string CartFieldName = "cart";
     private const int RedisRetryNumber = 30;
+    private const int CartMutationMaxAttempts = 30;
 
     private volatile ConnectionMultiplexer _redis;
     private volatile bool _isRedisConnectionOpened;
@@ -132,38 +133,28 @@ public class ValkeyCartStore : ICartStore
 
         try
         {
-            EnsureRedisConnected();
-
-            var db = _redis.GetDatabase();
-
-            // Access the cart from the cache
-            var value = await db.HashGetAsync(userId, CartFieldName);
-
-            Oteldemo.Cart cart;
-            if (value.IsNull)
+            await MutateCartAsync(userId, cart =>
             {
-                cart = new Oteldemo.Cart
-                {
-                    UserId = userId
-                };
-                cart.Items.Add(new Oteldemo.CartItem { ProductId = productId, Quantity = quantity });
-            }
-            else
-            {
-                cart = Oteldemo.Cart.Parser.ParseFrom((byte[])value);
                 var existingItem = cart.Items.SingleOrDefault(i => i.ProductId == productId);
                 if (existingItem == null)
                 {
-                    cart.Items.Add(new Oteldemo.CartItem { ProductId = productId, Quantity = quantity });
+                    if (quantity > 0)
+                    {
+                        cart.Items.Add(new Oteldemo.CartItem { ProductId = productId, Quantity = quantity });
+                    }
+                    return;
                 }
-                else
-                {
-                    existingItem.Quantity += quantity;
-                }
-            }
 
-            await db.HashSetAsync(userId, new[]{ new HashEntry(CartFieldName, cart.ToByteArray()) });
-            await db.KeyExpireAsync(userId, TimeSpan.FromMinutes(60));
+                existingItem.Quantity += quantity;
+                if (existingItem.Quantity <= 0)
+                {
+                    cart.Items.Remove(existingItem);
+                }
+            });
+        }
+        catch (RpcException)
+        {
+            throw;
         }
         catch (Exception ex)
         {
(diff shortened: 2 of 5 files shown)


bhuvan-somisetty:
@julianocosta89 PTAL, whenever u get chance.

.github/workflows/checks.yml
```diff
@@ -145,6 +145,34 @@ jobs:
...
+          --health-retries 10
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          persist-credentials: false
+      - name: Set up .NET
+        uses: actions/setup-dotnet@a98b56852c35b8e3190ac28c8c2271da59106c68  # v6.0.0
+        with:
+          dotnet-version: '10.0.x'
```
julianocosta89: Is there a way to use the same images defined on the Dockerfiles?
maybe abstract the images somewhere else that can be used in both places.

I can see this code going stale really soon. I'd rather try to avoid it

src/cart/src/cartstore/ValkeyCartStore.cs
```diff
@@ -175,6 +167,48 @@ public async Task AddItemAsync(string userId, string productId, int quantity)
         }
     }
 
+    // Applies `mutate` to the user's current cart and writes it back with an optimistic-concurrency
+    // check (Redis transaction conditioned on the hash field still matching what we just read), retrying
+    // on conflict. This closes the lost-update window that a plain read -> modify -> write sequence has
+    // when two callers mutate the same cart concurrently.
```
julianocosta89: Do we really need all those comments here?
Is that aligned with what we state in AGENTS.md?

src/checkout/main.go
```diff
@@ -524,11 +525,24 @@ func (cs *checkout) getUserCart(ctx context.Context, userID string) ([]*pb.CartI
 	return cart.GetItems(), nil
 }
 
-func (cs *checkout) emptyUserCart(ctx context.Context, userID string) error {
-	if _, err := cs.cartSvcClient.EmptyCart(ctx, &pb.EmptyCartRequest{UserId: userID}); err != nil {
-		return fmt.Errorf("failed to empty user cart during checkout: %+v", err)
+// removeOrderedItemsFromCart removes exactly the items that were placed in this order, rather than
+// clearing the whole cart. Clearing unconditionally would silently discard anything the user added
+// to their cart in the window between prepareOrderItemsAndShippingQuoteFromCart's GetCart call and
+// this call (e.g. a second browser tab, a retried request, or the load generator).
```
julianocosta89: Also here, do we need 4 lines of comments?
Adding a lot of comments to the code base is the easiest way to make comments stale.

Please take a look at AGENTS.md to find a definition on how and when to add comments in-code.

bhuvan-somisetty:
Hi @julianocosta89 - thanks for the feedback!

I've updated the CI step to use `docker compose up -d --wait valkey-cart` so it reuses `compose.yaml`/`.env`, removed the in-code comments per `AGENTS.md`, and rebased on latest `main`.

Let me know if this looks good to you!

src/cart/src/cartstore/ValkeyCartStore.cs
```diff
@@ -132,38 +133,29 @@ public async Task AddItemAsync(string userId, string productId, int quantity)
...
-                cart.Items.Add(new Oteldemo.CartItem { ProductId = productId, Quantity = quantity });
-            }
-            else
-            {
-                cart = Oteldemo.Cart.Parser.ParseFrom((byte[])value);
                 var existingItem = cart.Items.SingleOrDefault(i => i.ProductId == productId);
                 if (existingItem == null)
                 {
-                    cart.Items.Add(new Oteldemo.CartItem { ProductId = productId, Quantity = quantity });
+                    // A negative delta for a product that isn't in the cart has nothing to remove.
```
julianocosta89:
```suggestion
```

Please remove this comment.

We do not need comment to explain things that work as expected.

``` 
// This sums 1 plus 1
sum(1+1)
```

The code is self explanatory.
We should only add comments when we have workaround behaviours as explained in https://github.com/open-telemetry/opentelemetry-demo/blob/main/AGENTS.md#code-comments.

bhuvan-somisetty: Removed that comment as well. Thanks for pointing it out!

Issue #606, opened 2022-11-22
operator instrumentation inject support


devrimdemiroz:
# Feature Request

Currently, auto-instrumentation is injected via [Docker file](https://github.com/open-telemetry/opentelemetry-demo/blob/main/src/adservice/Dockerfile).

While having it for Docker deployment is fine, for Kubernetes deployment, it is a preferred option to get it injected automatically via opentelemetry-operator.

**Describe the solution you'd like:**

Instead, agent(s) can be injected via opentelemetry-operator

```
annotations:
  instrumentation.opentelemetry.io/inject-java: otel-agent-xxxxx

```

JaredTan95:
I think it should be solved by https://github.com/open-telemetry/opentelemetry-demo/pull/244

becauese the opentelemetry-operator will inject the `java` agent sidecar/initcontainer and override the `JAVA_TOOL_OPTIONS` value.

```bash
annotations:
  instrumentation.opentelemetry.io/inject-java: otel-agent-xxxxx
```

devrimdemiroz:
@JaredTan95 ,
 I put java as an example. Same applies for other languages so far supported via operator injection.

puckpuck:
All the existing services in the demo have manual instrumentation, which requires them to be compiled with the Otel SDKs/Agents for them to run.

There is the potential for this with the addition of new services where we don't do any manual instrumentation in them and have those new services auto-instrumented only using operator injection. In particular, I think the [Message queue](https://github.com/open-telemetry/opentelemetry-demo/pull/457) implementation with 2 new services would be a good candidate for this.

devrimdemiroz:
@puckpuck 

I still did not deep dive into particular services. However, [here](https://github.com/open-telemetry/opentelemetry-demo/tree/main/docs)  (except C++) each service is marked with both auto and manual instrumentation. If the page is correct, we will lose the manual instrumentation provided as demo but still have auto instrumentation in hand. Do I get it right?

puckpuck:
To provide some status on this issue.

As a top-level item, moving to support the Operator is dependent on either adding similar functionality to the existing docker-based dev environment or dropping support completely for docker as a local dev environment which is currently up for discussion in #635 

In the SIG meetings, we have decided to add auto-instrumentation via the Operator for the `accountingservice` and `fraudservice` components of the demo. Since these services communicate with Kafka, we need to ensure they are written in languages that have Kafka auto-instrumentation available and can be instrumented with the Operator.

[...]

Issue #650, opened 2022-12-21
Make it possible to disable OpenTelemetry Instrumentation


svrnm:
# Feature Request

**Is your feature request related to a problem?**

#589, #606 and some other issues ask for the capability to autoinstrument services, especially by using the otel operator which comes with the capability to inject Automatic Instrumentation. While creating new services which are then auto instrumented is a possibility, I'd like to propose another solution: 

By making it possibile to disable manual OpenTelemetry instrumentation, it would be possible to apply injected automatic instrumentation via the operator instead. 

This not only removes the need to add even more services, it also allows has a few other benefits:

* You could "gradually" instrument your application and see how a full picture builds, or you could leave a certain service un-instrumented on purpose to see what information you still get from an upstream service.
* from a vendor perspective I could use my existing APM agents instead and have a way to compare what I see with OpenTelemetry and what I see with my existing solution.

**Describe the solution you'd like:**

My naive assumption is that this might be fairly easy to implement:
* Add an environment variable like `OTEL_DO_NOT_INSTRUMENT="true"`
* In the code of a service, put the SDK initalisation into a condiational block that checks for the value of that (`if(!env.OTEL_DO_NOT_INSTRUMENT) { intializeSDK(); }`)

svrnm:
follow up on this, I have played around with this in [this branch](https://github.com/svrnm/opentelemetry-demo/tree/disable-instrumentation-feature):

* Java (ad-service, frauddetection-service) works just fine without the agent
* NodeJS (frontend, paymentservice) fail with a lot of errors, because the API returns `undefined` as spans if the SDK is not loaded (see https://github.com/open-telemetry/opentelemetry-js/issues/3316, cc @open-telemetry/javascript-approvers)

PR #3868, opened 2026-08-19
refactor(payment): depend on the OTel API only, not the SDK

TobiTomson:
# Changes

Until now the application source depended on the OpenTelemetry SDK. That dependency is removed; the source now depends only on @opentelemetry/api. The SDK and its instrumentations are supplied by copying in the OpenTelemetry Operator's auto-instrumentation image at build time.

For the payment service, this enables the use of the OTel Operator for instrumentation in Kubernetes, which is requested in #606. The same pattern can later be applied to other services as well.

## Verification:
Beside the passing unit tests - which verify that generally telemetry does get generated, I verified that logs, metrics and traces generated "manually" via the OTel API also still show up.

Traces:
image

Metrics:
image

Logs:
Screenshot 2026-08-19 at 19 45 09


julianocosta89:
@TobiTomson thanks for the contribution!

One thing that I didn't understand, if the auto-instrumentation is part of the container, how does that allow SDK swap?
Shouldn't the SDK be configured to come as build arg?
If AUTO_INSTRUMENTATION==true, download the SDK and call `--require`, else skip?

cyrille-leclerc:
Successfully tested:
I see the env var `NODE_OPTIONS: --require /otel-auto-instrumentation-nodejs/autoinstrumentation.js` domnstrating I test the changed code and the trace is good:
image

It's a great improvement 🎉

cyrille-leclerc (approved):

TobiTomson:
Good point actually - I hadn't thought of it that way. I think we need to differentiate between two things here:
1. the ability to easily swap SDKs, such as for auto-instrumentation with the operator (as requested in #606)
2. the ability to completely disable auto-instrumentation (as requested in #650)

What I've done here is 1., but it is explicitly not 2.

The way that swapping SDKs is possible is by changing the FROM in the Dockerfile that pulls it in. This approach works for docker-compose - because you can just edit the Dockerfile - and also works for the operator - because the way the operator injects the SDK is by adding an emptyDir mount to the exact same position we copied it to (which overrides our copy) and then copying in its version via an init-container. The operator also overrides the NODE_OPTIONS env var, thus completely replacing everything.

I take two questions from you:
a) should we introduce AUTO_INSTRUMENTATION as additional env var?
I think that this would just make the code harder to read without a big benefit (we are already able to swap SDKs - Cyrille tested it as well).

b) should we use a build ARG?
The upside of using ARG: makes the ability to swap agents more obvious, and allows doing so without touching the Dockerfile.
The downside of using ARG: it's not discoverable by dependabot (while using a FROM without ARG is discoverable).

I think that making it discoverable to dependabot is a bigger win.
One clear downside (irrespective of using ARG or not) of my approach: since the SDK now comes via the image rather than compose, a stale image means silently zero telemetry with a healthy container.
So we should probably agree on how to do this in general.

@julianocosta89 what do you think?

julianocosta89:
> the ability to easily swap SDKs, such as for auto-instrumentation with the operator

is that easily swappable though?

> The way that swapping SDKs is possible is by changing the FROM in the Dockerfile that pulls it in. 

If I want to use the operator, I'd need to change the Dockerfile to remove the SDK and the `--require` and later configure the operator. That is not easy.
I mean, it is easy, but how is that different from removing the SDK and rebuilding the service?

The user can do that already if they want.

Having a build arg, or an environment variable to control SDK injected `true/false`, and even point to the SDK would be easier, because then the user do not need to rebuild and redeploy the service.
The user would simply config the service to use SDK from XYZ.

> The downside of using ARG: it's not discoverable by dependabot (while using a FROM without ARG is discoverable).

@Kielek and I have plans to bring Renovate to the Demo. Renovate has custom rules that we can point to specific places in the codebase and it would keep track of it.

TobiTomson:
To use the operator you don't need to change anything. You use this image, and it just works (because it is now able to just swap out what we baked into the image at runtime). Which was not the case before. That's exactly why I did this change.

julianocosta89:
> To use the operator you don't need to change anything. You use this image, and it just works (because it is now able to just swap out what we baked into the image at runtime). Which was not the case before. That's exactly why I did this change.

Ohhh, that's interesting!

cyrille-leclerc:
@TobiTomson's change is IMO a great step forward, it's the right level of granularity for an incremental improvement.

Once this is in place, we'll be well-positioned to tackle SDK injection. This PR opens the door to several options, including:

* Adding a condition in the Docker image to control whether the bundled OTel SDK loads at startup:
   * Load the bundled OTel SDK in containerized environments without SDK injection support (e.g., Docker Compose, or Kubernetes without the OTel Operator)
   * Skip loading the bundled OTel SDK in environments that support OTel SDK injection (e.g., the OTel Operator)
* Introducing two separate Docker image layers: one with the OTel SDK bundled, one without
* The behavior Tobias explained above
* ...

julianocosta89:
@osullivandonal and @ps48 did you have a chance to take a look at this one?

julianocosta89:
@TobiTomson would you mind resolving the conflicts?

TobiTomson:
After the SIG call 2 weeks ago I did some more testing to validate telemetry produced after this change. Here is what I found:
1. regarding duplication of telemetry I did a test run (20 identical charges per variant, both images built from the same commit with only the PR diff applied, raw OTLP captured with no collector processors, success and failure paths):
	- Traces: identical. 44 spans each, same names, counts and shape. One charge span and one exception event per request on both sides on failure path - span.recordException is not doubled.
	- Logs: not identical - count goes from 82 to 41. But this is actually good! Before this PR, log export was duplicated, once from @opentelemetry/instrumentation-pino and once from payment-logger. By removing the manual pino instrumentation we actually inadvertently fixed this. Also: the old instrumentation was lacking service.name on logs and had an incorrectly hard-coded version set to 1.0.0 - now both contain the proper values. Additionally, before our change, stdout had an OTLP dump of the logs, now it's clean JSON (example below at the end of my comment).
	- Metrics: demo.payment.transactions unaffected

2. The operator image lags slightly behind the latest available SDK. Right now this means concretely that because we're one minor version behind, there is a small impact on generated telemetry: with latest SDK there is a change to "only emit stable http, network and database attributes"; with the operator image tcp.connect and the flagd client span go back to net.peer.ip/net.host.port, and http.client.request.duration becomes http.client.duration. I don't think it's a huge problem, but worth noting that with this change we're bound to the operator image's release cadence.


---


Example of stdout before and after:
**before:**
{
  resource: {
    attributes: {
      'service.instance.id': '5e5a601e-0fd9-4e91-90e5-9989a7c7c0fd',
      'process.pid': 1,
      'process.executable.name': '/nodejs/bin/node',
      'process.executable.path': '/nodejs/bin/node',
      'process.command_args': [
        '/nodejs/bin/node',
        '/usr/src/app/node_modules/thread-stream/lib/worker.js'
      ],
      'process.runtime.version': '22.22.0',
      'process.runtime.name': 'nodejs',
      'process.runtime.description': 'Node.js',
      'process.command': '/usr/src/app/node_modules/thread-stream/lib/worker.js',
      'process.owner': 'nonroot',
      'os.type': 'linux',
      'os.version': '6.8.0-117-generic',
      'host.name': '24b08a9b6bf5',
      'host.arch': 'arm64',
      'service.namespace': 'opentelemetry-demo',
      'service.version': '2.1.1',
      'service.criticality': 'critical',
      'service.name': 'payment'
    }
  },
  instrumentationScope: { name: 'payment-logger', version: '1.0.0', schemaUrl: undefined },
  timestamp: 1787824679664000,
  traceId: '7ca1127a3abce5e05a9ebe39ffef3f3f',
  spanId: 'f7c530d2b6678cfb',
  traceFlags: '01',
  severityText: 'info',
  severityNumber: 9,
  eventName: undefined,
  body: 'Charge request received.',
  attributes: {}
}

**after:**
{"level":"info","time":1787824783613,"pid":1,"hostname":"c00c21cea7ea","trace_id":"a85b4af26aecd6a85edf133889742d7a","span_id":"47e9935b96f05a04","trace_flags":"01","service.name":"payment","msg":"Charge request received."}

TobiTomson:
To make clear what I meant by the changes to stdout:
stdout becomes cleaner and easier to read - the resource attributes are still present on OTLP logs though. So nothing gets lost.

osullivandonal (approved):

julianocosta89 (approved):

PR #3875, opened 2026-08-22
fix(shipping): don't panic if flagd is unreachable at startup

### LINKED ISSUE #3845: shipping panics and crash-loops if flagd isn't reachable within 5 startup retries
# Bug Report

Commit 7675680588f61393da48f6077115e6070c0dfae3 (upstream `main`).

## Symptom

`shipping` panics and crash-loops on startup if `flagd` isn't reachable
within its 5 connection retries (~31s exponential backoff). Kubernetes
gives no ordering guarantee between `shipping` and `flagd` deployments,
so on a fresh install or rollout `shipping` frequently starts before
`flagd` is ready, exhausts its retries, and crashes.

`shipping` should start and serve traffic even if `flagd` isn't reachable
yet, similar to `checkout` (Go), which only logs an error and keeps
running when it can't connect to flagd — `shipping_service.rs` already
treats flag-resolution errors as non-fatal at the call site (defaults
`intlShippingSlowdown` to 0), so the service is designed to tolerate a
missing flag provider everywhere except this startup path.

`shipping`'s process panics and exits, so the pod crash-loops until it
happens to restart after `flagd` is up:

    WARN open_feature_flagd::resolver::rpc: Connection attempt 1 failed, retrying in 1000ms: transport error
    ...
    ERROR open_feature_flagd::resolver::rpc: Failed to establish connection after 5 attempts

    thread 'main' (1) panicked at src/main.rs:55:6:
    Failed to initialize flagd provider: Provider("transport error")

Root cause: `src/shipping/src/main.rs:50-55` calls
`FlagdProvider::new(...).await.expect("Failed to initialize flagd provider")`,
which panics on any connection failure at startup instead of degrading
gracefully.

## Reproduce

1. Fresh install the demo on Kubernetes (e.g. `helm install` the
   `opentelemetry-demo` chart into a new namespace on minikube).
2. Watch the `shipping` pod during the initial rollout:
   `kubectl logs -n <namespace> -l app.kubernetes.io/name=shipping -f`
3. If `flagd` isn't ready within ~31s of `shipping` starting, `shipping`
   panics and its pod restarts (`kubectl get pods -n <namespace> -l
   app.kubernetes.io/name=shipping` shows `CrashLoopBackOff` /
   non-zero restarts).

Reliable forced repro: scale `flagd` to 0 replicas, then delete/restart
the `shipping` pod. Wait ~31s and it panics every time. Scale `flagd`
back up and restart `shipping` again — it starts cleanly once `flagd` is
reachable.

## Additional Context

Verified locally: replacing the `.expect(...)` with a fallback to
`open_feature::provider::NoOpProvider` (already available via the
`open-feature = "0.3"` dependency) on connection failure — logging the
error instead of panicking — lets `shipping` start and serve traffic
normally during this same race, matching `checkout`'s non-fatal handling.
Happy to open a PR with this change if maintainers agree with the
direction; one thing worth discussing first: this fallback is a one-shot
decision, so if it triggers, that `shipping` process runs on the NoOp
provider for its process lifetime (no background reconnect) until the pod
restarts for some other reason.

Branch: https://github.com/dpacheconr/opentelemetry-demo/tree/fix/shipping-flagd-panic
PR-vs-upstream diff link: https://github.com/dpacheconr/opentelemetry-demo/pull/new/fix/shipping-flagd-panic

mwimpelberg28:
Fixes #3845.

`shipping` panics and crash-loops if it can't reach `flagd` within its
5 startup connection retries (~31s). On Kubernetes there's no ordering
guarantee between the `shipping` and `flagd` deployments, so this is
common on a fresh install or rollout.

This falls back to `open_feature::provider::NoOpProvider` (logging the
error) when `FlagdProvider::new` fails, instead of panicking. This
matches how `checkout` (Go) handles the same race — it only logs and
keeps running. `shipping_service.rs` already treats flag-resolution
errors as non-fatal at the call site (defaults `intlShippingSlowdown`
to 0), so the service tolerates a missing flag provider everywhere
except this startup path.

Note this fallback is a one-shot decision: if it triggers, that
`shipping` process runs on the NoOp provider for its process lifetime
(no background reconnect) until the pod restarts for some other reason.

Tested locally: scaled `flagd` to 0 replicas, restarted `shipping` —
it now starts cleanly and serves traffic instead of crash-looping.


julianocosta89:
I'd rather have the container in crashloop till it can connect with flagd than having the service running, but not working as expected because of NoOp flagd

dpacheconr:
@julianocosta89 this seems to differ to other services behaviour as per https://github.com/open-telemetry/opentelemetry-demo/issues/3845

Kielek (changes requested):
> I'd rather have the container in crashloop till it can connect with flagd than having the service running, but not working as expected because of NoOp flagd

Fully agree with this, we should wait until flagd is healthy on the docker startup level.

dpacheconr:
@julianocosta89 @Kielek do you want PR to change the checkout service to match this intended behaviour then, is that approach you want to go with?

julianocosta89:
@dpacheconr the main problem that I see here is the `NoOp` part.
AFAIK the other services start and keep trying to connect till it does.

If they were `NoOp` the flags would be flaky.

dpacheconr:
then I would recommend we go with initcontainer approach like other services already implement
PR raised https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2384

julianocosta89:
Approved the Helm chart update.
Closing this one

PR #3873, opened 2026-08-21
revert(load-generator): replace k6 with Locust

mwimpelberg28:
Reverts #3564. Restores the pre-k6 Locust-based load generator (locustfile.py, requirements, Envoy /loadgen/ route, compose service config) while keeping everything unrelated added since that commit.

This reverts commit 18b36c73ccc2dbc86759dab2e0ef05175a7a8ca5.

Assisted-by: Claude Sonnet 5

# Changes

Please provide a brief description of the changes here.

## Merge Requirements

For new features contributions, please make sure you have completed the following
essential items:

* [ ] `CHANGELOG.md` updated to document new feature additions
* [ ] Appropriate documentation updates in the [docs][]
* [ ] Appropriate Helm chart updates in the [helm-charts][]

Maintainers will not merge until the above have been completed. If you're unsure
which docs need to be changed ping the
[@open-telemetry/demo-approvers](https://github.com/orgs/open-telemetry/teams/demo-approvers).

[docs]: https://opentelemetry.io/docs/demo/
[helm-charts]: https://github.com/open-telemetry/opentelemetry-helm-charts

PR #3903, opened 2026-08-27
feat(load-generator): make browser scenario VU count configurable

### LINKED ISSUE #3813: Load-generator browser scenario shares one browser context across all iterations breaks RUM agents session tracking
browserScenario() in src/load-generator/script.js hardcodes vus: 1 for the browser scenario, independent of the LOAD_GENERATOR_VUS env var that controls the HTTP scenario's scale. This means only a single simulated browser session ever runs, no matter how much load-generator throughput or replica count is scaled up for the rest of the demo — there's no way to generate more than one concurrent browser/RUM session.

This makes it hard to load-test or exercise a browser-based RUM/analytics integration at any real scale: the browser side of the load generator is effectively fixed at "one user, forever," while every other part of the traffic generator scales normally.

What is the expected behavior?

The browser scenario's VU count should be independently configurable via an env var (following the same pattern LOAD_GENERATOR_VUS already uses for the HTTP scenario), so it can be scaled up separately from HTTP load.

What is the actual behavior?

vus: 1 is hardcoded in the browser scenario config; there is no way to run more than one concurrent browser session without editing script.js directly.

Reproduce

Run the demo with K6_BROWSER_ENABLED=true.
Try to increase concurrent browser/RUM traffic via any existing env var (LOAD_GENERATOR_VUS only affects httpScenario).
Observe that the browser scenario always runs exactly one VU regardless of configuration.
Proposed fix: add a BROWSER_GENERATOR_VUS env var (defaulting to 1 to preserve current behavior) and use it for the browser scenario's vus, mirroring how LOAD_GENERATOR_VUS already works for httpScenario: vus: parseInt(__ENV.BROWSER_GENERATOR_VUS || '1'). Branch: https://github.com/dpacheconr/opentelemetry-demo/tree/fix/k6-browser-vus-configurable

---
Assisted-by: Claude Sonnet 5

dpacheconr:
Fixes #3813

The k6 browser scenario's VU count was hardcoded to `1`, so browser/RUM traffic couldn't scale independently of the HTTP scenario (which already scales via `LOAD_GENERATOR_VUS`). Adds `BROWSER_GENERATOR_VUS` (default `1`, preserving current behavior) to configure it separately.

## Checklist

- [x] `CHANGELOG.md` updated
- [x] Corresponding documentation updated (`.env`, `compose.yaml`, `src/load-generator/README.md`)
- [ ] Appropriate Helm changes (not applicable — Compose-only env var passthrough, no Helm chart changes needed)

---
Assisted-by: Claude Sonnet 5


julianocosta89:
hey @dpacheconr we will move back to locust on PR: https://github.com/open-telemetry/opentelemetry-demo/pull/3873

K6 has a license that doesn't allow distribution, and it seems that we are not allowed to request an exception: https://github.com/cncf/foundation/issues/1482

I won't get this one and #3904 merged, because they would be overridden rather soon.
Sorry for the trouble

PR #3904, opened 2026-08-27
fix(load-generator): isolate browser context per iteration, fix RUM finalize and CORS preflight

### LINKED ISSUE #3812: k6 browser scenario silently loses RUM telemetry over long runs shared context + close() timing
Since #3564 replaced Locust/Playwright with k6's browser module, browserScenario() in src/load-generator/script.js opens a page on the browser's default context (browser.newPage()) rather than a fresh context per iteration. Because the scenario runs as a single constant-vus VU with duration: 9999h, that context — and its cookies/localStorage/sessionStorage — persists for the entire lifetime of the load generator, shared across every simulated "user."

browserScenario() also calls page.setExtraHTTPHeaders({ baggage: 'synthetic_request=true' }), which applies that header to every request the page makes — including third-party, cross-origin ones. Per the Fetch spec, a custom header on a cross-origin request is a non-simple request and requires a real CORS preflight (OPTIONS). Any third-party CDN that doesn't answer OPTIONS on the affected path then fails outright — confirmed against a real client-side RUM agent's chunk-loading domain, and independently against Google Fonts' fonts.gstatic.com (as shipped in this demo's own _document.tsx).

Separately, one interaction-based Web Vital metric consistently never arrived even in short, clean runs, while page-load metrics did. page.close()/context.close() tears the page down before an agent's visibilitychange/pagehide-triggered send can complete — that metric is finalized and sent at end-of-page-lifecycle by design. Agents that gate on document.visibilityState === "hidden" inside the handler won't see it as "hidden" from a bare dispatchEvent, since that's a real, browser-computed property that dispatching the event alone doesn't change.

Repro: run with K6_BROWSER_ENABLED=true; instrument the frontend with any client-side agent or resource that's cross-origin with crossorigin="anonymous" (a RUM/analytics agent, or just this demo's own Google Fonts loading); observe the request fail a CORS preflight and never complete — confirmable via K6_BROWSER_DEBUG=true browser console output, which the normal deployment doesn't surface.

Fix: isolate a fresh browser.newContext() per iteration; tag same-origin requests with the baggage header via page.route(), scoping the header to same-origin so third-party requests are passed through untouched and never forced into a preflight; override the visibilityState/hidden getters before dispatching visibilitychange, for agents that check the property directly rather than just listening for the event; and increase the post-dispatch dwell before context.close() from 300ms to 8000ms — 300ms isn't enough time for a RUM agent's harvest to actually flush over the network (observed 6-20s between harvests in a real browser session).

Branch: https://github.com/dpacheconr/opentelemetry-demo/tree/fix/k6-browser-rum-context

---
Assisted-by: Claude Sonnet 5

dpacheconr:
## Description

The k6 browser scenario had four related RUM/CORS bugs:

1. **Shared context leak** — `browser.newPage()` alone reuses k6 browser's default persistent context, so cookies/localStorage/sessionStorage leak across every simulated "user" for the VU's entire `9999h` lifetime, corrupting the RUM agent's session bookkeeping over time. Fixed by opening a fresh `browser.newContext()`/`context.newPage()` per iteration.
2. **CORS preflight breakage** — `page.setExtraHTTPHeaders({ baggage: ... })` applied the custom header to every request, including cross-origin ones. A custom header on a cross-origin request is non-simple under the Fetch spec, forcing a real CORS preflight (OPTIONS) — which broke both a third-party RUM agent's chunk-loading domain and Google Fonts' `fonts.gstatic.com`, since most CDNs don't handle an unexpected preflight. Fixed via `page.route()`, scoping the `synthetic_request` baggage header to same-origin requests only so third-party requests pass through untouched and are never forced into a preflight.
3. **RUM beacon never finalized** — many RUM agents only flush their final page-load/interaction metric when their `visibilitychange` handler observes `document.visibilityState === "hidden"` (checked live, not inferred from the event). Dispatching a bare `visibilitychange` event doesn't change that real, browser-computed property, so the pre-close beacon flush silently never ran. Fixed by overriding the `visibilityState`/`hidden` getters before dispatching `visibilitychange`/`pagehide` and closing the context.
4. **Beacon flush cut off mid-flight** — even with (3) fixed, closing the context 300ms after the visibilitychange dispatch wasn't enough dwell time for a RUM agent's harvest to actually flush over the network (observed 6-20s between harvests in a real browser session). Increased the pre-close wait to 8000ms.

## Link to tracking issue

Fixes #3812

## Testing

Verified locally with `K6_BROWSER_ENABLED=true K6_BROWSER_DEBUG=true` against minikube — confirmed no preflight OPTIONS requests to third-party origins, and real beacon requests firing before context close.

## Checklist

- [x] `CHANGELOG.md` updated
- [ ] Appropriate updates in [docs](https://opentelemetry.io/docs/demo/) (not applicable — internal load-generator implementation detail, no user-facing behavior change)
- [ ] Appropriate Helm chart updates in [helm-charts](https://github.com/open-telemetry/opentelemetry-helm-charts) (not applicable — no new/changed env vars or config surface)

---
Assisted-by: Claude Sonnet 5

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ the release.
 
 ## Unreleased
 
+* [load-generator] Fix the k6 browser scenario's RUM context isolation and
+  CORS preflight breakage: use a fresh browser context per iteration instead
+  of a persistent one, scope the `synthetic_request` baggage header to
+  same-origin requests so third-party/CDN resources don't get forced into a
+  CORS preflight, and force `document.visibilityState`/`hidden` before
+  closing so the RUM agent's visibilitychange-triggered beacon flush
+  actually fires
+  ([#3812](https://github.com/open-telemetry/opentelemetry-demo/issues/3812))
 * [react-native-app] Catch errors from `placeOrder` in the Cart screen and
   show an error toast so payment failures are visible to the user instead of
   being silently dropped
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,10 @@ the release.
   CORS preflight breakage: use a fresh browser context per iteration instead
   of a persistent one, scope the `synthetic_request` baggage header to
   same-origin requests so third-party/CDN resources don't get forced into a
-  CORS preflight, and force `document.visibilityState`/`hidden` before
-  closing so the RUM agent's visibilitychange-triggered beacon flush
-  actually fires
+  CORS preflight, force `document.visibilityState`/`hidden` before closing
+  so the RUM agent's visibilitychange-triggered beacon flush actually fires,
+  and give it 8s (up from 300ms) to actually flush over the network before
+  the context closes
   ([#3812](https://github.com/open-telemetry/opentelemetry-demo/issues/3812))
 * [react-native-app] Catch errors from `placeOrder` in the Cart screen and
   show an error toast so payment failures are visible to the user instead of
--- a/src/load-generator/script.js
+++ b/src/load-generator/script.js
@@ -315,11 +315,61 @@ export async function browserScenario() {
         return
     }
 
-    const page = await browser.newPage()
+    // Fresh context per iteration - a shared/persistent context (the k6
+    // browser default when only newPage() is called) leaks cookies and
+    // storage across every simulated "user" for the VU's entire 9999h
+    // lifetime, which breaks the RUM agent's session bookkeeping over time.
+    const context = await browser.newContext()
+    const page = await context.newPage()
+    // Tag first-party requests as synthetic via baggage, scoped to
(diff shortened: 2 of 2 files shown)


dpacheconr:
Pushed an update: switched the CORS-scoping mechanism from an in-page `addInitScript()` fetch/XHR patch to `page.route()`, and increased the post-visibilitychange dwell before `context.close()` from 300ms to 8000ms. Both changes are ported from further testing on a separate branch and were verified end-to-end (no CORS preflight errors, real beacon requests observed, 300ms was confirmed too short for a RUM agent's harvest to flush).

dpacheconr:
closing as per https://github.com/open-telemetry/opentelemetry-demo/pull/3903#issuecomment-5438900488

PR #3931, opened 2026-09-03
Add database lock contention scenario to flagd (productCatalogLockContention)

### LINKED ISSUE #3930: Add database lock contention scenario to flagd
**Is your feature request related to a problem?**

`demo.flagd.json` has no flag that exercises a database-level failure mode. The closest existing flag, `productCatalogFailure`, fails at the application layer, not the DB layer. Postgres (`astronomy-db`) is only used by `product-catalog` (Go) and `accounting` (.NET), and neither has a chaos/fault-injection path targeting the DB connection itself.

Add a `productCatalogLockContention` flag: when enabled, a background goroutine in `product-catalog` periodically opens a transaction and takes an `ACCESS EXCLUSIVE` lock on `catalog.products` for ~30s, blocking `ListProducts`/`GetProduct`/`SearchProducts` for that duration. This is visible via elevated span duration on the existing otelsql-instrumented DB spans, and via `pg_locks`/`pg_stat_activity` on Postgres.

This requires widening `astronomy_user`'s grant on the `catalog` schema from `SELECT` to `SELECT, UPDATE` in `src/postgresql/init.sql` (Postgres requires write-level privilege to take that lock mode), matching the grant `accounting` already has on its own schema.

I've implemented and tested this locally in minikube

**Describe alternatives you've considered.**

Considered resource starvation (connection pool exhaustion), an artificial slow query (`pg_sleep`), and query plan regression. Lock contention was chosen as the most realistic incident type that's still implementable inside the existing `product-catalog` service, with no new infrastructure or chaos service required.

## Additional Context

None.

dpacheconr:
# Changes

Fixes #3930

Adds a `productCatalogLockContention` flagd flag. When enabled, a background goroutine in `product-catalog` periodically takes an `ACCESS EXCLUSIVE` lock on `catalog.products` for ~30s, blocking `ListProducts`/`GetProduct`/`SearchProducts` for that duration — visible via elevated DB span duration and via `pg_locks`/`pg_stat_activity` on Postgres.

Also widens `astronomy_user`'s grant on the `catalog` schema from `SELECT` to `SELECT, UPDATE` in `src/postgresql/init.sql`, since Postgres requires write-level privilege to take that lock mode (matches the grant `accounting` already has on its own schema).

Tested locally in minikube: confirmed the flag toggles live via OFREP, the lock is granted in `pg_locks`, and a concurrent query blocks for the duration of the hold.

## Merge Requirements

* [x] `CHANGELOG.md` updated to document new feature additions (`.chloggen/3930-lock-contention.yaml`)
* [x] Appropriate documentation updates in the [docs][] — a table row needs to be added for `productCatalogLockContention` in opentelemetry.io's Feature Flags page (`open-telemetry/opentelemetry.io`, `content/en/docs/demo/feature-flags/_index.md`). Docs PR: open-telemetry/opentelemetry.io#11573.
* [ ] Appropriate Helm chart updates in the [helm-charts][] — no new service, port, env var, or deployment shape changed


julianocosta89:
Nice scenario — the ticker + atomic single-holder guard, the grant comment in init.sql, and the minikube verification all check out.

One ask before merge: since `src/product-catalog/flags/flags_gen.go` is OpenFeature CLI-generated, please regenerate it with the CLI from the updated `flags.json` (the hand-written block looks identical in style, but we want the generator to be the source of truth). 

Also consider making the 30s lock hold cancellable via `select { case <-ctx.Done(): case <-time.After(30s): }` so shutdown doesn't wait on the lock. Neither blocks the merge; thanks for the thorough description and linked docs PR.

dpacheconr:
> Nice scenario — the ticker + atomic single-holder guard, the grant comment in init.sql, and the minikube verification all check out.
> 
> One ask before merge: since `src/product-catalog/flags/flags_gen.go` is OpenFeature CLI-generated, please regenerate it with the CLI from the updated `flags.json` (the hand-written block looks identical in style, but we want the generator to be the source of truth).
> 
> Also consider making the 30s lock hold cancellable via `select { case <-ctx.Done(): case <-time.After(30s): }` so shutdown doesn't wait on the lock. Neither blocks the merge; thanks for the thorough description and linked docs PR.

@julianocosta89  Both addressed in 8fdc045, ready to review.

julianocosta89 (approved):

PR #3934, opened 2026-09-03
Add AI/LLM incident scenario (aiSlowResponse, aiRunawayAgent)

### LINKED ISSUE #3933: Add AI/LLM incident scenario (aiSlowResponse, aiRunawayAgent) to demo.flagd.json
## Description

`demo.flagd.json` has no scenario simulating an AI/LLM incident. The demo already
ships real LLM plumbing in `src/agent` (LangChain/LangGraph agent, VCR-replayed
by default via `USE_VCR`) and `src/mcp`, instrumented with Traceloop/OpenLLMetry
(`opentelemetry-instrumentation-langchain`), so `gen_ai.*` spans already flow —
but there's no chaos/fault-injection scenario exercising that path end to end,
the way `productCatalogLockContention` (#3930) exercises the database.

## Proposal

Two new `demo.flagd.json` flags, read by `agent`:

- `aiSlowResponse` — numeric flag (`off`/`5sec`/`10sec`), same shape as
  `intlShippingSlowdown`/`imageSlowLoad`. Injects a delay in `ChatLLM` around
  the model call, simulating LLM provider latency degradation — the most
  common real-world AI incident.
- `aiRunawayAgent` — numeric flag (`off`/`on`), forces the agent to keep
  invoking tools toward `GRAPH_RECURSION_LIMIT` instead of stopping, simulating
  a runaway/looping agent (token/cost blowup, deep trace, eventual
  `GraphRecursionError`) — a failure mode unique to agentic AI, not
  reproducible by a plain API-latency/error flag.

Both would be read via an OpenFeature flagd provider added to `src/agent`
(mirroring `src/recommendation`'s `check_feature_flag` pattern — `agent`
currently has none).

## Additional Context

While investigating, found the k8s deployment path also needs a couple of
gaps closed for `agent` and `load-generator` to work as a chart deployment at
all (independent of these two flags):

- `agent`'s chart component has no `FLAGD_HOST`/`FLAGD_PORT` env — without an
  explicit value, Kubernetes' auto-injected legacy `FLAGD_PORT=tcp://<ip>:8013`
  (present for any Service named `flagd`) breaks any `int()`-parsed default.
- `load-generator`'s chart component only sets k6-oriented env
  (`K6_TARGET_URL` etc.), not Locust's (`LOCUST_HOST`/`AUTOSTART`/`HEADLESS`),
  which the currently-shipped `src/load-generator` (Locust) needs to actually
  run headless.

dpacheconr:
Fixes #3933

## Description

Adds two `demo.flagd.json` flags exercising the agent/LLM path:

- `aiSlowResponse` (`off`/`5sec`/`10sec`) — delays `ChatLLM`'s model call by
  the chosen amount, simulating LLM provider latency degradation.
- `aiRunawayAgent` (`off`/`on`) — nudges the agent's system prompt to keep
  calling tools instead of finishing, driving it toward
  `GRAPH_RECURSION_LIMIT` to simulate a runaway/looping agent.

Both are read via a new `src/agent/src/agents/feature_flags.py`, which wires
up an OpenFeature flagd provider for `agent` (it had none — every other
fault-flag service already does; this mirrors `src/recommendation`'s
`check_feature_flag` pattern).

Verified locally end to end (docker compose + a local k8s deployment):
toggling `aiSlowResponse=5sec` measurably added ~5s per model turn to
`/prompt` responses; both flags resolve correctly via flagd's OFREP endpoint.

## Related PRs

- [open-telemetry/opentelemetry-helm-charts#2385][helm-pr] — `agent` and
  `load-generator` chart-level env var fixes needed for this to run as a k8s
  deployment at all (unrelated to these flags specifically — found while
  testing this PR, see #3933).
- [open-telemetry/opentelemetry.io#11575][docs-pr] — Feature Flags docs
  update for `aiSlowResponse`/`aiRunawayAgent`.

## Checklist

- [x] `CHANGELOG.md` updated (via `.chloggen`)
- [x] Documentation has been updated — see the linked docs PR above.

[helm-pr]: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2385
[docs-pr]: https://github.com/open-telemetry/opentelemetry.io/pull/11575


julianocosta89:
Thanks — this looks well put together: the flagd wiring mirrors the recommendation service pattern, both delays are async-safe (`asyncio.sleep` vs `time.sleep` split correctly), and the load-generator prompts replay against the bundled cassettes. 

Two small follow-ups: 

- Declare `gen_ai.input.prompt` in `telemetry-schema/services/load_generator.yaml` so the registry stays the single source of truth.
- Drop the shebang in `src/agent/src/agents/feature_flags.py` (shebang on a pure module; harmless but inconsistent with other agent modules.)

dpacheconr:
> Thanks — this looks well put together: the flagd wiring mirrors the recommendation service pattern, both delays are async-safe (`asyncio.sleep` vs `time.sleep` split correctly), and the load-generator prompts replay against the bundled cassettes.
> 
> Two small follow-ups:
> 
> * Declare `gen_ai.input.prompt` in `telemetry-schema/services/load_generator.yaml` so the registry stays the single source of truth.
> * Drop the shebang in `src/agent/src/agents/feature_flags.py` (shebang on a pure module; harmless but inconsistent with other agent modules.)

@julianocosta89  with our without demo prefix?
demo.gen_ai.input.prompt or just gen_ai.input.prompt

Also 
Other modules (agents.py, llm.py, mcp_client.py, patch_vcr.py, run.py) all have pre existing shebangs

julianocosta89:
without `demo`.
I would actually refer to the official semconv, gen_ai.input.messages: https://github.com/open-telemetry/semantic-conventions-genai/blob/main/docs/registry/attributes/gen-ai.md

dpacheconr:
@julianocosta89 what about the shebang 

other modules (agents.py, llm.py, mcp_client.py, patch_vcr.py, run.py) all have pre existing shebangs

julianocosta89:
yeah, my agent got crazy there.
Sorry about that. Let's keep it

dpacheconr:
@julianocosta89 done as requested, ready to review

src/agent/src/agents/agents.py
```diff
@@ -82,10 +83,18 @@ async def get_tool_list(self):
     async def run_agent(self, input_prompt, history: List[Dict] | None = None):
         model = ChatLLM()
         tools = await self.get_tool_list()
+        system_prompt = "You are a helpful assistant. Be concise and accurate."
+        runaway_iterations = get_int_feature_flag("aiRunawayAgent")
+        if runaway_iterations:
```
ps48, line 88: Built the branch and toggled this on (flagd served `aiRunawayAgent=20`, confirmed via OFREP). `/prompt` still returned a normal single answer, `finish_reason: stop`, zero extra get_recommendations calls, no recursion error. With the default `USE_VCR=True` the cassette replays a non-looping response regardless of the system-prompt change (recorded max ~12 calls, target 20), so this flag is effectively a no-op in the default demo. 

Should we record a looping cassette?

src/agent/src/agents/feature_flags.py
```diff
@@ -0,0 +1,23 @@
...
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+
+from openfeature import api
+from openfeature.contrib.hook.opentelemetry import TracingHook
+from openfeature.contrib.provider.flagd import FlagdProvider
+
+api.set_provider(
```
ps48: Recommendation sets the provider inside if `__name__ == "__main__":`; here set_provider/add_hooks run at import, and this module is imported by `llm.py` and `agents.py`, so importing them has flagd/hook side effects (e.g. in tests). Not fatal, but consider a lazy init or an explicit init from `run.py` to match the pattern.

ps48 (commented):
Thanks, this is a well put together contribution. I built the agent from this branch and brought up `agent + flagd + mcp + collector` to check both flags end to end.

`aiSlowResponse` works cleanly: with the flag off a /prompt call returns in ~0.43s, and with `aiSlowResponse=5sec` the same call takes ~10.3s (about 5s per model turn, 2 turns), and it works under the default `USE_VCR=True` because the delay is independent of the LLM response. Nice async-safe split between time.sleep and asyncio.sleep.

The one issue is `aiRunawayAgent`. Under the demo's default `USE_VCR=True`, toggling it on has no visible effect: the /prompt call returns a normal single answer with `finish_reason: stop`, no extra tool calls, and no recursion error. I confirmed flagd was actually serving the value (OFREP returned {"value":20,"variant":"on"}), so the flag resolves correctly, but the cassette replays the recorded non-looping response regardless of the runaway system-prompt injection. In other words the scenario only fires against a live LLM, not in the configuration the demo runs in by default. Details inline.

Nit/  Description says OFREP, but FlagdProvider(host, port=8013) uses the default RPC resolver on 8013, not OFREP (8016). Code is correct/consistent with recommendation, just the wording.

src/agent/src/agents/agents.py
dpacheconr: @ps48 are the existing cassettes synthetic created or run against those models claude opus 4.7 and azure/gpt5.5?

src/agent/src/agents/feature_flags.py
dpacheconr: Fixed — moved set_provider/add_hooks into an init_feature_flags() function, called explicitly from run.py at startup instead of at module import time. Importing llm.py/agents.py in isolation (e.g. in tests) no longer triggers flagd/hook setup.

will push the changes shortly

src/agent/src/agents/agents.py
dpacheconr: Fixed — recorded real looping interactions into both the claude-opus-4-7 and azure_gpt-5.5 cassettes with aiRunawayAgent enabled, so the flag now has an observable effect under the default USE_VCR=True config, not just against a live LLM.

ps48:
@dpacheconr Thanks for the updates. The import-side-effect fix looks good (`init_feature_flags()` called from `run.py`), and `aiSlowResponse` works. But I still can't get `aiRunawayAgent` to loop under the default `USE_VCR=True`. I built the branch and toggled it on: flagd serves `20`, and I confirmed the agent injects the runaway instruction into the system prompt, but the response for `"Show all available products in the store."` is byte-identical to the flag-off response (4 messages, no repeated `get_recommendations`). 

It looks like the fuzzy `similar_body_matcher` matches the older non-looping recording for that same prompt first (identical human message, score over threshold), so the new looping recording is never selected. Could you check whether the recorded loop actually replays on a clean checkout? As it stands the flag has no observable effect in the default demo config.

dpacheconr:
> @dpacheconr Thanks for the updates. The import-side-effect fix looks good (`init_feature_flags()` called from `run.py`), and `aiSlowResponse` works. But I still can't get `aiRunawayAgent` to loop under the default `USE_VCR=True`. I built the branch and toggled it on: flagd serves `20`, and I confirmed the agent injects the runaway instruction into the system prompt, but the response for `"Show all available products in the store."` is byte-identical to the flag-off response (4 messages, no repeated `get_recommendations`).
> 
> It looks like the fuzzy `similar_body_matcher` matches the older non-looping recording for that same prompt first (identical human message, score over threshold), so the new looping recording is never selected. Could you check whether the recorded loop actually replays on a clean checkout? As it stands the flag has no observable effect in the default demo config.

Good catch — you were right that this wasn't fully fixed. Turned out to be two separate issues:

The cassettes were recorded with runaway_iterations=8, but demo.flagd.json's aiRunawayAgent "on" variant is 20. The system prompt text baked into the recording ("at least 8 times") didn't match what the live flag now generates ("at least 20 times"), so replay diverged on the very first turn.

More fundamentally, get_recommendations hits frontend's /api/recommendations, which returns randomized results on every call. Since the runaway scenario is built on repeated get_recommendations calls, each replay's tool-result content differed from what was on tape, and our fuzzy VCR matcher was scoring that randomized content — so the match score decayed round over round until it dropped below threshold and playback failed outright, independent of the iteration-count mismatch.

Fixed both: re-recorded both cassettes at 20 iterations, and changed patch_vcr.py's similarity scoring to ignore tool-result content (it now only compares that a tool responded, not what it returned) — that's what actually makes the loop replay deterministically now, since tool output was always going to be random. Verified locally that USE_VCR=True replay now reproduces the same GRAPH_RECURSION_LIMIT crash the flag is meant to demonstrate, for both models, with zero real network calls.

ps48 (approved):
Verified on the latest head: with `aiRunawayAgent` on, the agent now loops and hits `GRAPH_RECURSION_LIMIT` (HTTP 500) under the default `USE_VCR=True`, and with it off the same prompt returns normally. The other load-generator prompts still replay cleanly, so the tool-result normalization in the matcher didn't regress normal playback. `aiSlowResponse` and the `init_feature_flags()` refactor look good too. LGTM, thanks for working through the VCR matching.

PR #3960, opened 2026-09-08
feat(agent): enable TRACELOOP_TRACE_CONTENT for LLM tracing

dpacheconr:
# Changes

Sets `TRACELOOP_TRACE_CONTENT=true` for the agent service so LLM prompt/response
content is captured in traces.

Currently the agent's Traceloop instrumentation (`Traceloop.init()` in
`src/agent/run.py`) only emits latency/token/status data for LLM spans — prompt
and completion text are not captured unless this flag is set.

Corresponding Helm chart update: open-telemetry/opentelemetry-helm-charts#2395

## Merge Requirements
- [ ] `CHANGELOG.md` updated to document new feature additions
- [ ] Appropriate documentation updates in the [docs][]
- [x] Appropriate Helm chart updates in the [helm-charts][]

Corresponding Helm chart update: https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2395


dpacheconr:
Corresponding Helm chart update: open-telemetry/opentelemetry-helm-charts#2395

.env
```diff
@@ -206,6 +206,7 @@ API_KEY=
 # Turn this to False for calling configured LLM always
 USE_VCR=True
 VCR_MATCH_THRESHOLD=0.85
+TRACELOOP_TRACE_CONTENT=true
```
ps48, line 209: I built the branch and checked both by running the stack and by evaluating the gate directly. `should_send_prompts()` in the langchain instrumentation is (`getenv("TRACELOOP_TRACE_CONTENT") or "true") == "true"`, so it returns True when the var is unset and when it is true, and only false disables it. 

In the running demo, LLM spans already carry prompt and completion content without this variable set. So setting it to true matches the existing default and does not change what is captured. Could you clarify where content was missing? If some environment had it set to false, that would explain it, but the demo default already captures content.

dpacheconr: Seems default behaviour is True 
https://www.traceloop.com/docs/openllmetry/privacy/traces
so probably issue on my setup

dpacheconr:
Default behaviour is True 
https://www.traceloop.com/docs/openllmetry/privacy/traces
this is no longer required

Issue #1368, opened 2024-02-06
Add micrometer spring boot 3 demo tracing app


majoferenc:
# Feature Request
Implementation of demo otlp micrometer tracing & metrics app with spring boot 3 & adding it to the service chain of this demo. As many real life use cases are using this type of configuration, it would be very helpful for many projects.

austinlparker:
If someone would like to spec this out we can consider it.

zzzk1:
@austinlparker @majoferenc , can i give it a try?

AnjaliMishra1st:
Hi everyone 

I'm Anjali Mishra, a 3rd-year CSE student interested in Cloud and DevOps. I’ve recently contributed to the Kubernetes ecosystem (kind and hydrophone – SIG Testing) with CI updates and documentation improvements.
I’m familiar with Java and Spring Boot and would like to work on this issue.
Before starting, I had a couple of quick questions:

Should the new Spring Boot service follow the structure used by other demo services?
Should Micrometer export telemetry directly via OTLP, or should the OpenTelemetry Java agent be used?

Happy to help implement the demo service. Thanks!

julianocosta89:
Hello @AnjaliMishra1st, welcome to the Demo 👋🏽

Ideally we wouldn't add a new service, but rather rewrite one that is already in Java, like the Ad service, or a service that has a language represented more than once in the Demo like Recommendation service (Python).

I personally would go for the 2nd option, but idk the opinion of the other maintainers. @open-telemetry/demo-maintainers wdyt?

AnjaliMishra1st:
Thanks for the clarification!
Got it — I’ll focus on enhancing an existing service instead of adding a new one.
I’ll explore updating a Java-based service (like Ad service) with Micrometer tracing and OTLP export.

Please let me know if that sounds good

cyrille-leclerc:
I'm aligned with @julianocosta89 . Thanks @AnjaliMishra1st for your enhancement proposal.

AnjaliMishra1st:
Thanks for the confirmation! 
I’ll proceed with enhancing an existing service and share updates soon.

jonatan-ivanov:
Since Spring Boot 3 will go out of OSS support at the end of next month, I would rename this issue and go with Spring Boot 4, see: https://spring.io/blog/2025/11/18/opentelemetry-with-spring-boot

AlistairHeus:
Hello, 
Is this still open? 

[...]

PR #3966, opened 2026-09-10
Rewrite the recommendation service in Spring Boot 4

adityareddy-dev:
# Changes

- Replace the Python recommendation service with a Java service built on Spring Boot 4.1 and Spring gRPC, as discussed in #1368. Same gRPC contract, same `recommendationCacheFailure` behaviour, same attribute and metric names from the telemetry schema.
- Telemetry comes from Micrometer instead of the Java agent: gRPC spans from the Spring gRPC observation interceptors, the `get_product_list` span and its `demo.*` attributes through Micrometer Tracing, the counter through the `MeterRegistry`, logs through the OpenTelemetry Logback appender, all exported over OTLP/HTTP from the standard `OTEL_*` environment variables that Spring Boot maps itself.
- Two known differences from the Python version: the gRPC span attributes are Micrometer's (`rpc.method`, `rpc.service`, `grpc.status_code`) rather than the OTel semconv set, and there are no `otel_sdk_*` self-observability metrics, so the service drops off that dashboard. The cache-leak memory growth shows on a new "JVM services (Memory)" panel in the demo dashboard.
- The frontend sends product ids as one repeated element per character (its API route hands the comma-joined query string to the gRPC client as an array). The Python undid that with a join and a split; the Java does the same, with a comment and a test that sends ids the way the frontend does.
- Tested with the module's unit tests and the full compose stack with the new image: recommendations never include the product being viewed, Jaeger shows frontend to recommendation to product-catalog with the demo attributes, counter and JVM metrics in Prometheus, logs with trace ids in OpenSearch, `test/telemetry` passes for the service. Docs page and helm chart env (collector HTTP port, Python variables removed) are follow-ups.

## Merge Requirements

essential items:


adityareddy-dev:
Ready for review. The helm chart change is open as a draft at https://github.com/open-telemetry/opentelemetry-helm-charts/pull/2398, to be merged only after a demo release with this service. The docs page for opentelemetry.io I will send separately.

Kielek:
I have some doubts. Demo already have 2 java services. I do not think that we should rewrite additional one to this langauge.

julianocosta89:
> I have some doubts. Demo already have 2 java services. I do not think that we should rewrite additional one to this langauge.

@Kielek we have 4 services in Python.
Spring is widely used in the industry and not part of the Demo.

We currently have only Ad that is Java Java.
Fraud detection is Kotlin, still JVM, but slightly different.

With this new service we can introduce some other ways to instrument Java code, as we have on the docs: https://opentelemetry.io/docs/languages/java/instrumentation/#instrumentation-categories

osullivandonal:
Hey, thanks for the contribution, I have some concerns with this, the big one is Java tends to a large resource hog on memory when running via Docker/k8s. I tried to run this locally and the service fails to start `x86-64` linux:

```bash
opentelemetry-demo recommendation-spring-boot-4 ❯ make start
docker compose --env-file .env --env-file .env.override -f compose.yaml -f compose.full.yaml -f compose.observability.yaml -f compose.extras.yaml up --force-recreate --remove-orphans --detach
[+] up 33/33
 ✔ Image ghcr.io/open-telemetry/opentelemetry-collector-releases/opentelemetry-collector-contrib:0.159.0 Pulled                                                                 7.9s
 ✔ Network opentelemetry-demo                                                                            Created                                                                0.0s
 ✔ Container astronomy-db                                                                                Healthy                                                               17.0s
 ✔ Container kafka                                                                                       Healthy                                                               18.5s
 ✔ Container valkey-cart                                                                                 Healthy                                                               17.0s
 ✔ Container prometheus                                                                                  Started                                                                1.1s
 ✔ Container jaeger                                                                                      Started                                                                1.2s
 ✔ Container grafana                                                                                     Started                                                                1.1s
 ✔ Container opensearch                                                                                  Healthy                                                               16.2s
 ✔ Container flagd                                                                                       Started                                                                1.1s
 ✔ Container opamp-server                                                                                Healthy                                                               17.0s
 ✔ Container telemetry-docs                                                                              Started                                                                1.3s
 ✔ Container otel-collector                                                                              Started                                                               16.3s
 ✔ Container cart                                                                                        Started                                                               17.6s
 ✔ Container flagd-ui                                                                                    Started                                                               17.2s
 ✔ Container payment                                                                                     Healthy                                                               21.9s
 ✔ Container email                                                                                       Healthy                                                               22.4s
 ✔ Container accounting                                                                                  Started                                                               17.6s
 ✔ Container fraud-detection                                                                             Started                                                               17.7s
 ✔ Container product-catalog                                                                             Started                                                               17.9s
 ✔ Container image-provider                                                                              Healthy                                                               24.0s
 ✔ Container quote                                                                                       Healthy                                                               24.0s
 ✔ Container shipping                                                                                    Started                                                               17.3s
 ✔ Container currency                                                                                    Healthy                                                               24.0s
 ✔ Container ad                                                                                          Healthy                                                               27.5s
 ✘ Container recommendation                                                                              Error dependency recommendation failed to start                      134.4s
 ✔ Container checkout                                                                                    Started                                                               23.4s
 ✔ Container frontend                                                                                    Created                                                                0.0s
 ✔ Container load-generator                                                                              Created                                                                0.0s
 ✔ Container frontend-proxy                                                                              Created                                                                0.0s
dependency failed to start: container recommendation is unhealthy
make: *** [Makefile:314: start] Error 1
```

The issue is probably around the resource limit for the container which is set to 500m.

Given that the demo is already very tight on resources, I am not sure about this addition.

adityareddy-dev:
Hey @osullivandonal ,

Thanks for taking the time to run this. I tried to reproduce the failure on my end but couldn't get it to fail. For context, my environment is an x86-64 machine running Docker Desktop on Windows 11 with the WSL2 backend, so the difference might be Windows and WSL2 versus native Linux. I ran the PR head (b4e5c76a) and tested both the standalone service and the full stack using your exact command. Here is what I saw:

- All 28 containers came up clean in 67 seconds with none unhealthy. The recommendation service stayed up, never restarted, and was not killed for memory.
- Under load, recommendation used 367 MiB of its 500 MiB limit (73%). In raw numbers, it uses about 80 MiB more than ad (287/300 MiB) and fraud-detection (282/300 MiB).
- This 500M limit is not from this PR. Upstream main has the same line and comment for the Python service.

To help figure out why it failed for you, could you check a few things?

1. The output of docker logs recommendation from the failed run.
2. Which commit you built.
3. How much memory your Docker or Podman VM has.

We can also look at a leaner configuration. I tested a run using -Xmx128m -XX:+UseSerialGC and a 300M limit. It started in 4.5 seconds and settled at 291 of 300 MiB under load, which is the same level the other two JVM services sit at. I checked the frontend logs over about 3.5 minutes of load and found no recommendation errors.

I did not test this leaner setup with the recommendationCacheFailure flag on, which is why the limit is currently 500M. I can test that flag next if you want to move toward the 300M limit.

Let me know what you think.

adityareddy-dev:
@osullivandonal Just checking in on this. Did you happen to get a chance to grab the docker logs recommendation output from that failed run? It would also be super helpful to know which commit you built and how much memory your Docker or Podman VM has.Like I mentioned before, I'm happy to test out that leaner 300M setup with the recommendationCacheFailure flag turned on if you want to keep the footprint as small as possible. Let me know if that sounds good to you!

@julianocosta89 Is there anything else you need from me on this one, or are we good to move ahead with the review?

julianocosta89:
@open-telemetry/java-instrumentation-approvers would someone be able to help us out reviewing this PR?

We know that OTel recomments one approach and Spring recommends another:
- https://opentelemetry.io/docs/zero-code/java/spring-boot-starter/ 
- https://spring.io/blog/2025/11/18/opentelemetry-with-spring-boot

Which approach do you see more on the Java users?
Would make sense to keep Micrometer in the OTel Demo?

trask:
I think the OpenTelemetry Demo should use OpenTelemetry's recommended approach (https://opentelemetry.io/docs/zero-code/java/spring-boot-starter/).

adityareddy-dev:
@trask and @julianocosta89, here is what was updated and verified:

- Traces & gRPC: Hooked the OTel gRPC library interceptors into Spring gRPC's global interceptor hooks since the starter doesn't handle gRPC out of the box. Spans are intact end-to-end with standard rpc.* attributes and our original demo.* attributes.
- Logs: The starter handles the logback appender natively, so I removed our custom logging setup. Logs successfully reach OpenSearch with trace IDs attached.
- Performance & Leak Testing: In a brief test run under load on an M4 Mac, memory sat around 194 MiB, which is roughly the same as the previous 211 MiB PR baseline. Testing with the cache-leak flag confirmed it successfully triggers restarts under the container limit as designed.
- Metrics & Dashboards: Turned on the SDK self-metrics using the OTEL_EXPERIMENTAL_SDK_TELEMETRY_VERSION=latest env var (matching the ad service), which restores recommendation to the self-monitoring dashboard. Separately, the starter labels JVM memory with jvm_memory_type like ad does, so the JVM memory panel query only needs that one label now.

**Just a heads-up for anyone reviewing the spans** 
The Java gRPC library still uses the older rpc.system naming convention (matching the frontend), while product-catalog has already moved to rpc.system.name.

PR #3987, opened 2026-09-16
fix(react-native): handle HTTP errors in request wrapper

### LINKED ISSUE #3782: Request() fetch wrapper doesn't check response.ok — HTTP errors with JSON bodies are treated as success

najahaboobackar:
## Summary

Fixes #3782.

The React Native `request()` fetch wrapper did not check `response.ok`, causing HTTP 4xx/5xx responses with JSON bodies to be treated as successful responses.

This change:

- Parses the response body safely.
- Checks `response.ok` before returning the response.
- Throws a meaningful error for HTTP error responses.
- Handles non-JSON error responses without exposing a JSON parsing `SyntaxError`.
- Preserves the existing behavior for successful responses.

The implementation is aligned with the existing behavior in `src/frontend/utils/Request.ts`.

## Testing

- Ran `npm run lint` in `src/react-native-app`
- Result: 0 errors, 3 pre-existing warnings unrelated to this change.
- No existing test/spec files were found for the React Native app.


osullivandonal:
Please sign the CLA

jpmunz (approved):

najahaboobackar:
@osullivandonal  could you pls approve the PR .

osullivandonal:
> @osullivandonal could you pls approve the PR .

Hi @najahaboobackar, thanks for the contribution!, I first need to review this PR when I get a chance.

PR #3990, opened 2026-09-17
fix(currency): fix getDouble fractional nanos calculation and validate currency codes

### LINKED ISSUE #3989: [currency] getDouble distorts fractional cent amounts and unchecked map lookup causes data races/division by zero
# Bug Report

Commit [3cb06d31](https://github.com/open-telemetry/opentelemetry-demo/commit/3cb06d31b7ad7e7eec8d9a2baef4f95d123d6a45) on `main`.

## Symptom

The `currency` service (`src/currency/src/server.cpp`) has two critical calculation, concurrency, and validation bugs in its `Convert` RPC implementation:

1. **10x–1000x Money Conversion Distortion on Sub-Dollar/Fractional Amounts (`getDouble`)**:
   In `getDouble(Money& money)`:
   ```cpp
   double decimal = 0.0;
   while (nanos != 0) {
     double t = (double)(nanos%10)/10;
     nanos = nanos/10;
     decimal = decimal/10 + t;
   }
   return double(units) + decimal;
   ```
   Protobuf `Money.nanos` represents fractional amounts as an absolute integer in nano-units ($10^{-9}$, e.g. `$0.50 = 500,000,000` nanos, `$0.05 = 50,000,000` nanos, `$0.01 = 10,000,000` nanos).
   The `while (nanos != 0)` loop counts iterations dynamically based on the number of significant digits until `nanos == 0`.
   - When converting `$0.05` (`50,000,000` nanos, 8 digits), the loop only runs 8 times instead of 9, dividing by $10^8$ instead of $10^9$. As a result, `$0.05` is parsed as `0.5` ($0.50) — **a 10x calculation error**.
   - When converting `$0.01` (`10,000,000` nanos, 8 digits), it is parsed as `0.1` ($0.10) — **a 10x calculation error**.
   - When converting `$0.001` (`1,000,000` nanos, 7 digits), it is parsed as `0.1` ($0.10) — **a 100x calculation error**.

2. **Thread Safety Violation, Map Pollution, and Division by Zero on Unknown Currencies (`currency_conversion[code]`)**:
   In `CurrencyService::Convert`:
   ```cpp
   Money from = request->from();
   string from_code = from.currency_code();
   double rate = currency_conversion[from_code];
   double one_euro = getDouble(from) / rate;

   string to_code = request->to_code();
   double to_rate = currency_conversion[to_code];
   ```
   - **Data Race / Memory Corruption**: `currency_conversion` is a global `std::unordered_map`. `operator[]` inserts a new entry with `0.0` when a key does not exist. Calling `Convert` concurrently with unknown/invalid currency codes performs concurrent unsynchronized writes while other gRPC threads are reading the map in `GetSupportedCurrencies` and `Convert`, causing data races and potential crashes.
   - **Map Pollution**: The inserted invalid code permanently pollutes `currency_conversion`, causing `GetSupportedCurrencies` to advertise the invalid currency with rate `0.0`.
   - **Division by Zero / Undefined Behavior**: `getDouble(from) / 0.0` yields `+INF`. Converting `+INF` to `long` in `(long)value` inside `getUnitsAndNanos` is undefined behavior in C++ (evaluating to `LONG_MIN` / `-9223372036854775808`). The service then returns `grpc::Status::OK` with corrupted negative values rather than returning `grpc::StatusCode::INVALID_ARGUMENT`.

1. `getDouble(const Money& money)` should compute fractional nanos as `static_cast<double>(money.units()) + static_cast<double>(money.nanos()) / 1e9`, accurately preserving all fractional cent amounts ($0.01, $0.05, etc.).
2. `Convert` should look up currencies with `find()` (read-only) and validate that both `from_code` and `to_code` exist with valid rates (`rate > 0.0`). If either currency is unsupported or missing, it should return `grpc::Status(grpc::StatusCode::INVALID_ARGUMENT, "Unsupported currency code")`.

1. Amounts with leading zeros in their fractional cents (e.g. `$0.05`, `$0.01`) are multiplied by 10x or 100x during conversion.
2. Unsupported currencies corrupt the global map, cause data races across gRPC threads, divide by zero, and return `Status::OK` with `units: -9223372036854775808`.

## Reproduce

1. Call `oteldemo.CurrencyService/Convert` with `from: { currency_code: "USD", units: 0, nanos: 50000000 }` ($0.05) and `to_code: "USD"`.
   - Result: Response is `{ currency_code: "USD", units: 0, nanos: 500000000 }` ($0.50) instead of `$0.05`.
2. Call `oteldemo.CurrencyService/Convert` with `from: { currency_code: "INVALID", units: 1, nanos: 0 }` and `to_code: "USD"`.
   - Result: Returns `Status::OK` with garbage units `-9223372036854775808`.
   - Call `GetSupportedCurrencies`: `"INVALID"` is now returned in the supported currencies list.

## Additional Context

- Affected component: `src/currency/src/server.cpp`
- Proposed fix:
  - Replace digit-loop in `getDouble` with `static_cast<double>(money.units()) + static_cast<double>(money.nanos()) / 1e9`.
  - Use `std::round` in `getUnitsAndNanos` to avoid float precision truncation.
  - Use `currency_conversion.find()` for safe read-only lookups and return `grpc::Status(grpc::StatusCode::INVALID_ARGUMENT, ...)` for unsupported currency codes.

bhuvan-somisetty:
# Changes

Fixes #3989

## Description

This PR fixes calculation accuracy, thread safety, map corruption, and error validation issues in the `currency` service:

1. **Fractional `nanos` Calculation (`getDouble`)**:
   - Replaced the digit-counting `while (nanos != 0)` loop with `static_cast<double>(money.units()) + static_cast<double>(money.nanos()) / 1e9`.
   - The previous loop counted iterations based on the number of non-zero digits, which truncated early for sub-dollar amounts containing leading fractional zeros (e.g. `$0.05 = 50,000,000` nanos or `$0.01 = 10,000,000` nanos ran only 8 iterations instead of 9, dividing by $10^8$ and resulting in a 10x calculation error to `$0.50` and `$0.10`).
   - Updated `getUnitsAndNanos` to round `nano` using `std::round` to avoid floating point truncation off-by-one errors.

2. **Safe Read-Only Currency Lookup & Validation (`Convert`)**:
   - Replaced unchecked `currency_conversion[code]` map access with read-only `.find()` checks for both `from_code` and `to_code`.
   - Previously, `operator[]` mutated the global `std::unordered_map` on unknown currencies across concurrent gRPC threads without locks, causing data races, memory corruption, and map pollution in `GetSupportedCurrencies`.
   - Validates that both source and target currencies exist with positive exchange rates (`rate > 0.0`), returning `grpc::Status(grpc::StatusCode::INVALID_ARGUMENT, ...)` on invalid currencies instead of dividing by zero (`+INF` -> `LONG_MIN`).

3. **Changelog**:
   - Added `.chloggen/3989-currency-nanos-validation.yaml` fragment.

## Merge Requirements

* [x] `CHANGELOG.md` updated to document new feature additions (via `.chloggen` fragment)


mwimpelberg28:
Nice cleanup overall, the nanos fix and the currency validation are both good. I think there's one real bug worth fixing before merge though.

In `getUnitsAndNanos` you switched from truncating to `std::round(rem * 1e9)`, and that can actually push `nanos` up to 1e9 without carrying into `units`. For example `3.9999999996` truncates to `units=3`, leaves `rem=0.9999999996`, and `round(rem*1e9)` comes out to exactly 1000000000 so you end up with `units=3, nanos=1000000000` instead of `units=4, nanos=0`. That breaks the invariant that `nanos` has to stay under 1e9, and it's a case the old truncating code just couldn't hit. Probably needs a carry into `units` after the round, or a clamp.

Also noticed the `Convert` span sets the rpc status-code attribute to "OK" up front and never touches it again on the two new invalid-currency early returns. The span status itself does get set to error, but the status-code attribute is stuck at "OK", so anything filtering/aggregating on that attribute in traces is going to miss these failures.

Two smaller things, not blocking:

- The from_code and to_code validation blocks are basically copy-pasted (find, check end(), check rate, add event, set status, return). Might be worth pulling out into one helper so they don't drift apart down the line.
- `#include <iostream>` got dropped but `main()` still uses `std::cout`. It's only compiling right now because `tracer_common.h` happens to include it transitively. I'd keep the explicit include so it doesn't break if that header ever changes.

bhuvan-somisetty:
Thanks for the review and great catch on the rounding overflow!

Updated the PR:
- Added carry handling into `units` in `getUnitsAndNanos` if `nano >= 1e9`.
- Updated the span's `rpc.response.status_code` attribute to `INVALID_ARGUMENT` / `CANCELLED` on failure paths.
- Extracted the lookup & validation logic into a `getRate` helper function.
- Restored `#include <iostream>`.

mwimpelberg28:
Nice, this addresses everything from the last round. 

One more thing though: the carry fix in getUnitsAndNanos only handles the positive overflow case (nano >= 1000000000). There's no symmetric check for negative underflow, so a negative value can round to nano == -1000000000, which breaks the same |nanos| < 1e9 invariant we just fixed the carry for, just on the other side of zero.

Convert is a public RPC, not restricted to checkout's always-positive amounts, so this is reachable. And we just fixed this exact class of bug on the shipping side (#3878)  might be worth mirroring that fix's symmetry here too, e.g.:
```

if (nano >= 1000000000) {
  unit += 1;
  nano -= 1000000000;
} else if (nano <= -1000000000) {
  unit -= 1;
  nano += 1000000000;
}
```

Everything else looks good to me.

bhuvan-somisetty:
Good catch on negative values. Added the symmetric underflow check (`else if (nano <= -1000000000)`) to `getUnitsAndNanos` as well.

mwimpelberg28:
Thanks for tightening up the currency validation and fixing the nanos rounding.  The fix itself works correctly, but I ran the built PR locally (docker compose) and confirmed a few gaps in how callers handle the new error, plus some smaller things worth addressing before merge.

Confirmed by running the app:

I rebuilt currency from this branch and brought up the full stack (checkout, frontend, product-catalog, payment, etc.) to exercise the new validation path end to end.

- Direct grpcurl call to currency.Convert with an unsupported to_code correctly returns:
`ERROR: Code: InvalidArgument, Message: Unsupported target currency: ZZZ
`  the validation itself is correct in isolation.
- GET /api/products?currencyCode=ZZZ returns a bare 500 Internal Server Error. The frontend log shows the raw gRPC error propagating unhandled:
`Error: 3 INVALID_ARGUMENT: Unsupported target currency: ZZZ
`  Previously this silently returned a wrong price; now it's a hard 500 for /api/products, /api/shipping, and every other route through ProductCatalogService.getProductPrice().
- More seriously: I placed a real order with a valid userCurrency (USD) but an invalid currencyCode query param on POST /api/checkout. The order actually completed, payment service logged Transaction complete with a real transactionId and the card was charged and the cart was emptied. Then the unguarded ProductCatalogService.getProduct() call in checkout.ts (used only for building the confirmation response) threw the same INVALID_ARGUMENT, and the customer-facing response was a bare 500. So this isn't a theoretical risk, a stale/unsupported currency code reaching checkout right now produces a charged, recorded order that the customer is told failed.

In this diff (src/currency/src/server.cpp):
- getRate()'s error path doesn't call logger->Error, unlike every other error path in this file. Log-based alerting won't see invalid-currency errors unless this is added.
- The two if (!status.ok()) { span->End(); return status; } blocks in Convert() are identical worth factoring into one helper so the two call sites can't drift.
- The getUnitsAndNanos carry-over fix (nano >= 1e9 / <= -1e9) patches a symptom of computing units and nanos independently. Computing a single scaled integer (e.g. std::llround(value * 1e9) then splitting into units/nanos via div/mod) would remove the boundary case structurally instead of needing this branch kept in sync.
- Minor: getRate's type param is const std::string&, which forces a temporary allocation from the "source"/"target" literals on every call, not just the error path. A const char* would avoid that.
- currency_conversion is still a plain mutable unordered_map. getRate() is now the only safe way to read it, but nothing enforces that, a future direct []/.at() access could reintroduce the concurrency bug this PR fixes.

Downstream, outside this diff, but now demonstrably reachable:
- checkout/main.go (prepareOrderItemsAndShippingQuoteFromCart) collapses the new INVALID_ARGUMENT into a generic Internal error, losing the distinction this PR was meant to introduce.
- Currency.provider.tsx restores the selected currency from a cookie without validating it against the current supported list, so a stale cookie is exactly what reaches the newly-strict backend.

Given the confirmed charge-then-500 behavior on checkout, I'd suggest at minimum wrapping the post-order getProduct calls in checkout.ts (and the equivalent calls in cart.ts/ProductCatalogService) so a currency error there degrades gracefully instead of masking a successful, paid order as a failure.

bhuvan-somisetty:
Thanks for the thorough review and testing the caller paths locally!

Updated the PR accordingly:
- Simplified `getUnitsAndNanos` using `std::llround(value * 1e9)` split via integer div/mod to handle all boundaries structurally.
- Made `currency_conversion` `const` to enforce read-only access.
- Added `logger->Error` in `getRate()` and switched `type` to `const char*`.
- Added fallback in `ProductCatalogService.getProductPrice()` to catch conversion errors and gracefully return the base price, preventing unhandled 500s on product routes and post-order checkout.

mwimpelberg28:
Re-tested the latest commit locally: rebuilt `currency` and `frontend` from this branch, brought up the core compose stack (ad, cart, checkout, currency, email, payment, product-catalog, quote, recommendation, shipping, flagd, image-provider, frontend), and re-ran the scenarios from my last comment plus a couple more.

## Fixed (confirmed by re-running against this commit)

**Checkout charge-then-500 is gone.** Same repro as before — cart with one item, `POST /api/checkout?currencyCode=ZZZ` with a valid `userCurrency: "USD"` body:

```
$ curl -s -w "\nHTTP %{http_code}\n" -X POST "http://localhost:$FP/api/checkout?currencyCode=ZZZ" -d "$BODY"
{"orderId":"87c66cb3-b672-11f1-bacd-9e59a4cba880","shippingTrackingId":"5495351e-bfdb-466a-a66e-228aaafecece",
 "shippingCost":{"currencyCode":"USD","units":8,"nanos":990000000},
 "items":[{"cost":{"currencyCode":"USD","units":175,"nanos":0}, ...}]}
HTTP 200
```
Payment log confirms the charge still happened, and this time the customer actually gets the success response instead of a 500:
```
{"service.name":"payment","transactionId":"3b08031a-05d9-4dfd-91c0-4da744ba092f","cardType":"visa",
 "lastFourDigits":"0454","amount":{"units":183,"nanos":990000000,"currencyCode":"USD"},"msg":"Transaction complete."}
```

**`server.cpp` cleanups**, all confirmed by diff (build succeeded, no warnings):
- `currency_conversion` is now `const std::unordered_map<...>` — no longer a bare mutable global.
- `getUnitsAndNanos` replaced the boundary-patch branch with `std::llround(value * 1e9)` + a single div/mod split.
- `getRate`'s `type` param changed from `const std::string&` to `const char*` — no more temp allocation per call.
- `getRate`'s error path now calls `logger->Error(...)` before returning, matching the rest of the file.

## New issue introduced by this commit

The checkout fix above works by adding a bare catch-all in `ProductCatalog.service.ts`:
```ts
try {
  return await CurrencyGateway.convert(price, currencyCode);
} catch {
  return price;
}
```
This swallows *any* failure from the currency call — not just an unsupported code, but the currency service being down, a timeout, anything — and silently falls back to the unconverted USD price. No log line, no span exception, nothing.

Confirmed live:
```
$ curl -s -o /dev/null -w "HTTP %{http_code}\n" "http://localhost:$FP/api/products?currencyCode=ZZZ"
HTTP 200
$ curl -s "http://localhost:$FP/api/products?currencyCode=ZZZ" | head -c 300
[{"id":"0PUK6V6EV0", ... "priceUsd":{"currencyCode":"USD","units":175,"nanos":0}, ...}]
```
compared to a valid code on the same product:
```
$ curl -s "http://localhost:$FP/api/products?currencyCode=EUR" | ...
"priceUsd":{"currencyCode":"EUR","units":154,"nanos":798761610}
```
`ZZZ` silently returns the raw USD price instead of erroring. And frontend logs for that request show nothing at all — no error, no warning, no trace of the failed conversion:
```
$ docker logs frontend --since 20s | grep -i "zzz\|invalid_argument\|unsupported\|error"
(node:1) MetadataLookupWarning: received unexpected error = All promises were rejected code = UNKNOWN
```
(that one line is an unrelated OTLP exporter warning — nothing currency-related at all). This is the same "fails silently with a wrong price" failure mode #3989 set out to fix, just moved up a layer and now covering every failure type, not only bad codes, with zero observability.

## Still open (unchanged from last comment)

**`/api/shipping` still bare-500s** on the same unguarded pattern — this file wasn't touched:
```
$ curl -s -w "\nHTTP %{http_code}\n" "http://localhost:$FP/api/shipping?currencyCode=ZZZ&itemList=%5B%5D&address=%7B%7D"
Internal Server Error
HTTP 500
```

**`checkout/main.go` still collapses `INVALID_ARGUMENT` into a generic `Internal`.** Placing an order with an invalid `userCurrency` itself (the realistic path once a stale `Currency.provider.tsx`-restored value reaches checkout) fails safely before charging the card, but still as an opaque error:
```
$ curl -s -w "\nHTTP %{http_code}\n" -X POST "http://localhost:$FP/api/checkout" -d '{"userCurrency":"ZZZ", ...}'
{"error":"Failed to place order."}
HTTP 500
$ docker logs payment --since 15s | grep -i "transaction complete"
(nothing — confirms no charge was attempted here, so this particular path is safe, just an unhelpful error)
```

**`Currency.provider.tsx`** still restores the persisted currency code on mount with no validation against the supported list — untouched by this commit.

**The two duplicated `if (!status.ok()) { span->End(); return status; }` blocks in `Convert()`** are still there verbatim — not factored into a helper.

## Summary

4 of the original findings are now fixed, 4 are still open, and this commit introduces one new issue: the checkout crash is gone, but by silently discarding currency-conversion failures instead of handling them, which trades a loud 500 for an unobservable wrong-currency price. Given `/api/shipping` sits right next to the fixed code path with the identical unguarded call, I'd suggest a shared, logged error-handling helper rather than patching each call site differently.

bhuvan-somisetty:
Thanks for the thorough review @mwimpelberg28! 

Addressed all the follow-up feedback:
- Simplified the `Convert` control flow in `server.cpp` so both `getRate` lookups share a single clean status check and span completion.
- Added `logger.warn` logging across `ProductCatalog.service.ts` and `shipping.ts` whenever conversion fails before falling back to USD.
- Wrapped gRPC errors with `%w` in `checkout/main.go` and extracted gRPC status via `errors.As`, allowing the frontend API route to catch `INVALID_ARGUMENT` and return HTTP 400 with the specific error details.
- Added validation in `Currency.provider.tsx` against supported currency codes, automatically falling back to USD if an invalid code is encountered in storage/cookies.

All CI checks and container builds are green. Ready for another pass when you have time!

maryliag:
Thanks for opening this pull request. Unfortunately, this contribution appears to have been generated primarily by a generative AI tool without sufficient human review or engagement, which is not in line with our [Generative AI Contribution Policy](https://github.com/open-telemetry/community/blob/main/policies/genai.md).

Maintainer time is limited, and we rely on contributors to be active, willing participants in the contribution process, understanding the code they submit, being able to discuss and iterate on it, and ensuring it fits the current state of the project. LLM output alone does not meet that bar.

I'm closing this PR for now. You are very welcome to come back with a contribution you have driven and reviewed yourself. Using an LLM as an assistant is fine, and disclosing that use (for example on the PR description) is encouraged. Please review the policy linked above before opening a new PR.

Thanks for your understanding.

Since we already provided you with several warnings about this, you will also be banned from interacting within the OpenTelemetry org for 7 days.

PR #3992, opened 2026-09-18
deps: bump anyio to 4.14.2 in agent, chatbot, mcp requirements (GHSA-82r6-8w77-94w6, GHSA-5p39-cfhj-2xmp)

katsugtgz:
Updates `anyio` to 4.14.2 in the pinned requirements of `src/agent`, `src/chatbot`, and `src/mcp`.

## What and why

`anyio < 4.14.2` is covered by two advisories published 2026-09-18:

- GHSA-82r6-8w77-94w6 (critical): TLSStream IDNA 2003 host name encoding enables potential certificate mismatch
- GHSA-5p39-cfhj-2xmp: TLSStream host argument does not participate in TLS handshake verification

The three services pinned 4.13.0 (chatbot) and 4.14.1 (agent, mcp), both in range. Only the anyio block in each requirements.txt changed: version plus the two sha256 hashes taken from the PyPI JSON metadata for the 4.14.2 sdist and wheel. No other pins or the pip-compile header were touched.

## How it was tested

- `python3 internal/tools/sanitycheck.py` passes (the repo's own check used by CI).
- `uv pip install -r src/<service>/requirements.txt` resolves and installs for all three services with hash verification against the updated pins; installed anyio reports 4.14.2.
- `osv-scanner --lockfile=src/mcp/requirements.txt` reports no anyio findings after the update (it reported both GHSAs before).

## Checklist

- [x] No secrets, no personal data in the diff.
- [ ] CI is green.
- [ ] Changes were tested against a running install of the demo.

Scope: the anyio pin and hashes in three `requirements.txt` files only.

Note: the remaining unchecked items do not apply to a dependency-only change; CI runs the sanity and build checks on this PR.


Kielek (approved):

Kielek:
@katsugtgz, you need to sing EasyCLA. Without this, we cannot proceed with this PR.

